PR #6, opened 2017-05-17
Update dependency of this package

eisoku9618:
I'm not familiar with PyPI, but we need this diff?
When I run ``pip install labelme && labelme xxx.jpg``, I got an error like ``no module name skimage.color``


eisoku9618:
close via https://github.com/wkentaro/labelme/pull/8

PR #7, opened 2017-05-17
Set the file's current position to the beginning

eisoku9618:
Hi @wkentaro ,
labelme is a very useful program ! (I found this repository through jsk apc repo)
But, this didn't work well in my environment without this PR.
I got an error ``IOError: cannot identify image file`` when I run ``labelme_draw_json hoge.json``.

http://stackoverflow.com/questions/8598673/how-to-save-a-pylab-figure-into-in-memory-file-which-can-be-read-into-pil-image

Thanks

### DIFF
--- a/labelme/utils.py
+++ b/labelme/utils.py
@@ -33,6 +33,7 @@ def labelcolormap(N=256):
 def img_b64_to_array(img_b64):
     f = StringIO.StringIO()
     f.write(base64.b64decode(img_b64))
+    f.seek(0)
     img_arr = np.array(PIL.Image.open(f))
     return img_arr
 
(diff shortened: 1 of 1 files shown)


wkentaro:
I confirmed this works, but it works without this change in my env.
Could you please tell me your platform? My env is Ubuntu 14.04 + Python2.7 and macOS Sierrra + Python2.7.

wkentaro:
I added tests: https://github.com/wkentaro/labelme/blob/master/tests/test_utils.py#L17
And the test passes on travis: https://travis-ci.org/wkentaro/labelme/builds/233339220

eisoku9618:
My env is also Ubuntu 14.04 + Python 2.7...
Let me check.

eisoku9618:
The difference between my env and yours is the version of pillow.
Mine is 2.3.0 (from apt ?), and travis one is 4.1.1(from pip)

```python
import PIL
PIL.PILLOW_VERSION
```

I think we should update pillow version automatically when we run ``pip install labelme``

eisoku9618:
I found we need pillow 2.8.0 or later to let ``labelme`` work well.

PR #9, opened 2017-05-18
Depend Pillow 2.8.0 or later

eisoku9618:
fix https://github.com/wkentaro/labelme/pull/7


wkentaro:
I confirmed that. Thank you for your investigation!

PR #16, opened 2017-08-22
Fix file open error

yuantangliang:
OS: ubuntu14.04 ,  python2.7,Qt4  
When I open image , I get error say as blow, and I change this line, everything is OK.
screenshot from 2017-08-22 11 30 06


wkentaro:
Thanks! I release v2.5.1

PR #40, opened 2017-11-18
Show supported image formats in error message while opening image

wkentaro:
Because JPEG image is not supported on macOS with PyQt4 installed via Conda.


wkentaro:
appveyor hangs for now.

PR #59, opened 2018-02-01
Stop sorting to avoid unexpected behavior

wkentaro:


simonsayshi:
@wkentaro sorry I'm confused, which part solved the all zero issue?

Issue #38, opened 2017-11-17
[Suggestion] Options in GUI not to save imageData replacing it by imagePath


mrgloom:
It would be great if I can check in GUI to not save `imageData` in json and to make `imagePath` relative to folder with images.

wkentaro:
I have implemented this feature in https://github.com/wkentaro/labelme/pull/73

PR #73, opened 2018-03-17
Support option not to save imageData

wkentaro:
Close https://github.com/wkentaro/labelme/issues/38

You can save label file without storing `imageData` with option of `labelme --nodata an_image.jpg -o an_image.json`.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -90,7 +90,7 @@ class WindowMixin(object):
 class MainWindow(QMainWindow, WindowMixin):
     FIT_WINDOW, FIT_WIDTH, MANUAL_ZOOM = 0, 1, 2
 
-    def __init__(self, filename=None, output=None):
+    def __init__(self, filename=None, output=None, store_data=True):
         super(MainWindow, self).__init__()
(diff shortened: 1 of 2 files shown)


labelme/app.py
```diff
@@ -949,16 +959,16 @@ def main():
     """Standard boilerplate Qt application code."""
     parser = argparse.ArgumentParser()
     parser.add_argument('filename', nargs='?', help='image or label filename')
-    parser.add_argument('-O', '--output', help='output label name')
+    parser.add_argument('--output', '-O', '-o', help='output label name')
+    parser.add_argument('--nodata', '-nd', dest='store_data',
```
ad-si: The short flag should only be one character, otherwise it looks like you'd call it with `-n` and `-d`

ad-si: Actually I wouldn't add a short flag at all. It makes shell scripts unreadable and normally you'd recall it from your shell history anyways, so no time saved...

wkentaro: Thanks! I updated my commit.

PR #76, opened 2018-03-21
Select label from QListWidget

### LINKED ISSUE #44: [Request] Select from labels list instead of creating new every time
It will great if we can select from labels list instead of creating new label every time like in `labelImg`

gui

Predefined classes are stored in `predefined_classes.txt` files that is near binary file.

wkentaro:
Close #44 

## Feature

- Select label from history
- Select label from label list specified by `labelme --labels=cat,dog,car,person`

## Demo

screen shot 2018-03-22 at 8 13 18

### DIFF
--- a/labelme/labelDialog.py
+++ b/labelme/labelDialog.py
@@ -39,7 +39,7 @@ class LabelDialog(QDialog):
     def __init__(self, text="Enter object label", parent=None):
         super(LabelDialog, self).__init__(parent)
         self.edit = QLineEdit()
-        self.edit.setText(text)
+        self.edit.setPlaceholderText(text)
         self.edit.setValidator(labelValidator())
         self.edit.editingFinished.connect(self.postProcess)
         layout = QVBoxLayout()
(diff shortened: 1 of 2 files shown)


labelme/labelDialog.py
```diff
@@ -36,22 +36,38 @@
...
+            self.labelList.addItems(labels)
+        self.labelList.currentItemChanged.connect(self.labelSelected)
+        layout.addWidget(self.labelList)
         self.setLayout(layout)
 
+    def addLabelHistory(self, label):
+        if self.labelList.findItems(label, Qt.MatchExactly):
+            return
+        self.labelList.addItem(label)
+        self.labelList.sortItems()
```
ad-si: Is this a good idea? Shouldn't the labels rather be sortable by the user? (e.g. most common first)
And keep the order of the CLI?

wkentaro: I think it is easier for the annotator to select if label list is sorted, so sorting is better in default.
And if requested, I can add an option like `--nosortlabels` to disable this sorting.
What do you think?

eitherlucas:
Is there an updated version of this?

PR #85, opened 2018-03-28
Check self.dirty before switching to double clicked file

### LINKED ISSUE #84: Switching to another file through the file list loads the clicked file without asking for saving
When annotating images in a directory, switching to the next file through double clicking the file in the list does not check if the current file needs to be changed.

jbvo:
Fixes #84.


seahawks8:
two lines of code and it solved, nice work.

Issue #90, opened 2018-03-29
Dropdown selection from label list


wkentaro:
Maybe with `--selectlabel` option.

wkentaro:
This is replaced by the completion in https://github.com/wkentaro/labelme/pull/92

PR #92, opened 2018-03-29
Show labels in main window

wkentaro:
Close https://github.com/wkentaro/labelme/issues/90

# What is this?

## Select label first (from unique label list) then annotate polygon

screenshot from 2018-03-30 20 13 25

I think this is better UI implementation of https://github.com/wkentaro/labelme/issues/44.
Do you have an idea? @ad-si

# How to try this?

```bash
cd examples/semantic_segmentation
labelme data_annotated --labels="$(tr '\n' , < labels.txt)" --nodata
```

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -87,6 +87,13 @@ class WindowMixin(object):
         return toolbar
 
 
+class EscapableQListWidget(QListWidget):
+
+    def keyPressEvent(self, event):
+        if event.key() == Qt.Key_Escape:
+            self.clearSelection()
+
+
 class MainWindow(QMainWindow, WindowMixin):
     FIT_WINDOW, FIT_WIDTH, MANUAL_ZOOM = 0, 1, 2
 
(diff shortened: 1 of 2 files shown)


jbvo:
I liked the UI as it was before, where the list was displayed below the edit box inside QLabelDialog. With the Combobox, the user has to click the combobox to display the list, while previously the list was already visible.

I've actually also made some small changes to improve the label dialog
1. Use arrow keys to change selected label from the list
2. Automatically select the correct row when opening the dialog for editing

This is the commit that applies these changes: https://github.com/wkentaro/labelme/commit/9df8cab5aade04a103c117f4dd8cfd41f3bc427d

jbvo:
One benefit I can imagine for having the labellist on the main window would be to highlight / select all matching labels, although I can imagine that requires quite some changes in the canvas class.

wkentaro:
I agree that showing label list in labeldialog.
The reason why adding label list in mainwindow is to display the annotator what object should be annotated.
So I think it is definitely necessary.

PR #96, opened 2018-03-30
UI adjustments (select label by arrow key, save/discard/cancel dialog, arrow labels.txt)

jbvo:
Some small UI adjustments that I implemented over the past few days while using labelme.

1. When closing or switching to another file, the dialog now asks to `save changes` aside of `discard changes` and `cancel`, more similar to other applications.
2. The label dialog preselects the previously selected label or the label that is under edit because of clicking in the main window label list. Also it allows to change the selected label of the list by using the arrow keys.
3. Extended loading labels from file, filtering empty line labels or clarity.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -796,7 +796,6 @@ class MainWindow(QMainWindow, WindowMixin):
             self.addRecentFile(self.filename)
             self.toggleActions(True)
             self.status("Loaded %s" % os.path.basename(str(filename)))
-            self.setWindowTitle('{} - {}'.format(__appname__, os.path.basename(str(filename))))
             if filename in self.imageList:
                 self.fileListWidget.setCurrentRow(self.imageList.index(filename))
             return True
@@ -1137,7 +1136,7 @@ def main():
     args = parser.parse_args()
 
     if args.labels is not None:
-        args.labels = args.labels.split(',')
+        args.labels = [l for l in args.labels.split(',') if len(l) > 0]
 
     app = QApplication(sys.argv)
     app.setApplicationName(__appname__)
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1130,18 +1130,19 @@ def main():
     parser.add_argument('--output', '-O', '-o', help='output label name')
     parser.add_argument('--nodata', dest='store_data', action='store_false',
                         help='stop storing image data to JSON file')
-    parser.add_argument('--labels', help='comma separated list of labels OR file containing one label per line')
+    parser.add_argument('--labels',
+                        help='comma separated list of labels OR file '
+                        'containing one label per line')
     parser.add_argument('--nosortlabels', dest='sort_labels',
                         action='store_false', help='stop sorting labels')
     args = parser.parse_args()
 
     if args.labels is not None:
         if os.path.isfile(args.labels):
-            args.labels = [l.strip()
-                           for l in open(args.labels, 'r').readlines()
-                           if len(l.strip()) > 0]
+            args.labels = [l.strip() for l in open(args.labels, 'r')
+                           if l.strip()]
         else:
-            args.labels = [l for l in args.labels.split(',') if len(l) > 0]
+            args.labels = [l for l in args.labels.split(',') if l]
 
     app = QApplication(sys.argv)
     app.setApplicationName(__appname__)
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -965,12 +965,22 @@ class MainWindow(QMainWindow, WindowMixin):
         return True
 
     def mayContinue(self):
-        return not (self.dirty and not self.discardChangesDialog())
-
-    def discardChangesDialog(self):
-        yes, no = QMessageBox.Yes, QMessageBox.No
-        msg = 'You have unsaved changes, proceed anyway?'
-        return yes == QMessageBox.warning(self, 'Attention', msg, yes|no)
+        if not self.dirty:
+            return True
+        mb = QMessageBox
+        msg = 'Save annotations to "{}" before closing?'.format(self.filename)
+        answer = mb.question(self,
+                             'Save annotations?',
+                             msg,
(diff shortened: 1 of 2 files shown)


labelme/app.py
```diff
@@ -794,7 +796,6 @@ def loadFile(self, filename=None):
             self.addRecentFile(self.filename)
             self.toggleActions(True)
             self.status("Loaded %s" % os.path.basename(str(filename)))
-            self.setWindowTitle('{} - {}'.format(__appname__, os.path.basename(str(filename))))
```
wkentaro: Why this change is necessary?

labelme/labelDialog.py
```diff
@@ -89,11 +89,20 @@ def postProcess(self):
...
         self.edit.setFocus(Qt.PopupFocusReason)
         if move:
             self.move(QCursor.pos())
         return self.edit.text() if self.exec_() else None
 
+    def keyPressEvent(self, e):
+        if e.key() in [Qt.Key_Up, Qt.Key_Down]:
+            self.labelList.keyPressEvent(e)
+        else:
+            super().keyPressEvent(e)
```
wkentaro: The label changes using arrow keys without this change in my env.

labelme/app.py
jbvo: Since I was added a `setWindowTitle` call to the `setDirty` method, I thought it would be clearer if this specific `setWindowTitle` call moved along with it from `loadFile` to `setClean`.

At this moment it won't make a difference if the call is in `loadFile` or in `setClean`, but if at some point an undo function would be implemented if should be in `setClean` anyway to remove the asterisk from the window title.

labelme/labelDialog.py
```diff
@@ -90,19 +90,15 @@ def postProcess(self):
...
+        items = self.labelList.findItems(text, Qt.MatchFixedString)
+        if items:
+            assert len(items) == 1
+            self.labelList.setCurrentItem(items[0])
         self.edit.setFocus(Qt.PopupFocusReason)
         if move:
             self.move(QCursor.pos())
         return self.edit.text() if self.exec_() else None
-
-    def keyPressEvent(self, e):
```
jbvo: Why did you remove the keyPressEvent?

jbvo: Why did you remove the `keyPressEvent`?

wkentaro: Do you really need this?
The label changes using arrow keys without this change in my env. It is only on macOS..?

labelme/app.py
wkentaro: Ok I see.

labelme/labelDialog.py
wkentaro: Could you re-check this?

jbvo: I'm on Ubuntu 16.04. In my case arrow keys only work if the listwidget has the focus, but we set the focus to the editbox. This keypress event allows moving the listwidget selection without first moving the focus to the listwidget.

jbvo: I've double checked it and without the keypress event I have to either press TAB 3 times or click on the listwidget before I can use the arrow keys for selecting a label from the list.

labelme/labelDialog.py
```diff
@@ -90,19 +90,15 @@ def postProcess(self):
             self.edit.setText(self.edit.text().trimmed())
 
     def popUp(self, text=None, move=True):
-        if isinstance(text, str):
-            self.edit.setText(text)
-            self.edit.setSelection(0, len(text))
-            i = self.labelList.findItems(text, Qt.MatchFixedString)
-            if len(i) == 1:
-                self.labelList.setCurrentItem(i[0])
+        if text is None:
```
jbvo: The reason I had put the complete code block inside the if was for the case of labeling many instances of the same label in 1 image. In that case it is convenient to not have to enter the label name again, but instead still having it from the previous label action.

If you prefer to not keep the previous label in the editbox, I would also suggest to add the following block below:

```py
if items:
    assert len(items) == 1
    self.labelList.setCurrentItem(items[0])
else:
    self.labelList.clearSelection()
```

labelme/labelDialog.py
wkentaro: I understood. But the implementation did not work on macOS, so I implemented it with another approach.

labelme/labelDialog.py
wkentaro: I think keeping the previous label is a good idea.
However, how can it be achieved?
Where do you keep the previous label?

wkentaro: Ah, now I understand. And I found it is necessary to ensure `text=None`.

wkentaro:
@jbvo Could you check and review my commits?

labelme/labelDialog.py
jbvo: For my understanding of python and pyQt: did the up and down key not work at all? Were the escape and enter key working for accepting and rejecting?

One thing I noticed when checking the differences is the call `else` case, my initial implementation:
```py
super().keyPressEvent(e)
```
your implementation:
```py
super(LabelQLineEdit, self).keyPressEvent(e)
```
Your implementation is more compatible as far as I understand from python docs, as `super()` was introduced in python 3 for convenience.

jbvo:
I tested the adjusted label dialog and it works. I've put a question about the `keyPressEvent`, but that's not very critical, so merging would be fine for me as is.

labelme/labelDialog.py
wkentaro: > For my understanding of python and pyQt: did the up and down key not work at all? Were the escape and enter key working for accepting and rejecting?

Hmm, I'm not sure about your question, but all keys, up, down, escape, enter key work well;
up and down keys are for selecting label from list, escape and enter keys are for accepting or rejecting.

wkentaro: > Your implementation is more compatible as far as I understand from python docs, as super() was introduced in python 3 for convenience.

Yes, my implementation is for compatibility on python2.

Issue #83, opened 2018-03-27
Remove last point while drawing


jbvo:
Sometimes I find myself wishing there was a way to move or remove the last point of a polygon while drawing it. Typically this happens in situations where the polygon consists of a large number of points (more than 10?) and I don't want to either finish it and remember to correct the points or start redrawing the polygon from scratch.

Maybe it would be possible to assign a key to remove the last point? E.g. backspace or something like that? I would be interested in other solutions as well.

wkentaro:
I think currently there is no solution for this.
Maybe adding this feature by `Ctrl-z` or `Ctrl-u` to **undo** the last annotated point.

jbvo:
`Ctrl-z` sounds good to me too.

seahawks8:
agree, it would be nice

PR #112, opened 2018-04-02
Undo last drawn point on current polygon

jbvo:
Solves #83.

The reason I choose backspace as the hotkey instead of Ctrl-Z is mainly laziness: it's easier to press 1 key than a 2 key combination. Since there's no text entry in the main window I figured this would not be a problem.


wkentaro:
I found Ctrl+Z is useful when I annotate polygons with a mouse with my right hand.

Issue #82, opened 2018-03-27
Change label priority / move label to back or front


jbvo:
When segmenting an image, sometimes labeling becomes easier if labels overlap. Typical examples of this are floors or walls in indoor environments.

Overlapping labels works OK, the last label in the list will be the one that will be applied, i.e. it has the highest priority. However I couldn't find a way to change the order of labels after drawing except by deleting and drawing again. What do you think of adding this?

jbvo:
Example of overlapping labels in the wrong:
![labelme_overlap](https://user-images.githubusercontent.com/24601408/37969715-b4cf5656-31d1-11e8-8f0a-47fb6f15d893.jpg)

jbvo:
Actually I was thinking of drag and drop within the list and a 'bring to top' and 'send to back' in the context menu that appears when right clicking a polygon on the canvas.

In https://github.com/jbvo/labelme/tree/label_priority I've made a start on the drag and drop, which works OK in terms of output. I'm figuring what is the nicest way to catch an internal drag / drop event, so that I can update `self.dirty` based on that and update the order of `canvas.shapes`.

PR #117, opened 2018-04-07
Make labeled polygons sortable

wkentaro:
Close https://github.com/wkentaro/labelme/issues/82

screen shot 2018-04-07 at 20 28 33
screen shot 2018-04-07 at 20 28 20


jbvo:
When generating a dataset based on a sorted label list, which label should be the highest priority, i.e. which one should be in the foreground? The top or bottom label?

wkentaro:
The bottom labe will have higher prioprity if you use shapes_to_label function. 
But you can change the order of shapes in your script, or use the coco dataset (which saves instance polygons into a json file) format rather than pascal voc format (which saves class and instance ids into a label image).

PR #122, opened 2018-04-10
Zooming fix

jbvo:
3 main changes:
1. Fixes scroll wheel zooming in my Ubuntu setup.
2. Zoom shortcut adjusted to be more similar to browsers: `Ctrl+0` to zoom to 100%, `Ctrl+=` to zoom in, `Ctrl+-` to zoom out.
3. Zoom into the cursor location instead of towards (0,0).

Regarding the Ubuntu setup: `pixelDelta()` is always 0 in my case, this is mentioned in the Qt documentation as a X11 specific issue: https://doc.qt.io/qt-5/qwheelevent.html#pixelDelta. I now changed it to `angleDelta()` instead, which based on the documentation I would expect to work on all platforms. However I don't have a Mac to test it.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -787,9 +787,9 @@ class MainWindow(QtWidgets.QMainWindow, WindowMixin):
             y_shift = round(pos.y() * canvas_scale_factor) - pos.y()
 
             self.scrollBars[Qt.Horizontal].setValue(
-                    self.scrollBars[Qt.Horizontal].value() + x_shift)
+                self.scrollBars[Qt.Horizontal].value() + x_shift)
             self.scrollBars[Qt.Vertical].setValue(
-                    self.scrollBars[Qt.Vertical].value() + y_shift)
+                self.scrollBars[Qt.Vertical].value() + y_shift)
 
     def setFitWindow(self, value=True):
         if value:
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -290,7 +290,7 @@ class MainWindow(QtWidgets.QMainWindow, WindowMixin):
         self.zoomWidget.setEnabled(False)
 
         zoomIn = action('Zoom &In', functools.partial(self.addZoom, 10),
-                        'Ctrl++', 'zoom-in', 'Increase zoom level',
+                        ['Ctrl++', 'Ctrl+='], 'zoom-in', 'Increase zoom level',
                         enabled=False)
         zoomOut = action('&Zoom Out', functools.partial(self.addZoom, -10),
                          'Ctrl+-', 'zoom-out', 'Decrease zoom level',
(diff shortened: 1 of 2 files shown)


serycjon:
I think Ctrl++ is much more sensible keybinding for zoom in(and I think it is default in all the applications I use). Could you perhaps keep both Ctrl++ and Ctrl+= for zoom in?

jbvo:
Since I was working on a laptop without numpad, I had not realized that `Ctrl++` also works with `numpad +`. To trigger `Ctrl++` I actually had to press `Ctrl`, `Shift` and the `=` key, so 3 keys. When I open Firefox or Chrome, `Ctrl+=` actually zooms in, so I would vote to keep both `Ctrl++` and `Ctrl+=`.

wkentaro:
Is `Ctrl+=` usual?
Below pages say Chrome and Firefox support `Ctrl++` and `Ctrl+-` to zoom in and out.
https://support.google.com/chrome/answer/96810?hl=en
https://support.mozilla.org/en-US/kb/font-size-and-zoom-increase-size-of-web-pages

wkentaro:
Actually, I am using Japanese keyboard. So I need to press 
- `Ctrl` `Shift` and `-` keys (3 keys) to trigger `Ctrl+=`

jbvo:
I read that on the support pages as well, but if I try it in Chrome and Firefox I don't need to press `shift` to zoom in, so in practice I'm really pressing `Ctrl+=` in that case.

For clarity, I'm using a US keyboard, a layout like this: https://en.wikipedia.org/wiki/QWERTY#/media/File:KB_United_States.svg

jbvo:
Maybe instead of the string definitions, we should use QKeySequence: https://doc.qt.io/qt-5/qkeysequence.html?

wkentaro:
I updated the code with using QKeySequence.
Could you try it on your environment? (I have only macOS environment at home..)

jbvo:
It doesn't help, so let's go back to `['Ctrl++', 'Ctrl+=']`

wkentaro:
I tested on Ubuntu 16.04.
On firefox

- `Ctrl+-`: zoomOut
- `Ctrl+=`: zoomIn
- `Ctrl++`: zoomIn

On chrome

- `Ctrl+-`: zoomOut
- `Ctrl+=`: zoomOut
- `Ctrl+^`: zoomIn

wkentaro:
Maybe the problem is

On Japanese Keyboard, 
`-` and `=` are on the same key
On English Keyboard,
`+` and `=` are on the same key

wkentaro:
I disagree with adding `Ctrl+=` for zoomIn, but I think it is acceptable adding `Ctrl+^` for zoomIn.
I'm merging this for now. Please comment if you have further opinion.

wkentaro:
Force pushed to remove revert commits.

serycjon:
Seeing this discussion and having a completely different keyboard layout myself (I use Programmer Dvorak layout), I think we could start to think about customizable keybindings, what do you think about it?

wkentaro:
>  I think we could start to think about customizable keybindings, what do you think about it?

That sounds great. Could you please open an issue?

serycjon:
Here: https://github.com/wkentaro/labelme/issues/126

PR #155, opened 2018-05-07
support CJK in labels file

xikuibin:
If label list file contains Asian characters,  --labels will cause error when staring application. Use utf-8 encoding to improve this.


xikuibin:
Close and redo again.

PR #156, opened 2018-05-07
support cjk label list file

xikuibin:
support Asian Character in label list files specified with --labels


wkentaro:
Thanks. I created another PR: https://github.com/wkentaro/labelme/pull/157 to support python2 also.

PR #166, opened 2018-06-10
Add image level flag annotation

wkentaro:
This PR adds a new key 'flags' to the JSON (annotation file).


alexisk1:
@wkentaro is it possible to leave out the flag key in the json if it is empty?

alexisk1:
@wkentaro I would be happy to help actually if you want me too :)

vishalmandley:
Please add a feature that can copy annotations from one image to another, if already exists, please explain about it. Thanks.

PR #172, opened 2018-06-13
Add the feature of adding vertex

AlexMa011:
Easily add a vertex to existing polygons.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
   global:
     # used by ci-helpers
     - SETUP_XVFB=true
-    - PIP_DEPENDENCIES='flake8 pytest pytest-qt'
+    - PIP_DEPENDENCIES='hacking pytest pytest-qt'
 
     - MPLBACKEND=TkAgg  # for osx
 matrix:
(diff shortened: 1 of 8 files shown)


wkentaro:
Thank you for the PR, and your implementation looks great!

Currently, it automatically adds a new point by selecting the segment, but IMO that would be great if it displays a dialog like below.

screen shot 2018-06-13 at 16 47 20

Also, sometimes user may want to add a new vertex for a segment which is very short. In this case, user must zoom-in very closely in current implementation.


Because of above reasons, I suggest below UI and implementation for that:

1. User selects a vertex.
2. Right click.
3. From dialog, user selects `Add a vertex before this`, or `Add a vertex after this`.
4. It adds a new vertex in the middle of the segment defined by the selected vertex and that before/after.

What do you think about this?

AlexMa011:
The right click operation seems more robust in the hard case. But the "before or after" thing is not appropriate in my opinion, because it's hard to tell which vertex is before or after.

wkentaro:
I agree with your idea. I updated the commit.

wkentaro:
Thanks!

PR #209, opened 2018-08-03
paint the interior of current drawing polygon

wenliwyan:
For better eye guidance, the interior of the current shape is painted and updated when moving mouse. This feature can be frequently found in other labeling tools.


wkentaro:
Could you please attach the Screenshot?

wkentaro:
Thanks. This feature is really nice!

PR #210, opened 2018-08-03
Fill drawing shape with a pale color

wkentaro:
Close #209 

screen shot 2018-08-03 at 21 50 21
screen shot 2018-08-03 at 21 50 32


wkentaro:
@wenliwyan I updated the code to make it optional to fill the drawing shape.

wenliwyan:
@wkentaro It's great~ btw, had a hard time searching for the difference between these two screenshots, before I found the labeled melon... :)

PR #211, opened 2018-08-03
Fix caption shown in label_viz

YutoUchimi:
I fixed loading order of labels in `draw_label` function.

Below are `label_viz.png` images generated after `labelme_json_to_dataset JSON_FILE`.

## Before this PR
before

## After this PR
after

### DIFF
--- a/labelme/utils/draw.py
+++ b/labelme/utils/draw.py
@@ -71,7 +71,7 @@ def draw_label(label, img=None, label_names=None, colormap=None):
 
     plt_handlers = []
     plt_titles = []
-    for label_value, label_name in enumerate(label_names):
+    for label_value, label_name in enumerate(sorted(label_names)):
         if label_value not in label:
             continue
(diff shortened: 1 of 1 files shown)


wkentaro:
It seems this issue is caused by the cli/draw_json.py and cli/json_to_dataset.py not utils/shape.py.
I tried to fix this. Could you try https://github.com/wkentaro/labelme/pull/212?

wkentaro:
BTW, the feature of `labelme_json_to_dataset` script is really poor and it is only used for the tutorial.
If you don't know, I recommend you to use [`labelme2voc.py`](https://github.com/wkentaro/labelme/blob/master/examples/semantic_segmentation/labelme2voc.py) for semantic segmentation (we also have the conversion script for instance segmentation).

YutoUchimi:
#212 works fine for me. Thanks a lot!

PR #241, opened 2018-11-03
Convert bounding box annotated files to VOC-like dataset

wkentaro:
screen shot 2018-11-03 at 20 52 32


jdhao:
just FYI, the xml produced by labelme2voc.py does not work for the voc dataset in the maskrcnn-benchmark repo (code [here](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/data/datasets/voc.py#L95)) since it requires the `difficult` flag for each object to be valid `0` or `1`.

But the difficult flag produced by labelme2voc.py is incomplete by default.  It seems that at least [this line](https://github.com/wkentaro/labelme/blob/master/examples/bbox_detection/labelme2voc.py#L116) should be changed from

```
 maker.difficult(),
```

to 

```
 maker.difficult('0'),
```

to make the xml files work for maskrcnn-benchmark.

PR #242, opened 2018-11-06
Support linestrip shapes

ompugao:
I've added linestrip (a series of lines) shape.
please take a look if you don't mind.

labelme_linestrip


ompugao:
thank you!

PR #271, opened 2018-12-07
add option to skip label selection popup

jrouquie:
If ~/.labelmerc containt `display_label_popup: false`, and if a label is already selected, then do not ask for which label to apply to the just completed polygon: use the selected label.


wkentaro:
Thanks. Merged via https://github.com/wkentaro/labelme/commit/23980e6e49b17839bbceac71b2eb32654f09251e

PR #274, opened 2018-12-12
added cli/json_to_label_instances.py

kaolin:
added a cli tool to export labelled segments as separate images for training/experimentation 

exports segmented items as transparent-backed pngs, masked by their shape, cropped to their bounding rect, in separate directories for each label.


wkentaro:
Sorry, could you explain why you need this, and why you don't save instance mask as mask image?

kaolin:
I may have overlooked something simple, but I'm handing the data off to folks who have their own frameworks for training and don't want to duplicate the effort of masking and cropping across them or training on masked data. (and given how long it took me to figure out how to get labelme's bool mask into something that PILlow would accept... ;_; )

A byproduct of this format is that it also made it easier to see labeling errors, when there are so many labels on an image that the colors become too self-similar.

If you thought saving out mask pngs (?) independently against the image would be useful for folks, I'd be happy to make a mode for that.

wkentaro:
I see.

For training, VOC format of instance masks is already supported: https://github.com/wkentaro/labelme/tree/master/examples/instance_segmentation#convert-to-voc-like-dataset,
but COCO format is not supported, so I think adding it would be useful for many people.

For validation while annotation, I think we should add some useful mode for that (e.g., differentiate colors of polygons according to the class and instance, mode to visualize only specified class masks).

PR #284, opened 2018-12-29
fix error AttributeError: 'module' object has no attribute 'QAbstract…

wpbird007:
if  set sort_labels == false in ~/.labelmerc file, could got error
AttributeError: 'module' object has no attribute 'QAbstractItemView'
fix this error


wkentaro:
Thanks!

Issue #232, opened 2018-10-15
 auto rotate


lijian-1994:
when i use labelme to open pictures,some will be auto rotate.why?who can help me?

wkentaro:
Maybe it is related to the EXIF, but not sure how to fix at this time. I put some links below related to this issue.
- https://photo.stackexchange.com/questions/84292/if-an-image-is-rotated-losslessly-why-does-the-file-size-change
- https://stackoverflow.com/questions/13872331/rotating-an-image-with-orientation-specified-in-exif-using-python-without-pil-in

wkentaro:
Could you give me an example image to reproduce this?

ramakrishna1592:
Can there be an option to rotate the image by 90 deg clockwise or anticlockwise.

kaolin:
I've run into this as an EXIF issue. In my processes I've found stripping/correcting the source images is most reliable because you need to make sure that EVERY tool in your system handles EXIF correctly... and invariably, someone/something doesn't. :/ Stackexchange answer looks good.

Adding an option to rotate the image gets wonky...you really don't want to add in a "secondary" EXIF property. Maybe a purely interface rotation would be alright? But I don't think that would be good until the EXIF was properly respected, or you're going to have to "unfix" your data...

ryouchinsa:
RectLabel is an offline image annotation tool for object detection and segmentation.
Although this is not an open source program, with RectLabel you can remove the EXIF orientation flags from images and convert images to the front orientation. If necessary, you can rotate the image right/left by 90 degrees.
https://rectlabel.com/help#remove_exif_orientation_flags

PR #302, opened 2019-01-25
Add rotate right and rotate left button in edit menu

ctudoudou:
I found the function that needs to be rotated when I use it, so I added these two functions.

Thanks.


wkentaro:
Hmm. I don't think it is a good idea to edit the original image file.
According to https://github.com/wkentaro/labelme/issues/232, the reason why you need to rotate is the python script does not correctly read EXIF information.
I think we should add fix to read EXIF before thinking of adding this kind of feature.

ctudoudou:
I think I fix the #232.

Thanks.

wkentaro:
Thanks. I created a refactored version at https://github.com/wkentaro/labelme/pull/305.

ryouchinsa:
RectLabel is an offline image annotation tool for object detection and segmentation.
Although this is not an open source program, with RectLabel you can remove the EXIF orientation flags from images and convert images to the front orientation. If necessary, you can rotate the image right/left by 90 degrees.
https://rectlabel.com/help#remove_exif_orientation_flags

PR #325, opened 2019-02-18
Delete label file

### LINKED ISSUE #324: Delete label file
Recently, I want to delete the label file (*.json) in some situations:

- Re-label the entire image when there are too many labels I want to remove.
- Ignore some data (another way is to add an 'ignore' flag).

I propose to add an action to delete the label file.

narumiruna:
Fixes #324


wkentaro:
@narumiruna Thanks!

PR #334, opened 2019-02-26
16bit image and mim file

ttagu99:
I need to draw label by 16bit Tiff Image  or mim Image.
mim file is matrox image library file.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -36,7 +36,8 @@ from labelme.widgets import LabelDialog
 from labelme.widgets import LabelQListWidget
 from labelme.widgets import ToolBar
 from labelme.widgets import ZoomWidget
-
+from skimage import io as skio
+from scipy.misc import bytescale
 
 # FIXME
 # - [medium] Set max zoom value to something big enough for FitWidth/Window
@@ -1190,7 +1191,10 @@ class MainWindow(QtWidgets.QMainWindow, WindowMixin):
 
     def loadImageFile(self, filename):
         try:
-            image_pil = PIL.Image.open(filename)
+            image_raw = skio.imread(filename)
+            image_np = bytescale(image_raw)
+            image_pil = PIL.Image.fromarray(image_np)
+            #image_pil = PIL.Image.open(filename)
         except IOError:
             return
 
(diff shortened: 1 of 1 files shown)


labelme/app.py
```diff
@@ -36,7 +36,8 @@
 from labelme.widgets import LabelQListWidget
 from labelme.widgets import ToolBar
 from labelme.widgets import ZoomWidget
-
+from skimage import io as skio
```
wkentaro: `scikit-image` is not included in the dependencies. Isn't that possible to use PIL to open mim file?

ttagu99: PIL can not open the mim file. If skimge can not be added to the dependency, this pull request will be closed.

wkentaro: It can be an optional dependency.

wkentaro:
Can you give me a sample file of .mim?

ttagu99:
16bit mim file is attatched.

2019년 3월 2일 (토) 오전 2:03, Kentaro Wada <notifications@github.com>님이 작성:

ttagu99:
> Can you give me a sample file of .mim?

[16bit_gray.zip](https://github.com/wkentaro/labelme/files/2925654/16bit_gray.zip)
mim file can not be attached so, I attached zip file

wkentaro:
I a bit worked on this and found there are a lot of image file formats: https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html,
and still `.mim` file is not listed there.
If I support all of the file extensions, `labelme` needs to load a lot of images to construct the file list, which is not a good thing.

I think maybe you should first convert `.mim` files to a standard file (e.g., `.jpg` or `.png`) and then annotate on them.
What do you think?

wkentaro:
Now I think maybe we can add additional image extensions with the custom configuration.

ttagu99:
> I a bit worked on this and found there are a lot of image file formats: https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html,
> and still `.mim` file is not listed there.
> If I support all of the file extensions, `labelme` needs to load a lot of images to construct the file list, which is not a good thing.
> (also, maybe in the future someone may say we should support `.npy` file also..)
> 
> I think maybe you should first convert `.mim` files to a standard file (e.g., `.jpg` or `.png`) and then annotate on them.
> What do you think?

I will use forked repo. so this pull request is closing

PR #339, opened 2019-03-03
Edit labelme2coco.py

lintangsutawika:
labelme2coco.py generates coco format observable with pycocotools library


wkentaro:
Thanks! Can you check if https://github.com/wkentaro/labelme/pull/343  does work?

PR #341, opened 2019-03-05
Assign width and height after loading

narumiruna:
Problem: `labelme --nodata` does not save image width and height.


wkentaro:
I can't reproduce it. (Also I'm not sure why 0bc9a05 does fix it).

narumiruna:
Sorry, I was wrong. LOL

PR #353, opened 2019-03-21
Fix problem saving to output directory

rogerthat94:
When images were open using a full filepath (either specified from the command line or from the dialogue that pops up in response to the "Open" or "Open Dir" QT buttons), they were not saved in the location that the user specified. Instead, they were being saved in the same directory as the image. This bug was caused by how Python's os.path.join() function handles full file paths given as the second argument. For example, os.path.join('first/path', '/second/full/path') will just return '/second/full/path'.

To recreate this problem, run "labelme --output path/to/output/directory --autosave /full/path/to/image.jpg" without the fix. Create a polygon on the image and click save. Notice that the save dialog opens up to "/full/path/to/". Even if you go to File->Change Output Dir and select a directory, the save dialog still opens up to "/full/path/to/". If you apply this fix and then repeat these steps, now the save dialog opens up to "--output path/to/output/directory". The same problem occurred when the "--autosave" flag was used. But this fix corrects this problem as well.

One issue that this change creates is when a directory that is being labeled contains subdirectory1/image.jpg and subdirectory2/image.jpg. This change would cause both of these images to use the label file "path/to/output/directory/image.json". However, I don't think this is a huge problem. This would already occur without the change if the directory being labeled contains image.jpg and image.gif. In addition, the program shows users existing annotations for a file. So this change should never cause someone to accidentally overwrite an existing annotation. People will see the annotations from subdirectory1/image.jpg when they go to annotate subdirectory2/image.jpg and realize that they either have to change the name of one of these files or not specify an output directory.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -660,7 +660,8 @@ class MainWindow(QtWidgets.QMainWindow):
         if self._config['auto_save'] or self.actions.saveAuto.isChecked():
             label_file = osp.splitext(self.imagePath)[0] + '.json'
             if self.output_dir:
-                label_file = osp.join(self.output_dir, label_file)
+                label_file_without_path = osp.basename(label_file)
+                label_file = osp.join(self.output_dir, label_file_without_path)
             self.saveLabels(label_file)
             return
         self.dirty = True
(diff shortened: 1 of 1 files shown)


wkentaro:
Nice! Thank you for the fix!

rogerthat94:
No problem! Thank you again for making this program!

PR #362, opened 2019-04-04
Fix missing fields

aboudev:
Converting annotations into [COCO format](http://cocodataset.org/#format-data):
- Add missing 'bbox' field.
- 'iscrowd' is 0 when using polygons to represent segmentation.


wkentaro:
Thanks!

PR #369, opened 2019-04-12
Exponential zooming and relative stickiness.

jonnedtc:
## Exponential zooming

Exponential zooming gives better control than linear zooming.
- Change `zoomWidget.value() + increment` to `zoomWidget.value() * increment`
- Change increment from -10 or 10 to 0.9 or 1.1
- Change minZoom from 1 to 10, because you can never zoom back in after reaching 1
- Change maxZoom from 500 to 1000, because zooming at this level goes faster now

## Scaling stickiness

Currently the 'snap-to-point' sensitivity or 'stickiness' is too high when zoomed in, but too low when zoomed out. The sensitivity should be relative to the size of the screen, not relative to the size of the image. When zooming in this would allow more control, while still remaining the 'snap-to-point' ability when zooming out.

- Dividing the sensitivity epsilon by the scale of the canvas to allow more 'snap-to-point' sensitivity when zoomed out and less sensitivity when zoomed in: `distance < (epsilon / scale)`
- The default epsilon is changed from 11.0 to 10.0 as this value exactly allows pixel perfect sensitivity at the new maxZoom level of 1000, while maintaining a good sensitivity when zoomed out

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -346,10 +346,10 @@ class MainWindow(QtWidgets.QMainWindow):
         )
         self.zoomWidget.setEnabled(False)
 
-        zoomIn = action('Zoom &In', functools.partial(self.addZoom, 10),
+        zoomIn = action('Zoom &In', functools.partial(self.addZoom, 1.1),
                         shortcuts['zoom_in'], 'zoom-in',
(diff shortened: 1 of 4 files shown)


wkentaro:
LGTM! Thanks.

PR #397, opened 2019-05-13
Add directory config options

caseymcc:
Add optional directory config file with directory options
  relative_paths: false #use relative path from image directory in load/saving label files
  label_directory: null #optional - relative directory from open directory to labels
  label_file_prefix: null #optional - prefix for label files
  label_file_postfix: null #optional - postfix for label files
  image_directory: null #optional - relative directory from open directory to labels
  output_directory: null #optional - output directory for labels, defaults to current directory

Altered save to use output_dir if available.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -15,6 +15,7 @@ from labelme import QT5
 
 from . import utils
 from labelme.config import get_config
+from labelme.config import get_directory_config
 from labelme.label_file import LabelFile
 from labelme.label_file import LabelFileError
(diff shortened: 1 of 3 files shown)


wkentaro:
Sorry, but I don't understand why this is so complicated, and why you need to add **default_** directory_config key to the config file.
The default should be already written there (config.yaml), so I don't understand what is your idea.

Could you please explain your motivation/issue, or please open an issue that describes your problem.

PR #405, opened 2019-05-15
Group operations on shapes

### LINKED ISSUE #404: Add group operation for shapes
Now only one shape can be moved (as well as deleted, duplicated and assigned by color) at once.
It would be very useful to be able to select several (or all) shapes and handle (move  etc.) them together. It is required at least then two images are almost the same apart then one is shifted relative to another.

I've made appropriate changes to make this enhancement

IlyaOvodov:
Fix #404

Several shapes can be selected at once (using Ctrl key) and handled together.

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -23,7 +23,7 @@ class Canvas(QtWidgets.QWidget):
     zoomRequest = QtCore.Signal(int, QtCore.QPoint)
     scrollRequest = QtCore.Signal(int, int)
     newShape = QtCore.Signal()
-    selectionChanged = QtCore.Signal(bool)
+    selectionChanged = QtCore.Signal(list)
     shapeMoved = QtCore.Signal()
     drawingPolygon = QtCore.Signal(bool)
     edgeSelected = QtCore.Signal(bool)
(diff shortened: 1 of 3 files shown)


labelme/widgets/canvas.py
```diff
@@ -315,49 +319,52 @@ def mousePressEvent(self, ev):
...
-            self.selectShapePoint(pos)
+            group_mode = (int(ev.modifiers()) == QtCore.Qt.ControlModifier)
+            self.selectShapePoint(pos, multiple_selection_mode=group_mode)
             self.prevPoint = pos
             self.repaint()
 
     def mouseReleaseEvent(self, ev):
         if ev.button() == QtCore.Qt.RightButton:
-            menu = self.menus[bool(self.selectedShapeCopy)]
+            menu = self.menus[min(len(self.selectedShapesCopy), 2)]
```
wkentaro: When I tried this. `len(self.selectedShapesCopy)` is always `0` because selection is reset in the `mouseSelectionEvent`.
How did you test this?

IlyaOvodov: SelectedShapeCopy is filled when you start to grag mouse, so you get "Move here" and "Copy here" only if you drag objects an drop them. Befeore it menu[0] seems to be normal.

wkentaro: Got it. Thanks.

PR #409, opened 2019-05-16
Fix #408 Undo works only once

### LINKED ISSUE #408: Undo works only once
When I press ctrl+Z first time it undoes the last editing, bun when I press it again, nothing heppens

IlyaOvodov:
(it worked only 1 time, then go in cycles on the latest backup)


wkentaro:
Merged via ba47546 Thanks!

PR #412, opened 2019-05-16
Group operations on shapes (#405) 

wkentaro:
Merge https://github.com/wkentaro/labelme/pull/405

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -302,14 +302,14 @@ class MainWindow(QtWidgets.QMainWindow):
         )
         editMode = action('Edit Polygons', self.setEditMode,
                           shortcuts['edit_polygon'], 'edit',
-                          'Move and edit polygons', enabled=False)
+                          'Move and edit the selected polygons', enabled=False)
 
-        delete = action('Delete Polygon(s)', self.deleteSelectedShape,
+        delete = action('Delete Polygons', self.deleteSelectedShape,
                         shortcuts['delete_polygon'], 'cancel',
-                        'Delete', enabled=False)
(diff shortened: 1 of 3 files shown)


wkentaro:
@IlyaOvodov I found that Ctrl-Z doesn't work if I move multiple polygons at the same time, and this code only handles less than 3 polygons, right?
Could you make these fixed?

IlyaOvodov:
I checked both and didn't find any problems either with handling 3,4,5 and even 6 polygons or with undoing it. Where did you find magic number "3"?

wkentaro:
Oh, sorry, maybe I was doing something wrong. When I tried again, it works without any problems.
One thing I'm not sure and why I thought about the magic number "3" is because of this line:
https://github.com/wkentaro/labelme/pull/412/files#diff-e970378747e9d6872c94dbcf67a84e62R334

Why did you need to add another menu to the canvas here?
https://github.com/wkentaro/labelme/pull/412/files#diff-e970378747e9d6872c94dbcf67a84e62R70

IlyaOvodov:
> Why did you need to add another menu to the canvas here?
Primaraly I expected that I will implement only Move for grope of shapes, but copy (and duplicate) will be available only if only 1 shape is selected (as I thought that it require some shape-specific uses dialog). So I made 2 different popup menus. But then I found that it is no problem to copy group of shapes as well, so these 2 menus became similar. But I decided not to remove this option, possibly it can be useful in a future. If You prefer to revert it, just correct 2 lines you pointed above (change 2 to 1 in the 1st one and revert the 2nd one) and also revert menu[2] initialization.

PR #416, opened 2019-06-07
Fix label editer bug

taashi-s:
Two fixes for around label editer bug.

## The not instance label proposal
Fix the bug that if the label is not instance label, the label proposal will be blank when adding a new shape.

## The instance label proposal when canceled.
Fix the bug that even if you cancel, the instance label will increase.


wkentaro:
Nice catch! Thanks!

PR #457, opened 2019-08-05
Update README.md

alexmyczko:


wkentaro:
Didn’t know that it will be possible to install via apt. Do you know who/how did this?

alexmyczko:
Yes I did :) 

https://packages.debian.org/labelme
https://packages.ubuntu.com/labelme

I'm glad you like it. We have some users in the photogrammetry of it, so we want it to be managable easy to install.

wkentaro:
Thank you. This will be super useful.

Could you let me know what kind of process is needed to 1) add a new package and 2) upgrade the package in ubuntu/debian?
Could you share some online documentations if you know?

alexmyczko:
Here's the official documentation: https://www.debian.org/doc/manuals/developers-reference/

And here's my CEO short summary: https://www.icloud.com/pages/0LWQ2LkV15Hvi49zYRwzS1AcA#Debian-Packaging

Adding a new package is via RFP/ITP bug report in Debian, I have a news tracker here:
http://bootes.ethz.ch/bts/

And updating a package also happens in Debian. Ubuntu just runs the archive-mirror stuff, and copies the packages from Debian. So I basically do nothing with Ubuntu, maybe with the launchpad.net bug reporters.

wkentaro:
(Sorry for late. I've been just back from the holiday.)
Thank you for a lot of precious information! Maybe I can manage to add some other package to apt at some point.

PR #461, opened 2019-08-09
Adding ability to remove a point from a closed shape (polygon).

hetfmt:
Making sure that "Save" is triggered with a point is added or removed.
Making minor change to ensure that "Save" is only triggered when a change has actually been made.


wkentaro:
Thanks 👍

PR #463, opened 2019-08-10
Disable 'add point to edge' feature for circle, line, point and rectangle

### LINKED ISSUE #462: adding points to circles and rectangles crashes
Hi, there is possibly a bug allowing to add points to rectangles and circles. When an attempt to add a point is made, the program crashes.

The issue can be easily reproduced, choose any rectangle or circle, right click on one of its points/vertices and try to add a point.

The menu option to add points to rectangles and circles is disabled when one clicks on their edges. It should probably behave equally when clicking on the shape points/vertices.

jakubgregorek:
Possible solution for #462.

Shape type is checked before the signal to allow adding points is emitted. The signal is emitted only if the shape type is polygon or linestrip (the only types where it is possible to add new points by right click on their edges).

The signal, originally called edgeSelected, was renamed to possibleToAddPoint because now it is not emitted every time an edge is selected.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -481,7 +481,7 @@ class MainWindow(QtWidgets.QMainWindow):
             onShapesPresent=(saveAs, hideAll, showAll),
         )
 
-        self.canvas.possibleToAddPoint.connect(self.actions.addPointToEdge.setEnabled)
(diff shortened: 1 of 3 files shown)


wkentaro:
Thanks! 👍

PR #493, opened 2019-10-02
Add support for double clicking labels

bartosz319:
Here's a simple patch that accepts the label selection widget when a label is double clicked. I found it useful and I think that the behavior is quite intuitive. What are your thoughts?


wkentaro:
thanks!

PR #495, opened 2019-10-04
coco conversion example update

imranparuk:
Skip image files that don't exist in coco conversion script


wkentaro:
why that happens?

PR #538, opened 2020-01-03
Add option for changing PIL.Image.MAX_IMAGE_PIXELS

### LINKED ISSUE #519: Is there any options for PIL.Image.MAX_IMAGE_PIXELS?
I want to work with large image.
However, when loading a large image, an error occurs with the following message.

> PIL.Image.DecompressionBombError: Image size (320461480 pixels) exceeds limit of 178956970 pixels, could be decompression bomb DOS attack.

This message is about PIL.Image.MAX_IMAGE_PIXELS.
I think it would be useful if user could change PIL.Image.MAX_IMAGE_PIXELS.

bganglia:
Fixes #519

New argument: "--max-image-pixels" (default = None)
New config setting: "max_image_pixels" (default = null)
New argument in LabelFile.load_image_file: "max_image_pixels" (default = None)

If the max_image_pixels argument is not None, then it will become the new MAX_IMAGE_PIXELS setting.

Note: when I ran labelme using a [large image](https://www.flickr.com/photos/88572252@N06/38400322472/in/pool-veryverylargephotos/), I only saw a warning, not an error. Still, this code should prevent the error as well.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1254,7 +1254,7 @@ class MainWindow(QtWidgets.QMainWindow):
         if QtCore.QFile.exists(label_file) and \
                 LabelFile.is_label_file(label_file):
             try:
-                self.labelFile = LabelFile(label_file)
+                self.labelFile = LabelFile(label_file, self._config["max_image_pixels"])
             except LabelFileError as e:
                 self.errorMessage(
                     self.tr('Error opening file'),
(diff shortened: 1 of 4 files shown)


bganglia:
It looks like the assertion in /tests/docs_tests/man_tests/test_labelme_1.py failed because my help string changed the help2man output.

What should I do?

wkentaro:
Please see https://github.com/wkentaro/labelme/pull/548, which fixes this issue.

PR #557, opened 2020-01-20
Fixed a bug when application crashes on polygon deletion

### LINKED ISSUE #556: Application crushes on polygon deletion
How to reproduce:

1. Select a polygon
2. Move the polygon
3. Do not move the mouse, do mouse right-click 
4. Select "Delete polygons"

Expected behavior: Polygon gets deleted
Observed behavior: Application crashes

The application crashes due to the `ValueError` happening on the following line in the function `mouseReleaseEvent` when the polygon was already deleted.

```
index = self.shapes.index(self.hShape)
```

Deeper investigation showed that at the moment of the crash both conditions are true:
`self.movingShape == True` and `self.hShape is not None`. It seems that `movingShape` only resets to `False` when the mouse moves.

sergeyshilin:
Closes #556

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -222,7 +222,6 @@ class Canvas(QtWidgets.QWidget):
             return
 
         # Polygon/Vertex moving.
-        self.movingShape = False
         if QtCore.Qt.LeftButton & ev.buttons():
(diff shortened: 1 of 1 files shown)


wkentaro:
Thanks!

Issue #522, opened 2019-12-11
set different color for different label


Ramlinbird:
Is there any way to set different color for my label list, now it's all green. Thanks a lot.

wkentaro:
Closed via https://github.com/wkentaro/labelme/pull/563

PR #564, opened 2020-01-25
Set different color for for different labels

grucin:
This change adds possibility to set color for specific label in `--labelcolors` option


grucin:
https://github.com/wkentaro/labelme/issues/522

wkentaro:
Closed via #563

grucin:
Great. There is a small issue with label edition. In this solution you set colors only in addLabel method but if user changes label, the color is not changing until you open this file again.

PR #565, opened 2020-01-25
Preserve other keys on shape level

grucin:
The other data on shape level is preserved in the same way as the other data on main level.


wkentaro:
Can you please resolve the conflicts?

grucin:
Conflicts resolved.

PR #566, opened 2020-01-25
Merge image flags from config and label file

grucin:
I've encountered the issue that flags from config don't show in the Flags widget in the interface if there is flags key in label file during loading the image. This small change fix this issue.


wkentaro:
Can you explain why you don't want to edit the config file? Or list all flags in the command line.

grucin:
I want to use flags from config or command line but when I use some image second time (with json file already created) only flags from json file are showed. It is because loadFlags method clear list of flags in widget on the beginning so you can't run  this twice in this place.

PR #625, opened 2020-03-16
Update readme for Windows install

DouglasLivingstone:
Since pillow version 7 works on Windows, there is no need to specifically install version 4 any more, so take this instruction out of the readme.

On Windows, the `source` command produces an error:

    'source' is not recognized as an internal or external command,
    operable program or batch file.

So update the setup instructions to invoke the `conda` binary instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,12 +140,13 @@ brew cask install wkentaro/labelme/labelme
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


wkentaro:
👍

DouglasLivingstone:
Thanks :-)

PR #626, opened 2020-03-20
Fixed edit error

pixx1:
When editing a label with an group ID and u delete it, labelme crashed:

Fixed error "TypeError: setData() takes no keyword arguments"

Deleting the keywords fixed this error


wkentaro:
Thanks! 👍

PR #627, opened 2020-03-21
Create labelme.desktop

alexmyczko:
should be installed to /usr/share/applications
so the gnome/kde menu systems have an icon (you'd also need to provide an icon, labelme.png (or .svg) and put it in /usr/share/icons)

### DIFF
--- a/labelme.desktop
+++ b/labelme.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+Name=Labelme
+Comment=Image Polygonal Annotation with Python
+Exec=labelme
+Icon=labelme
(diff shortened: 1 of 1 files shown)


wkentaro:
Thanks! That sounds great.

Are you thinking of installing the software with .deb file for the app and icon? Or with pip?

alexmyczko:
I already added it to the debian package, and Ubuntu and devuan, raspberry have or will get it by copying from debian.

I am only a pip user, so no idea if and how to add it there.

wkentaro:
Got it. Thanks!

PR #634, opened 2020-03-27
Fix for core dumped after creating any label

MrJavy:
Had to add a NULL parameter at line 46, probably due to version issue in Qt since there's a reimplementation of the subElementRect function.

See:
https://doc.qt.io/qt-5/qcommonstyle.html

Error log:
[me@my-pc folder]$ labelme
[INFO] Loading config file from: /home/myuser/.labelmerc (__init__.py:72)
qt5ct: using qt5ct plugin
qt5ct: D-Bus global menu: no
  File "/usr/lib/python3.8/site-packages/labelme/widgets/label_list_widget.py", line 46, in paint
    textRect = style.subElementRect(QStyle.SE_ItemViewItemText, options)
TypeError: subElementRect(self, QStyle.SubElement, QStyleOption, QWidget): not enough arguments
Aborted (core dumped)

### DIFF
--- a/labelme/widgets/label_list_widget.py
+++ b/labelme/widgets/label_list_widget.py
@@ -45,7 +45,7 @@ class HTMLDelegate(QtWidgets.QStyledItemDelegate):
 
         textRect = style.subElementRect(
             QStyle.SE_ItemViewItemText, options, None
-            )
+        )
 
         if index.column() != 0:
             textRect.adjust(5, 0, 0, 0)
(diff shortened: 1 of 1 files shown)


wkentaro:
Does this probably only happen with Python3.8?

MrJavy:
I think it must be something relative to Qt

MrJavy:
Well, I thought it could be a Qt thing since it has a reimplementation of the method for default values. I'm not an expert in the field, so I can't give you a definitive response. Let me help giving information from my Qt packages.

Using:   python3.8

Package                       Version     
----------------------------- ------------
PyQt5                           5.14.1      
PyQt5-sip                     12.7.1      
QtPy                             1.9.0

PR #649, opened 2020-04-22
avoid error -- cannot write mode P as JPEG

riven314:
an error "cannot write mode P as JPEG" would occasionally occur when reading .jpg image into GUI
it's a simple fix to avoid that.

### DIFF
--- a/labelme/label_file.py
+++ b/labelme/label_file.py
@@ -34,7 +34,7 @@ class LabelFile(object):
     @staticmethod
     def load_image_file(filename):
         try:
-            image_pil = PIL.Image.open(filename)
+            image_pil = PIL.Image.open(filename).convert('RGB')
         except IOError:
             logger.error('Failed opening image file: {}'.format(filename))
             return


wkentaro:
I can't reproduce this with below image
https://upload.wikimedia.org/wikipedia/commons/thumb/c/cf/Coleen_Gray_in_Kansas_City_Confidential.jpg/440px-Coleen_Gray_in_Kansas_City_Confidential.jpg

riven314:
you can try this image: [https://drive.google.com/file/d/1Qc061TR-PSmW15ut5jHj8oo7r8a2cOEo/view?usp=sharing](url)
put it into data_annotated folder, and then run ```$ labelme data_annotated --labels .\labels.txt --nodata --autosave```

without the change I made, I got the following error:
```
  File "c:\users\u3509\anaconda3\envs\brand-sim\lib\site-packages\PIL\JpegImagePlugin.py", line 612, in _save
    rawmode = RAWMODE[im.mode]
KeyError: 'P'

During handling of the above exception, another exception occurred:

  File "c:\users\u3509\anaconda3\envs\brand-sim\lib\site-packages\labelme\app.py", line 944, in fileSelectionChanged
    self.loadFile(filename)
  File "c:\users\u3509\anaconda3\envs\brand-sim\lib\site-packages\labelme\app.py", line 1291, in loadFile
    self.imageData = LabelFile.load_image_file(filename)
  File "c:\users\u3509\anaconda3\envs\brand-sim\lib\site-packages\labelme\label_file.py", line 55, in load_image_file
    image_pil.save(f, format=format)
  File "c:\users\u3509\anaconda3\envs\brand-sim\lib\site-packages\PIL\Image.py", line 2134, in save
    save_handler(self, fp, filename)
  File "c:\users\u3509\anaconda3\envs\brand-sim\lib\site-packages\PIL\JpegImagePlugin.py", line 614, in _save
    raise OSError("cannot write mode %s as JPEG" % im.mode)
OSError: cannot write mode P as JPEG
```

wkentaro:
I cannot open the image.
image

wkentaro:
I needed to rename to testing.png, so this error is because of the wrong file extension.

riven314:
I am using windows and I can open the image. 
If the problem comes from wrong file extension, would it be better if the codebase can also handle that by enforcing it to be RGB mode (from RGBA mode)?

wkentaro:
I'm not sure. Because in some cases, we want to keep the alpha channel. And I don't think that kind of auto-regression of the file extension is a good idea. Users should be aware of the file extension and its content.

PR #654, opened 2020-04-28
Fix segmentation generation for rectangles

lolipopshock:
If the shape is a rectangle, `points = np.asarray(points).flatten().tolist()` won't appropriately construct a segmentation vertex list.


YacobBY:
@lolipopshock thank you very much now my rectangle annotations work with Detectron2!

If possible it would be great if someone could look at circular annotations as well, those don't seem to be compatible with their coco format as well but I'm not sure if they support it in the first place

Issue #299, opened 2019-01-24
Error opening file


zhulf0804:
Label in windows and open in mac, error happens: 

Error opening file
'utf8' codec can't decode byte 0xb3 in position 0: invalid start byte

How to resolve the above problem?
thx

paidaxingNUDT:
Do you have any solutions?

northeastsquare:
labelme in windows is gb2312 encoding;
while in ubuntu , its ut-8

jjjjohnson:
@northeastsquare Is there any solution for using both gb2312 and utf-8?

Ruolingdeng:
@jjjjohnson @northeastsquare Have you solved the problem? I met with the same problem. I annotated my images using labelme software in windows system. And now I want to use the json file to do training in Ubuntu 16.04, but it occurred "FileNotFoundError: [Errno 2] No such file or directory: 'datasets/coco/train2017/..\\ResizeImage\\000010.jpg'." When I used the labelme software in Ubuntu system to open the json file, it occurred "make sure /home/xbd/000000.json is a valid label file" Could anybody help me, please? I will be very appreciated with you.

TomHeaven:
I find a workaround: Open `label_file.py` and change line 75 to
```python3
 #with open(filename, 'rb' if PY2 else 'r') as f:
with open(filename, encoding='gb2312') as f: # To read windows labels.
```
The error is caused by different default encoding used by Windows and Linux/Mac OS when opening a file. I'll offer a PR to fix that later.

TomHeaven:
This issue should be fixed by PR https://github.com/wkentaro/labelme/pull/676.

Bjoux2:
> This issue should be fixed by PR #676.
withi this:
 #with open(filename, 'rb' if PY2 else 'r') as f:
with open(filename, encoding='gb2312') as f: # To read windows labels.

I got the result:

Error opening file
Open() got an unexpected keyword argument 'encoding'

PR #676, opened 2020-06-09
Fix label encoding error across platforms (#299)

TomHeaven:
This PR aims to solve issue https://github.com/wkentaro/labelme/issues/299. On Windows, files are saved with GB2312 as default encoding. However, utf-8 is the default encoding on Linux and macOS. With previous version of labelme, if you label images on Windows and open the label json files on Linux or macOS, you will encounter issue https://github.com/wkentaro/labelme/issues/299.

This PR solve the issue by setting the default encoding of label files as utf-8. Therefore, no matter on what OS, you will not meet https://github.com/wkentaro/labelme/issues/299 using the new version.

Besides, this PR offers an extra option "--label_encoding" to make Labelme compatible with previous json files of labels. For instance, if you make labels on Windows using previous versions of Labelme and need to open them on macOS, simply start this version by
```
labelme --label_encoding gb2312
```
and it should work like a charm.


wkentaro:
Do you have any example windows file for testing?

TomHeaven:
Yes, I do. In fact, i've already tested the PR with those files, and it works.

wkentaro:
I mean when https://github.com/wkentaro/labelme/pull/683 is not enough?
Also, you can see the CI is not passing with Python2.

TomHeaven:
I'll email u some data later. I think it's time to give up Python 2 since it is deprecated for half a year.

wkentaro:
I found the below command works for old files.
```
iconv -f gb18030 -t utf-8 1.json > 1_utf8.json
```

Bjoux2:
> This PR aims to solve issue #299. On Windows, files are saved with GB2312 as default encoding. However, utf-8 is the default encoding on Linux and macOS. With previous version of labelme, if you label images on Windows and open the label json files on Linux or macOS, you will encounter issue #299.
> 
> This PR solve the issue by setting the default encoding of label files as utf-8. Therefore, no matter on what OS, you will not meet #299 using the new version.
> 
> Besides, this PR offers an extra option "--label_encoding" to make Labelme compatible with previous json files of labels. For instance, if you make labels on Windows using previous versions of Labelme and need to open them on macOS, simply start this version by
> 
> ```
> labelme --label_encoding gb2312
> ```
> 
> and it should work like a charm.


withi this:
#with open(filename, 'rb' if PY2 else 'r') as f:
with open(filename, encoding='gb2312') as f: # To read windows labels.

I got the result:

Error opening file
Open() got an unexpected keyword argument 'encoding'

PR #692, opened 2020-06-20
Drop Python 2.7

wkentaro:


wkentaro:
pytest-qt fails in CI with pyproject.toml.

PR #738, opened 2020-08-01
impossible to zoom in when the zoom value is less than 10.

ohkawa3:
This is a fix for the problem where addZoom cannot be used to zoom in when the zoom value is less than 10.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1331,7 +1331,13 @@ class MainWindow(QtWidgets.QMainWindow):
         self.zoom_values[self.filename] = (self.zoomMode, value)
 
     def addZoom(self, increment=1.1):
-        self.setZoom(self.zoomWidget.value() * increment)
+        difference_value = self.zoomWidget.value() * (1 - increment)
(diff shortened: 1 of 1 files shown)


wkentaro:
Where does the 10 come from?

wkentaro:
`self.zoomWidget.value()` can range 1 to 1000.

ohkawa3:
10 depends on the argument "increment" of addZoom(self, increment=1.1).
For example, if the current value of self.zoomWidget.value() is 9 and the argument "increment" is 1.1, then self.setZoom(9.9) will be executed.
However, the value of self.zoomWidget.value() remains unchanged at 9 because the decimal point is truncated inside self.setZoom.
This fix forces the absolute value of the change to be at least 1.

ohkawa3:
After opening the image, please press "Ctrl+-" at least 20 times to fully shrink the image.
Then pressing "Ctrl++" will not enlarge the image.
So we need to press "Ctrl+0" to return the image to its original size.

wkentaro:
Will be fixed via #740

PR #751, opened 2020-08-14
I fix some bugs of labelme, please check them

ILXR:
1. labelme crashed when removing points.  This is because the length of the points list of this (point type) shape is 0 when the point is deleted, so it would cause an exception.
2. when you have deleted all points of a shape, this shape wouldn't be removed. I added this feature.
3. when labelme read a json file, it will add all shapes in the list, and i filter those shape which has no points, so when you save the file, these empty shapes won't be saved
Finally, hope you accept my pull request, thank you fou your work on labelme


wkentaro:
Moved to https://github.com/wkentaro/labelme/pull/751

PR #752, opened 2020-08-14
Add functionality to change point size using the config file

jbutle55:
This allows a user to change the point size to be larger or smaller than the default size 8. This is done through the config file. Useful for annotating images with very small objects, which is why I made these changes to my own version of labelme in the first place.


jbutle55:
Failed tests, will fix and redo.

jbutle55:
Successfully ran through tests on my own computer, reopening.

PR #761, opened 2020-08-21
fix #760

### LINKED ISSUE #760: [BUG] masks generated from LineStrip have some notches
The mask generated from https://github.com/wkentaro/labelme/blob/master/labelme/utils/shape.py#L19 will have some notches as the following image shows:
image

And I think we can refer to this question from [stackoverflow](https://stackoverflow.com/questions/45172116/fix-pil-imagedraw-draw-line-with-wide-lines) to solve this problem.

nico-zck:
fix #760 that mask will have notches at turning point when drawing LineStrip annotation.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import functools
+import math
 import os
 import os.path as osp
 import re
@@ -1331,7 +1332,12 @@ class MainWindow(QtWidgets.QMainWindow):
         self.zoom_values[self.filename] = (self.zoomMode, value)
 
     def addZoom(self, increment=1.1):
-        self.setZoom(self.zoomWidget.value() * increment)
+        zoom_value = self.zoomWidget.value() * increment
+        if increment > 1:
+            zoom_value = math.ceil(zoom_value)
+        else:
+            zoom_value = math.floor(zoom_value)
+        self.setZoom(zoom_value)
 
     def zoomRequest(self, delta, pos):
         canvas_width_old = self.canvas.width()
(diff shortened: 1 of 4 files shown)


wkentaro:
Can you separate the drawing and deleteLine and deletePoints functions?

nico-zck:
> Can you separate the drawing and deleteLine and deletePoints functions as different PRs?

I don't know what should "deleteLine" and "deletePoints" do, my PR only fill circle to the inflection point of line.

labelme/widgets/canvas.py
```diff
@@ -376,6 +376,19 @@ def mouseReleaseEvent(self, ev):
...
+            ):
+                # Add point to line if: left-click + SHIFT on a line segment
+                self.addPointToEdge()
+        elif ev.button() == QtCore.Qt.LeftButton and self.selectedVertex():
+            if (
+                self.editing()
+                and int(ev.modifiers()) == QtCore.Qt.ShiftModifier
+            ):
+                # Delete point if: left-click + SHIFT on a point
+                self.removeSelectedPoint()
```
wkentaro, line 391: I was trying to indicate these changes.

labelme/app.py
```diff
@@ -1331,7 +1332,12 @@ def setZoom(self, value):
         self.zoom_values[self.filename] = (self.zoomMode, value)
 
     def addZoom(self, increment=1.1):
-        self.setZoom(self.zoomWidget.value() * increment)
+        zoom_value = self.zoomWidget.value() * increment
+        if increment > 1:
+            zoom_value = math.ceil(zoom_value)
+        else:
+            zoom_value = math.floor(zoom_value)
+        self.setZoom(zoom_value)
```
wkentaro, line 1340: This one as well.

wkentaro:

Thanks @nico-zck for spotting this and for the fix. The notch problem is real
and still reproduces on `main`, so the fix lives on in #2124 with you credited
as the original reporter/author.

I'm closing this PR because (a) it has merge conflicts after several years of
drift, and (b) it bundles three unrelated changes (zoom rounding, shift-click
add/remove vertex, re-enabling Black in CI) alongside the #760 fix, which is why
it couldn't be merged as-is. The linestrip notch fix will be reimplemented
cleanly against the current code with a regression test in #2124.

PR #763, opened 2020-08-26
Disable auto snapping while pressing key 'K'

jbutle55:
The changes to ci.yml and labelme.spec are to address the CI failures due to imgviz requiring matplotlib. I had to specify to use a version of matplotlib older than 3.3.0 due to the MATPLOTLIBDATA environment variable being depreciated and removed in 3.3.0. As well, I had to add the a.binaries line in the spec file due to failures occurring in the MacOS pyqt5 CI due to conflicting library verisons.

The changes to canvas.py, shape.py, and the default_config file are to add features to make labelling extra small objects easier. These changes allow for the size of the point markers when creating polygons to be changed from the default size of 8 to other values. 

As well, I added the ability to hold down the F key when drawing a polygon to disable the auto-close feature. Once released, the auto-close feature automatically starts again. This again made labelling extra-small objects easier.

Feel free to make any suggestions or corrections!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
 
     - name: Test with pytest
       shell: bash -l {0}
-      if: matrix.os != 'windows-latest'
+      if: matrix.os != 'windows-latest' && matrix.PYTEST_QT_API != 'pyqt4v2'
       env:
         PYTEST_QT_API: ${{ matrix.PYTEST_QT_API }}
       run: |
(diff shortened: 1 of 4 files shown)


jbutle55:
I brought this branch up to date with the upstream master.

wkentaro:
I think now CI is fixed. Can you remove the changes on ci.yml?

jbutle55:
The pull request now doesn't change ci.yml at all. I also removed a change to label.spec so that shouldn't change either.

wkentaro:
Maybe you need to run `git rebase origin/master`. (be careful as it might delete your changes if you make some mistakes)

jbutle55:
If I understood what you meant correctly, I used rebase interactively to remove any of the old commits that impacted ci.yml. Hopefully that's what you had in mind? I also ran the black/pytest tests on my own environment and all tests passed.

wkentaro:
Looks something is wrong. I’ll have a look

wkentaro:
It seems "Alt" key is a common key to disable snapping (powerpoint, affinity designer) although photoshop seems using "Ctrl/Command" which would be commonly used for duplicating.

PR #768, opened 2020-09-08
Fix PyInstaller around matplotlib

wkentaro:


wkentaro:
Fixed via https://github.com/wkentaro/labelme/pull/771

PR #776, opened 2020-09-20
Do not run as root in docker + add to docker infrastructure

bergercookie:
When running in docker, run under a standard user (not root). Also set the UID:GID of that user to be the same as the corresponding user on the host machine

On top of that, instead of instructing pip to install labelme from github, it uses
the locally downloaded version with any changes the user has made.

Building the container is done using `make -f Makefile-docker build` while
running it is `make -f Makefile-docker run-linux` if on a Linux machine,
otherwise `make -f Makefile-docker run-osx` on a Mac.


wkentaro:
I'm not sure if this is something really needed. What do you think?

bergercookie:
Hi Kentaro! :)

Yeah, I think this is useful.
If one runs this inside docker to edit files that exist on their filesystem, they wouldn't want to change the file's user:group to root.

Dropping to the standard user's privilages ensures that no changes to the permissions and ownership of the edited file will be made.

The Makefile added is just a convenience for one to run the docker without having to specify long one-liners.

In any case It's been a long time since I made this PR, feel free to close it if you think it's not necessary.

wkentaro:
I agree this would be useful for advanced users, but it would be better for most people to use the system Python or virtual environment to avoid troubles with docker. As this repo is mainly used by non-developers, probably it is less confusing not giving this option for users to use this software. 

Thanks for the contribution, anyway!

PR #801, opened 2020-11-11
Fix crash when opening pre/next image

InSightSuen:
Labelme may crash when opening pre/next image by shortcuts and dismissing LabelDialog by clicking OK button at same time.

After opened new file, app will clear canvas.shapes before canvas.setLastLabel.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1287,38 +1287,44 @@ class MainWindow(QtWidgets.QMainWindow):
 
         position MUST be in global coordinates.
         """
-        items = self.uniqLabelList.selectedItems()
-        text = None
-        if items:
-            text = items[0].data(Qt.UserRole)
-        flags = {}
-        group_id = None
-        if self._config["display_label_popup"] or not text:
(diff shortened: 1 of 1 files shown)


wkentaro:
Cannot be fixed in a more fundamental way? Like locking while label dialog is open.

wkentaro:

Thanks for this fix, @InSightSuen, and apologies for the long silence on it.

Since this PR was opened, `app.py` and the canvas have been substantially rewritten, so the patch no longer applies (the `newShape` / `uniqLabelList` / `setLastLabel` APIs it touches are gone).

More importantly, the race it addresses is no longer reachable: the label dialog is now application-modal (`exec_()`), so the main window receives no next/prev-image shortcut events while the dialog is open, and the labeling path runs synchronously with no event processing in between. There's no longer a window where navigation can clear the canvas mid-labeling.

I'm going to close this as no longer applicable. Thank you again for catching the original issue and taking the time to send a fix.

PR #803, opened 2020-11-17
add support to read label files in the name of image filename (helps …

umutkucukaslan:
If the filename of image and label files has changed and there is a label file, then the tool tries to load the image using the imagePath inside JSON file, which is not valid anymore. In this case, we can use the image filename provided before creating an instance of LabelFile class by passing the filename argument as an option to this class.


wkentaro:
I feel like this just makes things complicated e.g., what happens when you change names of multiple files?
What do you think?

Issue #716, opened 2020-07-07
[Feature]


yurikleb:
I find the "Circle" as a very useful annotation tool as I'm dealing with mostly round objects in my data.

Is it possible to export the circle annotations as VOC or COCO format dataset?

I saw those scripts:
https://github.com/wkentaro/labelme/blob/master/examples/instance_segmentation/labelme2voc.py
https://github.com/wkentaro/labelme/blob/master/examples/instance_segmentation/labelme2coco.py

Which seem to convert Instance and Semantic segmentations to VOC and COCO.

Do those script work with circle annotations?

here is how one of those labels looks in the JSON created by labelme:
```   
 {
      "label": "cap",
      "line_color": null,
      "fill_color": null,
      "points": [
        [
          129,
          161
        ],
        [
          141,
          172
        ]
      ],
      "shape_type": "circle"
 }
```

yurikleb:
Hey, after exchanging a few emails with the team at roboflow.com they now support LabelMe JSON format, which makes it really easy to convert to any other format:
https://roboflow.com/formats/labelme-json

Issue #537, opened 2020-01-02
Support Circle shape in labelme2coco.py JSON generation


stephanj:
The current labelme2coco.py doesn't support the circle shape.

stephanj:
You can achieve this by adding following logic to labelm2coco.py:

```
        # ...
            if label in masks:
                masks[label] = masks[label] | mask
            else:
                masks[label] = mask

            # Support for circle shape by Stephan Janssen (sja@devoxx.com)
            if shape_type == "circle":
                
                # Create polygon shaped based on connecting lines from/to following degress
                bearing_angles = [0, 15, 30, 45, 60, 75, 90, 105, 120, 135, 150, 165, 
                180, 195, 210, 225, 240, 255, 270, 285, 300, 315, 330, 345, 360]
                            
                orig_x1 = points[0][0] 
                orig_y1 = points[0][1] 

                orig_x2 = points[1][0]
                orig_y2 = points[1][1]

                # Calculate radius of circle
                cx = (orig_x2 - orig_x1)**2
                cy = (orig_y2 - orig_y1)**2
                radius = math.sqrt(cx + cy)

                circle_polygon = []
            
                for i in range(0, len(bearing_angles) - 1):
                    ad1 = math.radians(bearing_angles[i])
                    x1 = radius * math.cos(ad1)
                    y1 = radius * math.sin(ad1)
                    circle_polygon.append( (orig_x1 + x1, orig_y1 + y1) )

                    ad2 = math.radians(bearing_angles[i+1])
                    x2 = radius * math.cos(ad2)  
                    y2 = radius * math.sin(ad2)
                    circle_polygon.append( (orig_x1 + x2, orig_y1 + y2) )

                points = circle_polygon

            else:
                points = np.asarray(points).flatten().tolist()
            
            segmentations[label].append(points)                

        for label, mask in masks.items():
        # ...
```

pieterbl86:
@stephanj thank you for this good addition!!! I would recommend to add one additional line of code at the end of the circle procedure, as this line will flatten the array in such a way that it corresponds to the visualization method of the cocoapi. Again, thank you for this addition. 

> You can achieve this by adding following logic to /examples/instance_segmentation/labelme2coco.py:
> 
> ```
>         # ...
>             if label in masks:
>                 masks[label] = masks[label] | mask
>             else:
>                 masks[label] = mask
> 
>             # Support for circle shape by Stephan Janssen (sja@devoxx.com)
>             if shape_type == "circle":
[...]

PR #804, opened 2020-11-19
Add support circle shap_type on label2coco.py

firesh:
See issue #537 #716


firesh:
It seems CI not working well.

PR #819, opened 2021-01-10
nam: modify some features

namcusamlc:
- Added predefined labels to labelme/config/default_config.yaml

- Added predefine objects

- Added highlight by object number

- Added modification of select line width in yaml config file

- Same object number has the same color

- load predefined data from JSON file

- add exporting data to json in the right format

- add load video


namcusamlc:
error

PR #822, opened 2021-01-11
Added Constant Tracing Creation Mode

psomers3:
Since I'm not a fan of clicking a ton of times, I added a "trace" drawing mode that continuously places points as you move the mouse (keyboard shortcut Ctrl+T). Just left click once to start recording and left click again to stop. I would like to test this with a touch screen sometime (I don't think it'll work right this second), but I have an idea of how to make it work.

I also cleaned up the toggleDrawMode function.


psomers3:
have tested it with a touchscreen laptop and it works.

psomers3:
Last commit adds the point size change from #763 and an additional default smaller point size for tracing because it is obnoxious to have a large point size with as many points that are created when tracing.

psomers3:
@jbutle55
I added the hold F option similar to you for pausing the addition of points during tracing so that you can zoom/pan. We can probably make this the same variable and feature for the addition you wanted.

PR #834, opened 2021-02-03
Fix bug: After deletingFile, the mouse moves into the canvas and the program crashes.

MrWindWz:


wkentaro:

Thanks for this, and apologies for the very long delay.

The crash this PR targets (program crashing when the mouse enters the canvas after **Delete File**) no longer reproduces on `main`. The canvas has since been rewritten: after deleting a file the canvas resets to an empty `QPixmap` (rather than a `None` pixmap), and `mouseMoveEvent` is now guarded, so moving the mouse over the empty canvas is a no-op instead of a crash. Verified manually: delete the label file, move the mouse over the canvas — no crash.

The PR also bundles several unrelated changes (a `ConfigInfo` refactor, a `labelList.clicked -> edit mode` behavior, and a `label_dialog` change that makes `getGroupId()` return a string instead of an int), and it no longer merges cleanly against the current code.

Since the crash is resolved, I'm closing this. If you still hit a crash on the latest `main`, please open a new issue with steps to reproduce. Thanks again for reporting and digging into it.

PR #838, opened 2021-02-22
Add config option to show polygon labels on canvas

maxkarelov:
There are a new boolean config option `shape.show_labels` to draw label names over polygons for better visual labelling and `default_create_mode` for default type of labelling objects.

In our team we have the case where we continuously improve model by using it's predictions on unknown pictures and review it in labelme to speed up labelling process.

**Before**
Untitled before

**After**
Untitled


maxim-karelov-incountry:
@wkentaro could you please take a look?

wkentaro:
Thank you for the nice feature implementation. I'm happy to include this.

Can you have a look at the below results? Maybe we need to have a better algorithm to define the x,y coordinate of the label?
```
labelme examples/bbox_detection/data_annotated --config '{"shape": {"show_labels": true}}'
```
image

wkentaro:

Thank you for this contribution, and apologies it sat for so long. The "show labels on the canvas" idea is genuinely useful, especially for the model-review workflow you described.

Unfortunately this PR can no longer be merged as-is: the codebase has been substantially refactored since 2021 (the shape painting and config handling have moved/changed), so the diff no longer applies, and the original review feedback about label placement (the overlapping/offset text) was never resolved. The PR also bundles an unrelated change to the file-deletion confirmation that we don't want to carry over.

Rather than rebuild it inside this stale branch, I've captured the feature as a fresh, fully-specified issue so it can be implemented cleanly against the current code (including the new Settings dialog): #2127. The `default_create_mode` option is intentionally left out of that scope for now.

Closing this in favor of #2127. Thanks again for proposing the feature.

PR #843, opened 2021-03-09
opening a directory now discovers json files without images

hawkeye-will:
Currently if you open a directory in labelme it won't show annotation JSONs without an image. 

Often we delete the images when working with 10GB+ of annotation files as it reduces the storage needed. But every time we want to go through a directory of annotations and check them we've got to extract the image from the "imageData" field so we can go through a folder of annotations. 

This will mean that it finds non-labelme json files in the directory; but that's already handled and you get a pop-up asking you to check if its a valid label file.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -2050,6 +2050,8 @@ class MainWindow(QtWidgets.QMainWindow):
             for fmt in QtGui.QImageReader.supportedImageFormats()
         ]
 
+        extensions += [".json"]
+
         images = []
         for root, dirs, files in os.walk(folderPath):
             for file in files:


JAYCHOU2020:
> Currently if you open a directory in labelme it won't show annotation JSONs without an image.
> 
> Often we delete the images when working with 10GB+ of annotation files as it reduces the storage needed. But every time we want to go through a directory of annotations and check them we've got to extract the image from the "imageData" field so we can go through a folder of annotations.
> 
> This will mean that it finds non-labelme json files in the directory; but that's already handled and you get a pop-up asking you to check if its a valid label file.

Why can't I load a set of json files even though I used your modified code?

wkentaro:

Thanks for this contribution! Unfortunately this PR now has merge conflicts because `scanAllImages` was refactored into `_scan_image_files` in main. PR #1143 addresses the same directory-scan-for-JSON use case with a more complete implementation. Given the age of this PR and the conflict, I'm closing in favour of #1143. Feel free to revisit if you have ideas that aren't covered there.

PR #860, opened 2021-04-20
File search after pushing enter key

steven0129:
This commit is for performance reason. When thousands of files is searched with textChanged, it would be slow. I suggested use returnPassed event instead of textChanged event.


steven0129:
Cancel this pull request

nai-kon:
@steven0129 
Hello, I'm also struggling with current slow search.
Could you please continue the pull request?
Thank you.

PR #863, opened 2021-04-29
 Added train_test_split json script

GabrielLima1995:
It was  created a script that allow you input the json and it gives json test and json train.


wkentaro:
I think this is out of the scope of this project.

PR #877, opened 2021-06-27
Add batch_convert.py under batch_convert_json_to_dataset directory

zzowenzz:
I added a simple python file that can batch convert all the .json files manually labeled by a polygon.
Prerequisite:
- [ ] All the .json files should be placed under the same directory.
- [ ] The output will also be placed at the same place as those .json files.


wkentaro:
I think this is out of the scope of this project.

PR #903, opened 2021-08-05
Display a line length in pixels for measuring reference objects

healthonrails:


wkentaro:

Thanks for the contribution! Unfortunately this PR can no longer be merged as-is: it has merge conflicts with current main, there are no CI checks, and the implementation hardcodes font size (25pt Arial) and always shows the label with no way to toggle it. If you'd like to resubmit, please rebase on main, make the display optional (e.g. behind a flag or only on hover/selection), and use the project's existing text-rendering conventions. If the feature is still relevant, a fresh PR against current main would be the way forward.

Issue #871, opened 2021-06-01
[adding an ellipse with rotation]


ShivaKothuru:
I've implemented an ellipse with three points. but I was trying to rotate it. any help in this regard would be great.
Thanks!

ShivaKothuru:
Done!

Issue #391, opened 2019-05-10
create rotated ellipse


zfb-blog:
I've referred issue [#298](https://github.com/wkentaro/labelme/issues/298), and tried to add ellipse. It is successful, but only can get a vertical or horizontal one. Is there a way to rotate it?

I'm defining the bounding rectangle by three points, one define center, second define left border (which can calc the theta of rotation angle), third define the height.

wkentaro:
For annotating a rotated ellipse, I guess you only need 2 points.

1. First point is the center.
2. Second point is half of the longest axis.

So I think you don't need to rotate, but you need to define ellipse like above.

RSKothari:
@wkentaro Could you perhaps provide an example of annotating a rotated ellipse in the repo?

ShivaKothuru:
@zfb-blog I added an ellipse like you. I am now trying to rotate it. could you let me know if you've already implemented rotating an ellipse?

Issue #298, opened 2019-01-19
Guideline for adding new label shapes and enabling rotation?


jonbakerfish:
Thanks for open sourcing this handy tool! 
I want to add some new label shapes (e.g., ellipse, triangle) and enable the user to rotate the labels. Could you shed some light on the overall workflow of the program and which part of the code should be changed?

wkentaro:
This is a logic to add new shape on canvas: https://github.com/wkentaro/labelme/blob/86bec5e0d7f48bff018696e898b18d158adb3d70/labelme/widgets/canvas.py#L178-L192
to finalize shape drawing on canvas: https://github.com/wkentaro/labelme/blob/86bec5e0d7f48bff018696e898b18d158adb3d70/labelme/widgets/canvas.py#L287-L309

Also I think you should search `createMode` in `canvas.py` to see how the program behaves differently for different shapes.

PR #909, opened 2021-08-25
Add support for rotated ellipses defined via ellipse fitting method

colluca:
**Description**

This pull request addresses issues #871, #391 and #298.

While rotated ellipses require at most 5 points to be defined, it is useful to define ellipses based on ellipse fitting techniques applied on a greater number of points to more easily and better approximate elliptical contours in images.

The implementation proposed here is based on the [EllipseModel](https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.EllipseModel) class provided in the scikit-image library which returns exact ellipses when four or five points are provided, and approximating ellipses when more points are given. To finalize an elliptical shape, being the number of points required undefined, a double mouse click is required (as used also for polygonal shapes).

**Issues**

The biggest issue is in supporting **rotated** ellipses.

For example, the [makePath()](https://github.com/wkentaro/labelme/blob/4efcad25d91b4950364afacf1da88bd9e7abe367/labelme/shape.py#L242-L257) function is not easily extended as the QPainterPath object does not support rotated ellipses.

Similarly, it is unclear if and how the [boundingRect()](https://github.com/wkentaro/labelme/blob/4efcad25d91b4950364afacf1da88bd9e7abe367/labelme/shape.py#L259-L260) function should be extended, either depending on [makePath()](https://github.com/wkentaro/labelme/blob/4efcad25d91b4950364afacf1da88bd9e7abe367/labelme/shape.py#L242-L257) or customized.
For example, you could find the bounding box around the rotated ellipse, but this is non-trivial and is a worse approximation than a rotated bounding box. So it is probably easier to customize the [calculateOffsets()](https://github.com/wkentaro/labelme/blob/4efcad25d91b4950364afacf1da88bd9e7abe367/labelme/widgets/canvas.py#L481) function which invokes [boundingRect()](https://github.com/wkentaro/labelme/blob/4efcad25d91b4950364afacf1da88bd9e7abe367/labelme/shape.py#L259-L260) directly.

For the moment, I have marked these functions as unsupported by raising an error through the logger when they occur, e.g.:
https://github.com/bigcola317/labelme/blob/df8ceab1949ec3949e4609af4c4d6949acee7f37/labelme/shape.py#L312-L317
In my tests I wasn't able to excite these conditions, but I await feedback.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -344,6 +344,14 @@ class MainWindow(QtWidgets.QMainWindow):
             self.tr("Start drawing circles"),
             enabled=False,
         )
+        createEllipseMode = action(
+            self.tr("Create Ellipse"),
+            lambda: self.toggleDrawMode(False, createMode="ellipse"),
+            shortcuts["create_ellipse"],
+            "objects",
+            self.tr("Start drawing ellipses"),
+            enabled=False,
+        )
         createLineMode = action(
             self.tr("Create Line"),
             lambda: self.toggleDrawMode(False, createMode="line"),
@@ -599,6 +607,7 @@ class MainWindow(QtWidgets.QMainWindow):
             editMode=editMode,
             createRectangleMode=createRectangleMode,
             createCircleMode=createCircleMode,
+            createEllipseMode=createEllipseMode,
             createLineMode=createLineMode,
             createPointMode=createPointMode,
             createLineStripMode=createLineStripMode,
@@ -633,6 +642,7 @@ class MainWindow(QtWidgets.QMainWindow):
                 createMode,
                 createRectangleMode,
                 createCircleMode,
+                createEllipseMode,
                 createLineMode,
                 createPointMode,
                 createLineStripMode,
@@ -651,6 +661,7 @@ class MainWindow(QtWidgets.QMainWindow):
                 createMode,
                 createRectangleMode,
                 createCircleMode,
+                createEllipseMode,
                 createLineMode,
                 createPointMode,
                 createLineStripMode,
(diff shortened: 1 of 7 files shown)


maxchiron:
Dear sir, after using your branch with scikit ellipsis, I've encountered trouble when turning labelme .json to mapping image used for training. Those ellipsis points seems to be treated as polygon points though. Could u please ignite how to face with this challenge, please? Thanks a lot.

wkentaro:

Thanks for this, and apologies it sat so long.

We've decided not to add an ellipse / rotated-ellipse shape, so I'm closing this as `wontfix`. The reasoning, in short:

- **Demand is thin and stale.** Ellipse has come up three times across the project's history (#298, #391, #871), all back in 2019–2021, with nothing since. This PR offered a working implementation on a branch for ~5 years and drew a single user comment. Latent demand that doesn't convert even with an implementation available is weak demand.
- **The practical cases are already covered.** The domains that want ellipses (cells, nuclei, etc.) are well served by `polygon` and `mask`, which labelme already supports and which annotation pipelines consume directly.
- **The codebase has moved on.** Since 2021 the `Shape` model became Qt-free (numpy-based), rendering moved out of `QPainterPath`, and `oriented_rectangle` landed with full rotation support. This PR can't rebase onto that, and its core unsolved problem — rendering rotated ellipses — would need a from-scratch redo.

For the record: if we ever revisit this, the right path is to reuse `oriented_rectangle`'s inscribed ellipse (which already solves the rotation rendering this PR couldn't), not scikit-image fitting. That's captured in [`.out-of-scope/ellipse-shape.md`](https://github.com/wkentaro/labelme/blob/main/.out-of-scope/ellipse-shape.md) so the decision and its caveats aren't lost.

Genuinely appreciate the effort here — the geometry write-up in the description was a useful reference for thinking it through.

Issue #930, opened 2021-10-08
Adding point to polygon in edition mode


markub3327:
Hi,
 
   Can I add a point to the existing polygon in editing mode, because I don't like to delete the whole polygon and draw it again?

You may add the button to draw a new point between 2 existing points in polygon during editing polygons.

Thanks a lot.

PR #931, opened 2021-10-11
Interactively adding point to edge

markub3327:
Hi @wkentaro,

  I replaced the `Add Point to Edge` button with an interactive method of adding a vertex to the edge. We must only click on the edge in editing mode where isn't an existing vertex. After this operation is a new vertex created.

Solution of #930

Thanks.
Have a nice day.


markub3327:
@wkentaro What you think about this changes?

PR #940, opened 2021-10-26
Arrow keys & Deselecting shape & fixed movement

markub3327:
Hi @wkentaro ,

I done some new updates:
- Add polygon movement with arrow keys
- Deselect shape when clicking and already selected(?)
- Intuitively using of Right click (like in another application)
- Remove vertex on mouse press (not on release mouse - it is more naturally)
- Fixed issue with moving more then one shape at same time (offsets was not calculated correctly because not include the position of every shape that is moving)
 
Thanks, if something will be incomprehensible I explain it.
Have a nice day.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: ci
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
 
 jobs:


wkentaro:
Can you split commits and PRs for each different feature? It'll make the review much easier.

PR #942, opened 2021-11-04
Movement with arrow keys

markub3327:
Hi @wkentaro,

  here is feature for moving shapes with arrow keys and updated `calculateOffsets` for moving multiple shape simultaneously.

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -478,7 +478,6 @@ class Canvas(QtWidgets.QWidget):
         else:
             for shape in reversed(self.shapes):
                 if self.isVisible(shape) and shape.containsPoint(point):
-                    self.calculateOffsets(shape, point)
                     self.setHiding()
                     if multiple_selection_mode:
                         if shape not in self.selectedShapes:
@@ -487,17 +486,15 @@ class Canvas(QtWidgets.QWidget):
                             )
                     else:
                         self.selectionChanged.emit([shape])
+                    self.calculateOffsets(point)
                     return
         self.deSelectShape()
 
-    def calculateOffsets(self, shape, point):
-        rect = shape.boundingRect()
-        left, right, top, bottom = (
-            rect.left(),
-            rect.right(),
-            rect.top(),
-            rect.bottom(),
-        )
+    def calculateOffsets(self, point):
+        left = self.pixmap.width() - 1
+        right = 0
+        top = self.pixmap.height() - 1
+        bottom = 0
         for s in self.selectedShapes:
             rect = s.boundingRect()
             if rect.left() < left:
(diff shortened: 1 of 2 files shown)


labelme/widgets/canvas.py
```diff
@@ -476,7 +478,7 @@ def selectShapePoint(self, point, multiple_selection_mode):
         else:
             for shape in reversed(self.shapes):
                 if self.isVisible(shape) and shape.containsPoint(point):
-                    self.calculateOffsets(shape, point)
+                    self.calculateOffsets(self.selectedShapes, point)
```
wkentaro, line 481: This can be out of the for-loop?

wkentaro (changes requested):

labelme/widgets/canvas.py
markub3327: Yes of course.

wkentaro: Ok. Why it has to be in the for-loop?

markub3327: Why is it iterating over all `self.selectedShapes` or why it is iterating over `reversed(self.shapes)`?

1. check all shapes in `self.selectedShapes` and predict new offsets that accord all the shapes.
2. it's run only once because `480: if self.isVisible(shape) and shape.containsPoint(point)` and `490: return`

labelme/widgets/canvas.py
```diff
@@ -490,10 +492,27 @@ def selectShapePoint(self, point, multiple_selection_mode):
...
+        for s in self.selectedShapes:
+            rect = s.boundingRect()
+            if rect.left() < left:
+                left = rect.left()
+            if rect.right() > right:
+                right = rect.right()
+            if rect.top() < top:
+                top = rect.top()
+            if rect.bottom() > bottom:
+                bottom = rect.bottom()
```
wkentaro, line 510: I'm still not sure why you first iterate through `shape` and integrate through `self.selectedShapes` in `self.calculateOffsets()`.

wkentaro: Is this nested for-loop really necessary?

markub3327: Try select more then one shape and move with that by mouse ... In one direction it will be outside the pixmap becuse self.calculateOffsets() in original version calculates only with one shape not all that selected ... this creates new virtual rectangle.

wkentaro: I updated `self.calculateOffsets()`. Does that look good to you?

markub3327 (commented):
Ok.

wkentaro (approved):

wkentaro:
Please resolve the conflicts.

PR #944, opened 2021-11-04
Deselect shape when clicking and already selected

markub3327:
Hi @wkentaro 

  here is the last part of my big PR focused on: `Deselect shape when clicking and already selected`

Thanks.

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -76,7 +76,6 @@ class Canvas(QtWidgets.QWidget):
         self.prevhEdge = None
         self.movingShape = False
         self.snapping = True
-        self.wasSelected = False
         self._painter = QtGui.QPainter()
         self._cursor = CURSOR_DEFAULT
         # Menus:
@@ -414,16 +413,6 @@ class Canvas(QtWidgets.QWidget):
                 # Cancel the move by deleting the shadow copy.
                 self.selectedShapesCopy = []
                 self.repaint()
-        elif ev.button() == QtCore.Qt.LeftButton:
-            if self.editing():
-                if (
-                    self.hShape is not None
-                    and self.wasSelected
-                    and not self.movingShape
-                ):
-                    self.selectionChanged.emit(
-                        [x for x in self.selectedShapes if x != self.hShape]
-                    )
 
         if self.movingShape and self.hShape:
             index = self.shapes.index(self.hShape)
@@ -499,9 +488,14 @@ class Canvas(QtWidgets.QWidget):
                             )
                         else:
                             self.selectionChanged.emit([shape])
-                        self.wasSelected = False
                     else:
-                        self.wasSelected = True
+                        self.selectionChanged.emit(
+                            [
+                                x
+                                for x in self.selectedShapes
+                                if x != self.hShape
+                            ]
+                        )
                     return
         self.deSelectShape()
 
@@ -549,7 +543,6 @@ class Canvas(QtWidgets.QWidget):
         if self.selectedShapes:
             self.setHiding(False)
             self.selectionChanged.emit([])
-            self.wasSelected = False
             self.update()
 
     def deleteSelected(self):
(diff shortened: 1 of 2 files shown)


labelme/widgets/canvas.py
```diff
@@ -476,15 +485,18 @@ def selectShapePoint(self, point, multiple_selection_mode):
...
                             self.selectionChanged.emit(
                                 self.selectedShapes + [shape]
                             )
+                        else:
+                            self.selectionChanged.emit([shape])
+                        self.wasSelected = False
                     else:
-                        self.selectionChanged.emit([shape])
+                        self.wasSelected = True
+                    self.calculateOffsets(self.selectedShapes, point)
```
wkentaro, line 499: This should be shape instead of selectedShapes?

labelme/widgets/canvas.py
```diff
@@ -381,16 +382,25 @@ def mousePressEvent(self, ev):
...
                 self.prevPoint = pos
                 self.repaint()
         elif ev.button() == QtCore.Qt.RightButton and self.editing():
-            group_mode = int(ev.modifiers()) == QtCore.Qt.ControlModifier
-            self.selectShapePoint(pos, multiple_selection_mode=group_mode)
+            if (
+                not self.selectedShapes
+                or self.hShape not in self.selectedShapes
+            ):
+                self.selectShapePoint(pos, multiple_selection_mode=False)
```
wkentaro, line 401: Can we keep multiple_selection_mode=group_mode?

wkentaro (changes requested):

labelme/widgets/canvas.py
markub3327: Yes, I missed that `self.calculateOffsets` will changed in another PR to enable operating with multiple selected shapes.

labelme/widgets/canvas.py
markub3327: Yes.

labelme/widgets/canvas.py
```diff
@@ -403,13 +414,12 @@ def mouseReleaseEvent(self, ev):
...
-        elif ev.button() == QtCore.Qt.LeftButton and self.selectedVertex():
-            if (
-                self.editing()
-                and int(ev.modifiers()) == QtCore.Qt.ShiftModifier
-            ):
-                # Delete point if: left-click + SHIFT on a point
-                self.removeSelectedPoint()
+        elif ev.button() == QtCore.Qt.LeftButton:
+            if self.editing():
+                if self.wasSelected and not self.movingShape:
```
wkentaro, line 419: `self.wasSelected` can be replaced with `self.hShape in self.selectedShapes`?

markub3327: Hmm ... I think no because `self.hShape` is always in `self.selectedShapes` when it's called at `mouseReleaseEvent()` after the `selectShapePoint()`. That's what `self.wasSelected` do ... Makes a difference between newly added shape and previously selected shape (both are in `self.selectedShapes` until Line 425).

labelme/widgets/canvas.py
```diff
@@ -478,13 +492,16 @@ def selectShapePoint(self, point, multiple_selection_mode):
...
+                        if multiple_selection_mode:
                             self.selectionChanged.emit(
                                 self.selectedShapes + [shape]
                             )
+                        else:
+                            self.selectionChanged.emit([shape])
+                        self.wasSelected = False
                     else:
-                        self.selectionChanged.emit([shape])
+                        self.wasSelected = True
```
wkentaro, line 504: Can't we move 
```
                    self.selectionChanged.emit(
                        [x for x in self.selectedShapes if x != self.hShape]
                    )
```
here?

wkentaro: I'm not sure how it would be different whether shapes are unselected when the mouse is clicked or released.

markub3327: Done.

wkentaro (approved):

markub3327:
What do with that?

PR #945, opened 2021-11-06
adaptively approximate the circle shape

CosmosHua:
As the radius of the circle increases, the difference between the circular arc and the approximate line segement increase. So, a fixed section number may result in terrible approaches for the big circles, which may lead to a bad instance segmetantion trained model when we actually want some circular segments.

### DIFF
--- a/examples/instance_segmentation/labelme2coco.py
+++ b/examples/instance_segmentation/labelme2coco.py
@@ -132,7 +132,9 @@ def main():
             if shape_type == "circle":
                 (x1, y1), (x2, y2) = points
                 r = np.linalg.norm([x2 - x1, y2 - y1])
-                n_points_circle = 12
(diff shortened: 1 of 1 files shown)


wkentaro:
Can you give visual results?

CosmosHua:
> Can you give visual results?

Screenshot 2021-11-11 16:07:56

The radius of the wheel in the above image is about 350 pixels, the bottom is used N=12, the upper is used the new N formula. You can calculate the difference between the arc and line segment **using simple trigonometric math**, when r=350 and N=12. Further more, you will easily obtain the same math result as mine. Or, you can find some images including big circles, such as wheels, and then label them and convert to the COCO format using different N, and you will see the difference. The bigger the circle, the more obvious the difference.

wkentaro:
Thanks. And yes, the formula that minimizes the gap between the segment and arc makes sense.

PR #947, opened 2021-11-08
labelme2coco.py: added train/test split and allowing for multiple input dirs

sebastian-ruiz:
The script will now return 3 .json files instead of only one.

- _train.json
- _test.json
- _coco.json

The split parameter lets you decide the proportion of training images, number between 0-1. The _coco.json file contains all the annotations.

The `input_dir` parameter can take a list of directories. This is useful when one has multiple folders of labelled images, but wants to create a single `_coco.json` for all the images. The train/test split is taken proportionally to the files in each folder.

### DIFF
--- a/examples/instance_segmentation/labelme2coco.py
+++ b/examples/instance_segmentation/labelme2coco.py
@@ -12,6 +12,9 @@ import uuid
 
 import imgviz
 import numpy as np
+from sklearn.model_selection import train_test_split
+import regex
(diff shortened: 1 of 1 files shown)


wkentaro:
I think we could achieve the same thing by creating `data_annotated_train` and `data_annotated_test` from `data_annotated`, and that would be more flexible to have different splits.

PR #950, opened 2021-11-09
Revert "Revert "Deselect shape when clicking and already selected""

markub3327:
Reverts wkentaro/labelme#949

@wkentaro 

Here bottle is too tight and epsilon too high (I can't effectively pick the object) .... I try to go back to last commit and for me everything is working. In this video you can see why this shape is not good for testing.

Thanks have a nice day.

https://user-images.githubusercontent.com/74611856/140948458-303bd3a9-1642-4e8a-91d4-093fadf9a624.mp4

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -38,7 +38,6 @@ from labelme.widgets import ZoomWidget
 
 # TODO(unknown):
 # - [high] Add polygon movement with arrow keys
-# - [high] Deselect shape when clicking and already selected(?)
 # - [low,maybe] Preview images on file dialogs.
 # - Zoom is too "steppy".
 
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -76,7 +76,7 @@ class Canvas(QtWidgets.QWidget):
         self.prevhEdge = None
         self.movingShape = False
         self.snapping = True
-        self.wasSelected = False
+        self.hShapeIsSelected = False
         self._painter = QtGui.QPainter()
         self._cursor = CURSOR_DEFAULT
         # Menus:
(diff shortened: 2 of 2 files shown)


wkentaro:
I know it works, but we have to keep the mouse clicked, right?
I sometimes release the mouse while dragging and click again to continue moving the selected shapes. However, with the current implementation, it deselects the polygon the mouse is hovering and disables the movement of that.

wkentaro:
I'm not saying deselection is a bad idea, but deselecting with Ctrl+Click would be inconvenient because of the above reason.
In MS PowerPoint, the multi-object selection is done by Ctrl+Click, movement is done by Click+drag (instead of Ctrl+Click+drag), deselection is done by Ctrl+click or Clicking blank area. Maybe we can follow this UI.

markub3327:
> I know it works, but we have to keep the mouse clicked, right?
> I sometimes release the mouse while dragging and click again to continue moving the selected shapes. However, with the current implementation, it deselects the polygon the mouse is hovering and disables the movement of that.

You don't have true. I test it again with release mouse button and the shape is not deselected. And when I select multiple object use Ctrl+Click, but when I move with them I use only Click+drag. Please try this version of code again... The deselecting was not activate because this if statement:
```python
                 if (
                     self.hShape is not None
                     and self.wasSelected
                     and not self.movingShape
                 ):
                     self.selectionChanged.emit(
                         [x for x in self.selectedShapes if x != self.hShape]
                     )
```

wkentaro:
Thanks. With the latest commit 4c23f44, it works as I expect.
Is that collect that `self.wasSelected` is used for distinguishing the state of `self.hShape` whether that is hovered or selected? If so, please rename that to something like `self.hShapeIsSelected`.

markub3327:
Yeah, you have a true 😊 .... I renamed it.

Thanks.
Have a nice day.

wkentaro (approved):

PR #951, opened 2021-11-11
Add pyproject.toml for black

wkentaro:
No longer it is a problem with pyinstaller.


wkentaro:
Python2.7 won't support this for some reason.

PR #959, opened 2021-11-24
Add pyproject.toml for black

wkentaro:


wkentaro:
pyproject.toml cannot work with python2. It fails to import PyQt with pyproject.toml.

PR #975, opened 2022-01-12
Update ReadMe Testing on MacOS

cuicaihao:
I was following the ReadMe to test all the commands and find that on macOS users will need to install **help2man** and overwrite the **labelme.1** file before testing. 

```bash
## macOS testing 
brew install help2man 
help2man labelme > $(pwd)/docs/man/labelme.1
```
Then go with the following testing.

code_review

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,15 @@ For more advanced usage, please refer to the examples:
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


wkentaro:
I'm not sure why you're using 4.5.9 but labelme.1 is 4.6.0?

PR #955, opened 2021-11-17
Sort File-List in numerical order

pit-ray:
Currently, if the following files exist, they are displayed in string order.
```
$ ls
image1.png image2.png image10.png image3.png image200.png
```

In labelme.
```
image1.png
image10.png
image2.png
image200.png
image3.png
```

However, I think it is natural to use a list of string numbers converted to numbers for sorting. I added a utility function for this to `utils/_io.py` because file names are i/o identifiers.

PR #989, opened 2022-02-25
images are sorted based on number in the image name

sebastian-ruiz:
I had problems where the order of the images is messed up when I labelled my images 0.png, 1.png, ...., 10.png, 11.png, etc...

This is a quick solution for sorting by number.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -6,6 +6,7 @@ import os
 import os.path as osp
 import re
 import webbrowser
+import regex
 
 import imgviz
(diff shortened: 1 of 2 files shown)


sebastian-ruiz:
Oh I now see that this is similar to https://github.com/wkentaro/labelme/pull/955 Sorry!

This would sort:
1_1
1_100
1-10
1-11
1_1-1

To:

1_1
1-10
1_1-1
1-11
1_100

PR #1012, opened 2022-04-27
[fix] fix a bug when opening a label file

Leojc:
Opening a label file using the same encoding when it was saved. 

By doing this, labelme can open a label file when it contains Unicode characters.


Leojc:
I'm thinking that maybe using "utf8" as the only one encoding is not a perfect solution. Maybe we could add an encoding configuration so that people can choose what encoding they want, like "GB18030", "BIG-5", "Shift JIS", etc...

PR #1034, opened 2022-06-13
fix: QPoint arg must be int

HHongSeungWoo:
```
  File "/home/hong/PycharmProjects/labelme/labelme/widgets/canvas.py", line 618, in paintEvent
    p.translate(self.offsetToCenter())
  File "/home/hong/PycharmProjects/labelme/labelme/widgets/canvas.py", line 659, in offsetToCenter
    return QtCore.QPoint(x, y)
TypeError: arguments did not match any overloaded call:
  QPoint(): too many arguments
  QPoint(int, int): argument 1 has unexpected type 'float'
  QPoint(QPoint): argument 1 has unexpected type 'float'
```

fix bug


HHongSeungWoo:
python 3.10 bug

PR #1035, opened 2022-06-14
When loading the image file, try both original and basename paths

drjasonharrison:
- if the label file contains an imagePath is an absolute path, then os.path.join will return the original absolute path which may no longer reflect the imagePath of the file on a different computer
- if loading the imagePath fails, try using the dirname of the label file and the basename of the imagePath

### DIFF
--- a/labelme/label_file.py
+++ b/labelme/label_file.py
@@ -111,6 +111,14 @@ class LabelFile(object):
                 # relative path from label file to relative path from cwd
                 imagePath = osp.join(osp.dirname(filename), data["imagePath"])
                 imageData = self.load_image_file(imagePath)
+                if imageData is None:
+                    originalPath = imagePath
+                    # use relative path of label file and basename of imagePath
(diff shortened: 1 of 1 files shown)


wkentaro:
I guess we should avoid this: "if the label file contains an imagePath is an absolute path" first (right?).
Should be done here: https://github.com/wkentaro/labelme/blob/a6a8ee81540582cc6e4a28de400ec153452e03b3/labelme/app.py#L1260

Do you know why it is still happening?

wkentaro:

Thank you for this, @drjasonharrison, and apologies for the very long delay. The fallback you proposed — trying the image's basename next to the label file when the recorded `imagePath` can't be loaded — is a good idea and genuinely helps datasets that move between machines.

Since this PR was opened, the label-file reader was refactored (`label_file.py` no longer exists), so the diff no longer applies. Rather than ask you to rebase a four-year-old branch, I've captured the behavior in a fresh, ready-to-implement issue and credited you there: #2125.

Closing this PR in favor of that issue. Thanks again for the contribution.

PR #1143, opened 2022-06-30
Read directory with only json labels

HowcanoeWang:
Fix this issue: https://github.com/wkentaro/labelme/issues/1009

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -2036,37 +2036,36 @@ class MainWindow(QtWidgets.QMainWindow):
         # it will return fileNames=[], labelFiles=[]
         # then don't set openNextImg button active
         if len(fileNames) == 0:
-            pass
-        else:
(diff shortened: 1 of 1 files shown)


wkentaro:

Thanks for contributing this feature! Unfortunately this PR has accumulated merge conflicts with current `main` that would require substantial rework to resolve — the codebase has seen several refactors to `app.py` and `LabelFile` since you opened this. The implementation also has some design issues worth addressing in a fresh attempt: `scanAllImages` now returns a tuple which breaks its contract, and mutating `self.output_dir` as a side-effect inside a scan helper is fragile. If you'd like to revisit this, please rebase on current `main`, keep `scanAllImages` returning only image paths (or introduce a new helper), and avoid side-effects in the scan step. Happy to re-review a refreshed PR!

PR #1145, opened 2022-07-11
Group_id visualization by color 

IkondorLogi:
Add new feature into LabelMe to make it visualize group_id by color!
This feature is necessary for data validation when there is a need to show a user the relationship between parent and child shapes with the same group_id value.
177431710-cdf26754-f38b-4f07-aed3-b4eb730b8ce4
177431708-efd8bac2-f7ef-48b2-9721-dfe273dad447

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -43,7 +43,7 @@ from . import utils
 
 
 LABEL_COLORMAP = imgviz.label_colormap()
-
+GROUP_ID_COLORMAP = utils.shape.generate_random_colors(100, bright=True)
 
 class MainWindow(QtWidgets.QMainWindow):
     FIT_WINDOW, FIT_WIDTH, MANUAL_ZOOM = 0, 1, 2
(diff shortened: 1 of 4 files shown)


IkondorLogi:
@wkentaro please take a look at this PR when you have time

wkentaro:

Thanks for this feature idea — group_id color visualization is useful. Unfortunately this PR is now ~4 years old and has unresolvable merge conflicts against the current codebase, which has changed substantially since July 2022. The color generation also has a bug (saturation should be in [0,1] but 256 is passed), and the design (adding state directly to Canvas, a config key for custom colors, random shuffling of the colormap) would need to be rethought against the current architecture. If you still want this feature, please open a new PR against the current main branch with a fresh implementation.

wkentaro:

Thanks for the group_id coloring idea. The PR uses the old PyQt5 color path (`_update_shape_color`, `generate_random_colors`) which has been replaced by the typed `Palette`/`_shape_render.py` pipeline, so it conflicts and no longer applies. The feature is useful for relationship visualization, so I've opened #2192 to track a reimplementation. Recommending close (your call to review or close).

Issue #1021, opened 2022-05-20
When changing the label value with label edit, the program crashes.


sangwoo89:
app.py (1171 row)
item = self.uniqLabelList.findItemsByLabel(label)[0]
there is out of range error.

add code.. good working.

 def _get_rgb_by_label(self, label):
        if self._config["shape_color"] == "auto":
 if not self.uniqLabelList.findItemsByLabel(label):
                item = QtWidgets.QListWidgetItem()
                item.setData(Qt.UserRole, label)
                self.uniqLabelList.addItem(item)

item = self.uniqLabelList.findItemsByLabel(label)[0]

wkentaro:
Will be fixed by v5.0.2

```
pip install --no-cache-dir labelme==5.0.2
```

Issue #963, opened 2021-12-03
[BUG] "IndexError: list index out of range" when edit label


hoozh:
**Describe the bug**
label crashed when edit label.

**To Reproduce**
1. Load Labelme
2. Click "Open Dir"
3. Check "Save Automatically"

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Ubuntu 18.04]
 - Labelme Version [e.g. 4.2.9]

**Additional context**

ShuaiLYU:
I have met the same bug. have you found a solution?

hoozh:
> I have met the same bug. have you found a solution?

try another version

floyddcn:
Latest version 5.0.1 still has the crash problem when changing existing label value to a non-existing string value.
For me,the only choice seems to be a fallback to 4.2.10 is Ok

shihaoyin:
> Latest version 5.0.1 still has the crash problem when changing existing label value to a non-existing string value.
> For me,the only choice seems to be a fallback to 4.2.10

Have you solve the crash problem when editing labels？

happyamyhope:
> **Describe the bug** labelme crashed when edit label.
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
> 1. Load Labelme
> 2. Click "Open Dir"
> 3. Check "Save Automatically"
> 4. Uncheck "Save With Image Data"
> 5. Check "Keep Previous Annotation"
> 6. Select last image in _**File List**_, and annotate it with a label
> 7. Press A to open previous image,
> 8. Edit the label
> 9. Labelme crashed.
> 
> **Expected behavior** Label can be updated.
> 
> **Screenshots**
> 
> ```
> Traceback (most recent call last):
>   File "D:\Anaconda3\envs\labelme_env\lib\site-packages\labelme\app.py", line 1083, in editLabel
>     self._update_shape_color(shape)
>   File "D:\Anaconda3\envs\labelme_env\lib\site-packages\labelme\app.py", line 1163, in _update_shape_color
>     r, g, b = self._get_rgb_by_label(shape.label)
>   File "D:\Anaconda3\envs\labelme_env\lib\site-packages\labelme\app.py", line 1173, in _get_rgb_by_label
>     item = self.uniqLabelList.findItemsByLabel(label)[0]
> IndexError: list index out of range
> ```
> 
> **Desktop (please complete the following information):**
> 
> * OS: `Windows 10`
[...]

PR #1151, opened 2022-07-22
fix bug in editLabel

ochaheaven:
fix crash bug (#963, #1021)

#1027 also solves the same problem, but my solution is different from that. In my PR, I use the label adding logic that appears in `addLabel()` function [here](https://github.com/wkentaro/labelme/blob/819a93cbbc8a7ae3009e9f380111d742e6c06615/labelme/app.py#L1139-L1143).

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1073,6 +1073,12 @@ class MainWindow(QtWidgets.QMainWindow):
         shape.flags = flags
         shape.group_id = group_id
 
+        if not self.uniqLabelList.findItemsByLabel(shape.label):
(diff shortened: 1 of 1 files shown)


wkentaro:

Thanks for the fix! Unfortunately this PR can no longer be merged: `app.py` has been substantially refactored since 2022 — `editLabel` and `uniqLabelList` no longer exist in their original form, so there is nothing to apply this patch to. The underlying bug was addressed as part of the broader refactor. Closing as superseded; please feel free to open a new issue if you still see this crash on the current `main`.

PR #1187, opened 2022-10-07
Fix loading buggy images

wolfrosnet:
Fix an error when loading an image that had the original extension "PNG" but was saved as "JPG".
I have attached a sample image.
road pothole144


pelennor (commented):
How about not to decide buggy images by comparing extensions and Image.mode, but using file MIME type.

For example, using get_format_mimetype method(https://pillow.readthedocs.io/en/stable/releasenotes/5.2.0.html#imagefile-get-format-mimetype)

In my opinion, using MIME type gives you much more reliable informations.

Issue #1020, opened 2022-05-17
Unable to run，lots of TypeError


FallenSequoia:
First off, I performed the following commands when installing (Windows 10, using Anaconda):
`conda create --name=labelme python=3`
`conda activate labelme`
`pip install labelme`

When i try to run labelme, I got the following error:

> `Traceback (most recent call last):
 File "E:\ProgramData\Anaconda3\envs\Image-style-transfer-gp\lib\site-packages\labelme\widgets\canvas.py", line 618, in paintEvent
    p.translate(self.offsetToCenter())
  File "E:\ProgramData\Anaconda3\envs\Image-style-transfer-gp\lib\site-packages\labelme\widgets\canvas.py", line 659, in offsetToCenter
    return QtCore.QPoint(x, y)
TypeError: arguments did not match any overloaded call:
  QPoint(): too many arguments
  QPoint(int, int): argument 1 has unexpected type 'float'
  QPoint(QPoint): argument 1 has unexpected type 'float'

I don't know what caused the problem, maybe a version update of Qt?

andrster1:
Hello, I've run into a similar problem. I tried opening the apc2016_obj3.jpg image for the tutorial and it throws a QtCore error, more specifically: 

> Traceback (most recent call last):
>   File "C:\.conda\envs\labelme\lib\site-packages\labelme\app.py", line 1537, in loadFile
>     self.loadLabels(self.labelFile.shapes)
>   File "C:\.conda\envs\labelme\lib\site-packages\labelme\app.py", line 1226, in loadLabels
>     self.loadShapes(s)
>   File "C:\.conda\envs\labelme\lib\site-packages\labelme\app.py", line 1187, in loadShapes
>     self.addLabel(shape)
>   File "C:\.conda\envs\labelme\lib\site-packages\labelme\app.py", line 1137, in addLabel
>     self.labelList.addItem(label_list_item)
>   File "C:\.conda\envs\labelme\lib\site-packages\labelme\widgets\label_list_widget.py", line 167, in addItem
>     item.setSizeHint(self.itemDelegate().sizeHint(None, None))
>   File "C:\.conda\envs\labelme\lib\site-packages\labelme\widgets\label_list_widget.py", line 64, in sizeHint
>     return QtCore.QSize(
> TypeError: arguments did not match any overloaded call:
>   QSize(): too many arguments
>   QSize(int, int): argument 1 has unexpected type 'float'
[...]

Issue #1006, opened 2022-03-22
Unable to run after installation


lightbearer88:
Hi,

First off, I performed the following commands when installing (Windows 10, using Anaconda):
```
conda create --name=labelme python=3
source activate labelme
conda install labelme -c conda-forge
```
After that, I used the following command in Jupyter Notebooks:

`!labelme`

And I got the following error:
```
Traceback (most recent call last):
  File "C:\Pythons\envs\labelme\Scripts\labelme-script.py", line 6, in <module>
    from labelme.__main__ import main
  File "C:\Pythons\envs\labelme\lib\site-packages\labelme\__init__.py", line 17, in <module>
    QT4 = QT_VERSION[0] == "4"
TypeError: 'NoneType' object is not subscriptable
```
Did I do something wrong or is there a bug?

llokng922:
This is probably because pyqt is not installed probably.

ksooklall:
I had the same issue then ran `conda install -c anaconda pyqt`, but now I have this issue

```
Traceback (most recent call last):
  File "/home/kenan/anaconda3/envs/labelme/lib/python3.10/site-packages/labelme/widgets/canvas.py", line 618, in paintEvent
    p.translate(self.offsetToCenter())
  File "/home/kenan/anaconda3/envs/labelme/lib/python3.10/site-packages/labelme/widgets/canvas.py", line 659, in offsetToCenter
    return QtCore.QPoint(x, y)
TypeError: arguments did not match any overloaded call:
  QPoint(): too many arguments
  QPoint(int, int): argument 1 has unexpected type 'float'
  QPoint(QPoint): argument 1 has unexpected type 'float'
Aborted (core dumped)
```

parksangback:
i resolved this problem!!

mkpvasu:
After solving the above error, I am running to an additional error

Traceback (most recent call last):
  File "C:\Users\munikumar\anaconda3\envs\labelme\lib\site-packages\labelme\widgets\canvas.py", line 245, in mouseMoveEvent
    if QtCore.Qt.RightButton & ev.buttons():
TypeError: unsupported operand type(s) for &: 'MouseButton' and 'MouseButtons'

Does anyone know how to rectify this?

aurelien-m:
Any news on this? How could it work before with `return QtCore.QPoint(x, y)`?

oikonochristina:
i am having the same issue with @mkpvasu , how could we solve this typeError?

[...]

PR #1199, opened 2022-10-23
Use QPointF and QSizeF to avoid TypeErrors in Python3.10

wkentaro:
Close https://github.com/wkentaro/labelme/issues/1006
Close https://github.com/wkentaro/labelme/issues/1020


mapfiable:
Hi, I experience the same TypeError issues running Python 1.11 under Windows 11. However, fixing it by using `QPointF` instead does not work for me, because then I encounter a different TypeError ("arguments did not match any overloaded call"), when I try to call `QGroupBox.move(p)`, where `p` is an instance of a `QPoint`.

PR #1216, opened 2022-11-20
Support High DPI Displays

pelennor:
Add new feature that supporting high DPI displays.

When I switch position of labelme window from 4K display to FHD one and vice versa, scale of labels on 'Polygon Labels'  does not match OS setting.
So I add this feature, and it looks well.


wkentaro:

Thanks for the contribution! HiDPI support has since been added to `labelme/app.py` (lines 69-72) with `hasattr` guards for PySide6 compatibility, covering the same use case. This PR is now superseded, so I'm closing it. If you notice any remaining HiDPI issues on the current main branch, please open a new issue.

PR #1223, opened 2022-12-20
fix(i18n): add qm files to runtime

xiyaowong:


xiyaowong:
@wkentaro Please review

PR #1224, opened 2022-12-20
Application shutdowns when sliding canvas out of bixmap

mhannani:
When sliding the shape out of bixmap the application stops and complains that the parameters `QtCore.QPoint` needs to be `int` insread of `float`.

`pos += QtCore.QPoint(
TypeError: arguments did not match any overloaded call:
  QPoint(): too many arguments
  QPoint(int, int): argument 1 has unexpected type 'float'
  QPoint(QPoint): argument 1 has unexpected type 'float'`


wkentaro:
@mhannani Thanks!

PR #1262, opened 2023-04-16
Integrate with Segment Anything Model (SAM)

wkentaro:
- Close https://github.com/wkentaro/labelme/discussions/1260
- Close https://github.com/wkentaro/labelme/discussions/1257

## TODO

- [x] Positive/negative click to run AI-assisted polygon annotation
- [x] Fix weird behavior when annotating many images https://github.com/wkentaro/labelme/pull/1262#issuecomment-1511834423
- [x] Support portrait images https://github.com/wkentaro/labelme/pull/1262#issuecomment-1521299774
- [x] Support 1-click annotation
- [x] Support double_click="close"
- [x] Real-time feedback without click https://github.com/wkentaro/labelme/pull/1265
- [x] Dropdown menu to change AI models

## Installation

1. Install conda
```
# Linux/Mac: command to install anaconda3 and labelme
curl -L https://github.com/wkentaro/dotfiles/raw/main/local/bin/install_anaconda3.sh | bash -s .
source .anaconda3/bin/activate

# Windows: check https://www.continuum.io/downloads
```

2. Install Labelme with `sam` branch.

```
git clone https://github.com/wkentaro/labelme.git
cd labelme
git checkout sam
pip install -e .
```

## Usage

```
labelme examples/instance_segmentation/data_annotated
```

1. Select "Create AI-Polygon"
2. Do `click` for the region you want to include
3. Do `shift+click` for the region you want to exclude
4. Do `ctrl/cmd+click` or `ctrl/cmd+shift+click` or double-click to finish
5. Press `esc` to go back and continue clicking

https://user-images.githubusercontent.com/4310419/235450790-e70e0351-21eb-4e2b-a14a-c874bb2bae1d.mp4

---

**NOTE: The response for the first annotation is SLOW (~20s)** because the AI model needs to extract image embedding from the entire image, but it is cached in the second you annotate.

Issue #1164, opened 2022-08-26
i want to move an box,but always draw a point  that's make so bad to use? it's very boring


cqray1990:
### Provide environment information

i want to move an box,but always draw a point  that's make so bad to use? it's very boring

### What OS are you using?

ubantu

### Describe the Bug

i want to move an box,but always draw a point  that's make so bad to use? it's very boring

lxgyChen:
You should click the `EditPolygons`  button, which on the left sidebar

cqray1990:
> You should click the `EditPolygons` button, which on the left sidebar

i just want to move the box, so i need selected the box , not create box

wkentaro:
Can you provide how to reproduce? It seems working for me.

https://user-images.githubusercontent.com/4310419/192311325-b84cff87-fbb4-4dec-8700-9cae0a28538b.mp4

wkentaro:
> *This was generated by AI during triage.*

## Triage Notes

**What we've established so far:**

- Revisited this against the current code (`labelme/_widgets/canvas.py`).
- The canvas has two mutually exclusive modes, Edit and Create. In **Create** mode a left click always creates — `_press_left` dispatches on the mode alone and does no hit-testing against existing shapes, so with a point tool active, clicking on top of an existing box drops a point on it rather than selecting the box. In **Edit** mode, clicking and dragging a shape moves it as expected.
- That matches the video @wkentaro posted in 2022, and the existing end-to-end test `tests/e2e/canvas_interaction_test.py::test_move_shape_by_drag` still covers drag-to-move. The canvas widget test suites pass on current `main` (153 tests green).
- **Could not reproduce** "always draws a point" under the normal flow. The only way to produce that symptom is to click while a creation tool is still active without first switching to Edit — which is the mode confusion @lxgyChen's original suggestion pointed at.
- No repro steps or version info were ever provided (the template's environment field was left blank), and there has been no reply from @cqray1990 since September 2022.
- No matching entry in `.out-of-scope/`, and no clear duplicate among open issues. #1316 (right-button multi-select move) and #2398 (keyboard nudge) are different symptoms.

**Parked for a maintainer call.** As filed this reads as a usage question about the explicit Create-vs-Edit split rather than a reproducible defect. If there is appetite for the underlying UX request — clicking an existing shape to select it without first leaving Create mode — that would be better tracked as a fresh feature issue.

PR #1265, opened 2023-04-25
Sam realtime render

luzhenyv:
1. Add real-time rendering snippets, refer to "Fill Drawing Polygon" function.  #1262 
2. Config `shape.fill_color = [0, 0, 0, 128]` in default_config.yaml, to render SAM mask
3. Fix pytest bugs

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -218,7 +218,6 @@ class Canvas(QtWidgets.QWidget):
                 pos = self.transformPos(ev.posF())
         except AttributeError:
             return
-
         self.prevMovePoint = pos
         self.restoreCursor()
 
@@ -331,6 +330,7 @@ class Canvas(QtWidgets.QWidget):
                 self.update()
                 break
             elif index_edge is not None and shape.canAddPoint():
+                logger.debug(f"index_edge: {index_edge}")
                 if self.selectedVertex():
                     self.hShape.highlightClear()
                 self.prevhVertex = self.hVertex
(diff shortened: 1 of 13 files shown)


labelme/widgets/canvas.py
```diff
@@ -732,6 +732,28 @@ def paintEvent(self, event):
             drawing_shape.addPoint(self.line[1])
             drawing_shape.fill = True
             drawing_shape.paint(p)
+        elif (
+            self.createMode == "ai_polygon"
+            and self.current is not None
+        ):
+            drawing_shape = self.current.copy()
+            drawing_shape.addPoint(self.line[1])
+            points = self._ai_callback(
```
d710055071, line 741: This code should not be executed when pressing the ctrl key

luzhenyv: **Describe**

A new pull addresses two issues in the AI-polygon feature:

1. A bug caused by function labelme.utils.distancetoline. #1164 

https://user-images.githubusercontent.com/49972828/234761611-42eef991-4732-47bc-b772-26fff35d246b.mp4

2. The addition of `Mouse Double Click` to end drawing AI-polygon, making it more user-friendly.

**Issue one**
The first issue was related to a bug caused by function `labelme.utils.distancetoline `. This bug resulted in an error in point-to-line distance calculation when two polygon vertices were the same. 



So the system always finds a non-none nearestEdge, users had to add a new vertex before editing labels. To address this, I've modified the code (in Line 94 of `labelme.utils.qt.py`) to return a point-to-point distance instead of 0 in this case.

**Issue Two**
The second issue was adding the functionality of `Mouse Double Click` to end drawing AI-polygons, which makes it more user-friendly as users need no longer press "Ctrl" and click mouse left keys at the same time.

PR #1267, opened 2023-04-25
Portrait aspect ratio image embedding

f-macfarlane95:
Updated `segment_anything.py` to allow for portrait and square aspect ratio images as discussed in #1262 

1. Updated `compute_image_embedding` to calculate the required scale factor to reduce the maximum image dimension to the desired `image_size` and removed the assertion that `assert image.shape[1] > image.shape[0]` to allow for portrait and square images.
2. Identical changes as above, but to `compute_polygon_from_points`. 

Tested on both portrait and square images and both the embedding process and returned polygons seem sensible.

### DIFF
--- a/labelme/ai/models/segment_anything.py
+++ b/labelme/ai/models/segment_anything.py
@@ -83,12 +83,12 @@ class SegmentAnythingModel:
 
 
 def compute_image_embedding(image_size, encoder_session, image):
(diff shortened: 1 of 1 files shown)


wkentaro:
Thanks! Fixed by de2754eeb4422dccaf463242638c6f13c32a9c19

PR #1269, opened 2023-04-28
fix `description` not being defined if `display_label_popup=false`

### LINKED ISSUE #370: Point annotation without label
I just want to use point annotation and save points' coordinate. I don't need labels.
1. How to disable label dialog? Or it automatically chooses a default label without label dialog.
2. Can labelme only save  points' coordinate without other information?

kodonnell:
As per the title. Fixes #370 which no longer works.


wkentaro:
Thanks! https://github.com/wkentaro/labelme/pull/1270 will fix this.

PR #1276, opened 2023-05-11
Ordered annotations

ilovemicroplastics:
it works fine


wkentaro:

Thank you for this contribution. Unfortunately this PR cannot be merged as-is for several reasons: (1) it has merge conflicts with the current main branch; (2) it imports `PyQt5` directly instead of going through the `qtpy` abstraction layer the project uses; (3) `highlightOutliers` references an undefined `labelList` variable (should be `self.labelList`), which would cause a runtime error; (4) the `findItemByShape` API change (returning `[bool, item]` instead of raising) breaks the existing call contract; (5) adding color state as class-level variables on `Shape` means all shapes share the same color state, which will cause rendering bugs. If you'd like to resubmit, please rebase onto current main, fix the PyQt5 import, and address the correctness issues above.

PR #1290, opened 2023-06-03
Labelme plugins system

leonelhs:
Consider this feature.
In order to don't increase the complexity of Labelme source code I have added a plugin manager system that allows the users extends the capabilities  without make changes to the whole labelme source code. 

For a demo please visit https://github.com/leonelhs/labelme-plugins, this plugin creates person body shape contours using PyTorch  Resnet101 segmentation API.

Regards.


wkentaro:

Thanks for this contribution! Unfortunately we can't merge this as-is: (1) it adds a mandatory dependency on `yapsygui`, a package with very little adoption and no recent maintenance, which would become a hard requirement for all labelme users; (2) the PR has merge conflicts against current main that would need to be resolved; (3) the maintainer has a plugin system planned as part of the upcoming roadmap that will be designed to fit the new architecture — it's better for that to be driven from within the project so the design is consistent. We're closing this PR but appreciate the idea; the demo repo at https://github.com/leonelhs/labelme-plugins shows a good use case that can inform the official design.

wkentaro:

Thanks for proposing a plugin system. The PR adds a hard dependency on the third-party `yapsygui` and patches a since-restructured `app.py`, so it conflicts and is architecturally misaligned with the current private `_`-prefixed internals. A first-party plugin/entry-point design is already tracked in #2155. Recommending close (your call to review or close).

PR #1332, opened 2023-10-02
Fix error when running SAM on onnx

evanfebrianto:
explicitly set the providers parameter when instantiating InferenceSession on OnnxRuntime


wkentaro:
Since this seems only happen with onnxruntime==1.16.0, let's fix like this: https://github.com/wkentaro/labelme/pull/1364

PR #1372, opened 2023-12-19
Fix crash on Ubuntu 22.04 (Python 3.10)

step305:


labelme/widgets/canvas.py
```diff
@@ -827,7 +827,7 @@ def offsetToCenter(self):
         aw, ah = area.width(), area.height()
         x = (aw - w) / (2 * s) if aw > w else 0
         y = (ah - h) / (2 * s) if ah > h else 0
-        return QtCore.QPointF(x, y)
+        return QtCore.QPointF(int(x), int(y))
```
wkentaro, line 830: QPointF should support float, so seems something else is wrong.

krenax:
I had the same issue. This small correction seems to resolve this.

wkentaro:

Thanks @step305 for the PR and @krenax for confirming — you both hit a real crash (the PyQt5 5.15 / Python 3.10 strictness where integer Qt APIs reject Python `float`s).

This is now resolved on `main`. We verified on the exact configuration from this report (Python 3.10, Qt 5.15.14, PyQt5 5.15.11):

- The app constructs and starts cleanly, no crash.
- Every place where a float computation feeds an integer-typed Qt API (`QPoint`/`QRect`/`QSize`/`setValue`/`drawLine`) now casts with `int()` at the boundary.

Note on the diff here: `QPointF` is the floating-point point class and accepts floats natively, so the `int()` cast on it was not the actual fix — the real crash was at the integer APIs, which `main` now guards. The `app.py` change (`selectionChanged` → `selectionchanged`) would also have broken launch, since that signal was never lowercase (it is now `selection_changed`).

Since the codebase has been refactored since 2023 and the diff no longer applies, I am closing this PR. The underlying bug is fixed. Thanks again for reporting and digging in.

PR #1440, opened 2024-05-15
solves IOError: image file truncated in windows

nouranali:
this issue appeared on the latset version of labelme on windows 11

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -32,6 +32,7 @@ from labelme.widgets import LabelListWidgetItem
 from labelme.widgets import ToolBar
 from labelme.widgets import UniqueLabelQListWidget
 from labelme.widgets import ZoomWidget
+from PIL import ImageFile
 
 from . import utils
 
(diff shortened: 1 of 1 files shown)


nouranali:
@wkentaro

wkentaro:
@nouranali Can you give me an example image so that I can reproduce?

nouranali:
> @nouranali Can you give me an example image so that I can reproduce?

unfortunately the images I am working on are confidential

wkentaro:

Thanks for the patch, @nouranali, and apologies for the slow response.

After review I'm going to decline this one. Setting `ImageFile.LOAD_TRUNCATED_IMAGES = True` globally makes Pillow silently *succeed* on truncated files by filling the missing portion with padding, rather than raising. In a general image viewer that's a convenience, but labelme is an annotation tool: a user would then label an image whose truncated region is fabricated padding, and those annotations get saved as ground truth against pixels that were never really in the file, with no warning. We'd rather fail loudly on a truncated image, since it almost always signals a broken download/transfer or disk write that's best fixed at the source.

We also weren't able to reproduce the original error (the sample images are confidential, which is completely understandable), so we can't confirm the failure mode or that this is the right remedy.

Closing as won't-fix. If you're hitting this regularly, the more robust path is to validate/re-fetch the truncated files before loading them into labelme. Thanks again for taking the time to contribute.

PR #1455, opened 2024-06-13
Support multiple edit of annotations

wkentaro:
Close https://github.com/labelmeai/labelme/pull/1369

```
1. person -> [select] -> human
2. person -> [select] -> human
3. bottle
```

```
1. person -> [select] -> <can change group_id but not label>
2. person -> [select] -> <can change group_id but not label> 
3. bottle -> [select] -> <can change group_id but not label>
```

If we want to edit multiple different labels at the same, we need different button.

https://github.com/labelmeai/labelme/assets/4310419/237cca70-9426-4a36-ad37-21523f6b3d75


wkentaro:
@Daasein Let's introduce the new action when it's absolutely needed (e.g., we need to edit "red" and "green" back to "traffic lights".

PR #1460, opened 2024-06-18
Optimize label\shape.py

PairZhu:
Optimized the rendering logic of shapes so that the thickness of shape lines and crosshair remains consistent on the screen regardless of the zoom level.

Additionally, another advantage is that shapes do not become distorted when the x and y axes have different scaling ratios (this is useful in my project, but such a requirement is uncommon, so the related changes have not been uploaded).

### DIFF
--- a/labelme.spec
+++ b/labelme.spec
@@ -1,7 +1,10 @@
 # -*- mode: python -*-
 # vim: ft=python
 
+import os.path as osp
 import sys
+
+import osam._models.yoloworld.clip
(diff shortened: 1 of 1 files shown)


PairZhu:
The first CI failure doesn't seem to be caused by my code, can you check the cause? @wkentaro

wkentaro:
Reopened in https://github.com/labelmeai/labelme/pull/1471

PR #1462, opened 2024-06-21
fix bug editlabel window is marked while cursor on the below of screen

happybuby:
while editlabel,if the cursor on the below of screen,the popup editlabel window will be marked most,only show alittle,this push can  fix the  bug and show the whole window

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1115,6 +1115,7 @@ class MainWindow(QtWidgets.QMainWindow):
             flags=shape.flags,
             group_id=shape.group_id,
             description=shape.description,
+            centralWidget=self.centralWidget()
         )
         if text is None:
             return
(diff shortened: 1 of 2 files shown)


wkentaro:

Thanks for the report! This PR has merge conflicts with current main and bundles two unrelated changes: an unfinished 'coordinate list' panel (pointList with all signal connections commented out) alongside the popup-position fix. The popup fix itself uses the wrong approach — it moves the system cursor position (`QCursor.setPos`) rather than repositioning the dialog window, which is unexpected UX. Please open a focused PR that only fixes the dialog positioning by adjusting the dialog's `move()` call to clamp the window within the screen bounds (e.g. using `QScreen.availableGeometry()`), without the pointList changes.

wkentaro:

Thanks for catching the off-screen label-dialog bug. The fix is bundled with unrelated unfinished work (a coordinate-list dock) and the branch conflicts; the dialog code has also been refactored (`popUp` -> `popup`, changed signature). The underlying bug is real, so I've opened #2189 to track clamping the dialog to the available screen geometry. Recommending close (your call to review or close).

PR #1463, opened 2024-06-21
Add an option to specify whether to apply a smooth to the picture.

Mirei124:
Add an option to specify whether to apply a smooth to the picture. It's useful when the picture is too small while there is a need of pixel-level precision annotation.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -548,6 +548,14 @@ class MainWindow(QtWidgets.QMainWindow):
             checked=self._config["keep_prev_scale"],
             enabled=True,
         )
+        smoothPixmap = action(
+            self.tr("&Smooth Pixmap"),
+            self.enableSmoothPixmap,
+            tip=self.tr("Whether smooth pixmap"),
+            checkable=True,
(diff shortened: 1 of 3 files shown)


wkentaro:

Thank you for the contribution! Unfortunately this PR has merge conflicts with the current main branch that would need to be resolved before it can be considered. Beyond that, the implementation ties Antialiasing and HighQualityAntialiasing (which affect shape rendering) to the same toggle as SmoothPixmapTransform (which affects image rendering) — disabling smooth pixmap would also degrade shape rendering quality, which is an unintended side effect. If you'd like to revive this, please rebase onto main and limit the toggle to only QPainter::SmoothPixmapTransform.

wkentaro:

Thanks for this. Since June 2024 the canvas rendering pipeline was rewritten (`paintEvent` now delegates to `_setup_world_transform`/`_render_layers`) and the config path moved to `labelme/_config/`, so the diff conflicts and the smoothing hook no longer exists where patched. This is already tracked by #2142 (option to disable image interpolation / nearest-neighbor when zoomed in), which covers the same need. Recommending close (your call to review or close).

PR #1472, opened 2024-08-04
Fix Side scroll request in the wheelEvent with the shift modifier in qt5 mode

Kokika:
This is fix for QT5.
Convient function for side scrolling on canvas with shift modifier. This idea copied other program behavior.

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -922,7 +922,7 @@ class Canvas(QtWidgets.QWidget):
                 self.zoomRequest.emit(delta.y(), ev.pos())
             elif QtCore.Qt.ShiftModifier == int(mods):
                 # side scroll
-                self.scrollRequest.emit(-delta.y(), QtCore.Qt.Horizontal)
+                self.scrollRequest.emit(delta.y(), QtCore.Qt.Horizontal)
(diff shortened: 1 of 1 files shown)


wkentaro:
@Kokika looks like qt5 ev.angleDelta() handles it automatically. (I confirmed on my macOS + PyQt5)

https://doc.qt.io/qt-5/qwheelevent.html

image

Kokika:
@wkentaro  I am testing on Ubuntu 24.04 with PyQt5 5.15.9 and here it isn't handles it automatically.

wkentaro:
I see, but your change breaks the other systems where qt5 automatically flips the order (because you flip again), so I can't merge.

We need to understand why qt5 won't work for you.

Kokika:
I realized it and I totally agree. I am searching the reason.

wkentaro:
@Kokika I tested on my Linux machine, and reproduced your phenomenon.

And on macOS, shift modifier is in mods (so your change makes it both horizontal scroll regardless of shift press).

Probably we need to add if-block for platform...

Kokika:
I tested this on Windows and same like Linux, so it is also need.

Kokika:
I added Mac os check for side scroll. @wkentaro, could you test it please?

Kokika:
Solved by https://github.com/wkentaro/labelme/pull/1513

PR #1474, opened 2024-08-13
Allow spacebar for finalizing a shape

akx:
Depending on where your hands are, it can be more convenient than Return.


akx:
@wkentaro Rebased.

akx:
Rebased.

wkentaro:
can you add tooltip for this? (along with for Return)

akx:
@wkentaro Can we merge this as is, and I'll follow up with a PR that adds proper key documentation in the status bar for all edit modes?

PR #1479, opened 2024-08-13
Allow enabling/disabling the label popup in the UI

akx:
Follows up on https://github.com/labelmeai/labelme/commit/34887aeef13390d1a2ae0835276e92c9d93548b6 (https://github.com/labelmeai/labelme/pull/271).

This makes the otherwise pretty obscure configuration setting visible in the label list instead:

When no label is selected (or available), the checkbox is disabled and forced on:

Screenshot 2024-08-13 at 23 29 11

When a label is selected in the list, the user is free to select the checkbox:

Screenshot 2024-08-13 at 23 29 22


wkentaro:
sorry for the late response. I agree this should be in the settings. maybe in the menu?

image

akx:
To be honest, workflow-wise, I think it would be nicer here, because you always see what the current mode is and it's one click to toggle the mode. In the View menu, it's not always visible, and it's two clicks (and additional cognitive load to find the option in the menu) to toggle the mode.

Maybe, if you're concerned about the visual real estate the checkbox takes up, it could be moved to the Label List panel header (assuming Qt supports that)...

akx:
@wkentaro rebased

wkentaro:

Thanks for this PR! Since you opened it, the `display_label_popup` setting has been added to the Settings dialog (General tab), so users can already toggle it from the UI without touching config files. Adding a second checkbox in the label dock panel would create duplicate control surface for the same setting. Combined with the current merge conflicts, I'm going to close this PR. If you think inline placement in the label dock is still better UX than the Settings dialog, feel free to open a new issue to discuss the design before implementing.

PR #1481, opened 2024-08-13
Allow holding shift to delete polys without confirmation

akx:
Like it says on the tin – if Shift is pressed, skip the confirmation popup for deleting polys.


labelme/app.py
```diff
@@ -1852,9 +1855,7 @@ def openPrevImg(self, _value=False):
 
     def openNextImg(self, _value=False, load=True):
         keep_prev = self._config["keep_prev"]
-        if QtWidgets.QApplication.keyboardModifiers() == (
-            Qt.ControlModifier | Qt.ShiftModifier
-        ):
+        if shift_and_ctrl_pressed():
```
wkentaro, line 1858: unnecessary change?

labelme/app.py
```diff
@@ -396,7 +398,10 @@ def __init__(
             self.deleteSelectedShape,
             shortcuts["delete_polygon"],
             "cancel",
-            self.tr("Delete the selected polygons"),
+            "{} {}".format(
+                self.tr("Delete the selected polygons"),
+                self.tr("(Hold Shift to skip confirmation)"),
+            ),
```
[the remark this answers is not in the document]
akx: Because I didn't want to invalidate existing translations.

labelme/app.py
akx: No - just using the same new function for the same functionality.

wkentaro: i mean, this pr is for deleting poly right? why you change keep_prev?

akx: It was a bit of a side effect – I think, back last year when I originally made this PR, I figured it'd be worth it to unify the modifier key functionality used both for "hold shift to delete polys" and "hold shift+ctrl" as used here, into canonical helper functions.

labelme/app.py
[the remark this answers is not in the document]
akx: Up to you. I'm fine doing it either this way or by changing the entire translation.

akx:

wkentaro:

## Agent Review

**Recommendation:** `recommend-close` (maintainer's call)

**What it does:** Adds a `shift_pressed()` helper and lets holding Shift skip the delete-confirmation dialog, plus a tooltip update and translations across all locales.

**Assessment:**
- The feature is genuinely useful and the core logic is ~3 lines, clean and well-typed.
- The 40-file diff is mostly translation churn (19 `.ts` + 19 compiled `.qm`); only `utils/qt.py` and `app.py` carry logic.
- The branch is currently **conflicting** with `main` — `app.py` has drifted since the PR was opened, so it would need a rebase.
- Minor nit: `shift_pressed()` reads the global `QApplication.keyboardModifiers()` snapshot rather than the event modifiers (the existing `canvas.py` idiom). It works for a keyboard-triggered action but is subtly different from the in-repo pattern.

**Why close rather than merge now:** The planned PySide6 migration (#2144) will heavily rewrite `app.py` and regenerate every translation file, so rebasing the 38 translation files now would be wasted effort. The cleanest path is to re-apply the small logic change on `main` after #2144 lands.

Thanks @akx for the contribution and the thorough translations.

wkentaro:

Thanks for this, @akx, and for the thorough translations.

We've decided to go a different route for skipping the delete confirmation. Rather than a Shift modifier, the confirmation dialog itself will get a "Don't ask again" checkbox, with the preference persisted and toggleable in Settings (currently in progress). That covers this use case more discoverably and lets users flip it back without learning a hidden shortcut.

Tracking that in #2156. Closing this in favor of that approach, but really appreciate the contribution.

PR #1482, opened 2024-08-13
Improve dock feature configuration

akx:
* Don't leak `features` in loop
  * The features were ANDed in to `features` in the loop, so configuring things to `false` might not have had an effect
* Don't crash if configuration is missing
* Don't use stringly typed `getattr` magic


akx:
This was apparently done in 4601ce27a8f0a6cf5328ab99b05f7cdc544d98b7.

PR #1484, opened 2024-08-13
Rephrase "Polygons" as "Shapes" where suitable

akx:
Duplicate/Delete/... will work on not-polygon shapes, so it's a bit strange that the actions talk about Polygons only.


wkentaro:
right. `flags` is also a part of "Annotation", so only calling `shapes` as "Annotation" is probably not consistent.

let's stick with "Shapes" for now, and come back when needed.

wkentaro:
can you also try to update translation file? otherwise, I'll do when I can.

image

wkentaro:
> Done (by hand). Are there instructions somewhere on how to regenerate those translation files?

Actually I also forgot how to do since last time. I'll make CI so we don't.

```
brew install qt@5 pyqt@5  # mac
pylupdate5 labelme/*.py labelme/widgets/*.py -ts labelme/translate/zh_CN.ts
lrelease labelme/translate/zh_CN.ts -qm labelme/translate/zh_CN.qm

```

wkentaro:
just merged https://github.com/wkentaro/labelme/pull/1636. can you please reset the change on the translation files?

akx:
Looks like this was done in 5717d98a799941a71497880db331f4eae3989b3f.

PR #1490, opened 2024-09-02
add rectangle select freature

txh0001:
select shapes by rectangle , just like autocad 

https://github.com/user-attachments/assets/2ffb9484-d209-4154-bbb8-6799e41e8732

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -334,6 +334,14 @@ class MainWindow(QtWidgets.QMainWindow):
             self.tr("Start drawing rectangles"),
             enabled=False,
         )
+        rectangleSelectMode = action(
+            self.tr("Rectangle Select"),
+            lambda: self.toggleDrawMode(False, createMode="selrectangle"),
+            shortcuts["rectangle_select"],
+            "objects",
+            self.tr("Start selecting rectangles"),
+            enabled=False,
+        )
         createCircleMode = action(
             self.tr("Create Circle"),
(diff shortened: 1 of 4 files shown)


maplelost:
awsome! but i think if you can use the edit polygon action would be better

wkentaro:

Thanks for the idea — rubber-band multi-select is a useful feature. Unfortunately this PR has merge conflicts with the current codebase that would need to be resolved, and the implementation has a few issues worth addressing before merge: (1) `selrectangle` is registered as a real shape type in `shape.py` and `default_config.yaml`, but it is only meant to be an ephemeral selection tool — it should not be persisted or serialized; (2) there is commented-out code left in `canvas.py`; (3) the selection logic only checks whether any single point of a shape falls inside the rectangle, which misses shapes whose body overlaps but whose vertices are outside. If you'd like to resubmit, please rebase on current `main`, remove the shape-type registration for `selrectangle`, clean up the commented lines, and fix the containment check. Happy to re-review a fresh PR.

wkentaro:

Thanks for proposing drag-to-select. The PR adds selection as a fake `shape_type` in the shape-creation pipeline, which conflicts with the rewritten typed canvas and isn't the right structure for a transient selection tool. Rubber-band select is a nice complement to the existing Ctrl+click multi-select, so I've opened #2188 to track it as a distinct interaction mode. Recommending close (your call to review or close).

PR #1496, opened 2024-09-19
Make point manipulation reasonable: ALT+Click (add), ALT+SHIFT+Click (delete)

wkentaro:
## Why?

1. Currently click creates a new point, which is inconvenient when you just want to select or move polygons.
2. Ctrl-Click also conflicts with group-selection.
3. So Alt+Click seems most reasonable.

NOTE: previously, SHIFT+Click was to "delete point", but for consistency, ALT+SHIFT+Click will be "delete point".

## Example

image

image


Kokika:
Great idea! I had a constant adding problem while wanting to delete small polygons.

PR #1502, opened 2024-09-28
Make point, line and two-point linestrip shapes selectable

Ynjxsjmh:
Currenly, point, line and two-point linestrip won't be selected by mouse left click, which leads to hard deletion. The main problem is that `self.makePath().contains(point)` doesn't consider these three special condition. These three shapes don't possess enclosing structures like rectangle and circle.

https://github.com/wkentaro/labelme/blob/b9c83ff54db8a83e6adfba77c08db9aad71bbbf1/labelme/shape.py#L326-L355

To solve this, I add judgement to check if the clicked point is near these shapes.

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -342,12 +342,17 @@ class Canvas(QtWidgets.QWidget):
                 self.hEdge = None
                 shape.highlightVertex(index, shape.MOVE_VERTEX)
                 self.overrideCursor(CURSOR_POINT)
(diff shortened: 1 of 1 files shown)


Ynjxsjmh:
@wkentaro Hi, could you take some time to view my two prs? I don't know how to trigger checks on github, but they pass the test on my local machine.

wkentaro:
resolved via https://github.com/wkentaro/labelme/pull/2107

PR #1503, opened 2024-09-29
Fix index error due to missing previously selected shapes in current image

Ynjxsjmh:
The problem can be reproduced by selecting a shape in current image `a`, and go on next image `b` then press backspace to delete point. `self.selectedShapes` is in image `a` thus it doesn't exist in any shapes in image `b`.

Another possible solution is to clear selected shapes after loading a new image, but user may want to go back to original image and use those selected shapes.


wkentaro:
resolved in main

Issue #1504, opened 2024-10-02
Side scroll not working in qt5 mode


Kokika:
### Provide environment information

Python 3.9.19

Package             Version
------------------- -----------
backports.tarfile   1.2.0
beautifulsoup4      4.12.3
certifi             2024.7.4
cffi                1.16.0
charset-normalizer  3.3.2
click               8.1.7
coloredlogs         15.0.1
contourpy           1.2.1
cryptography        43.0.0
cycler              0.12.1
darkdetect          0.7.1
docutils            0.21.2
exceptiongroup      1.2.2
filelock            3.15.4
flatbuffers         24.3.25
fonttools           4.53.1
gdown               5.2.0
github2pypi         1.0.0
humanfriendly       10.0
idna                3.7
imageio             2.34.2
imgviz              1.7.5
importlib_metadata  8.2.0
importlib_resources 6.4.0
iniconfig           2.0.0
jaraco.classes      3.4.0
jaraco.context      5.3.0
jaraco.functools    4.0.1
jeepney             0.8.0
keyring             25.2.1
kiwisolver          1.4.5
lazy_loader         0.4
markdown-it-py      3.0.0
matplotlib          3.9.1
mdurl               0.1.2
more-itertools      10.3.0
mpmath              1.3.0
natsort             8.4.0
networkx            3.2.1
nh3                 0.2.18
numpy               1.26.4
onnxruntime         1.18.1
packaging           24.1
pillow              10.4.0
pip                 24.0
pkginfo             1.10.0
pluggy              1.5.0
protobuf            5.27.3
pycparser           2.22
Pygments            2.18.0
pyparsing           3.1.2
PyQt5               5.15.9
pyqt5-plugins       5.15.9.2.3
PyQt5-Qt5           5.15.2
PyQt5_sip           12.15.0
pyqt5-tools         5.15.9.3.3
pyqtdarktheme       2.1.0
PySocks             1.7.1
pytest              8.3.2
pytest-qt           4.4.0
python-dateutil     2.9.0.post0
python-dotenv       1.0.1
PyYAML              6.0.1
qt5-applications    5.15.2.2.3
qt5-tools           5.15.2.1.3
QtPy                2.4.1
readme_renderer     44.0
requests            2.32.3
requests-toolbelt   1.0.0
rfc3986             2.0.0
rich                13.7.1
ruff                0.1.9
scikit-image        0.24.0
scipy               1.13.1
SecretStorage       3.3.3
[...]

PR #1513, opened 2024-11-13
fix: horizontal scroll when shift modifier is pressed

ardiya:
## Issue:

Both Scroll and Shift+Scroll resulted in vertical(up-down) scrolling in QT5.

It seems the else(QT5) condition properly handle the shift+scroll condition, ref: https://github.com/ardiya/labelme/blob/57fbf2b9a749cd77ac5f03f005ee6206d671ece8/labelme/widgets/canvas.py#L928-L929 So I just need to handle the QT5 condition.

## Expected result:

In Qt5, the Shift+Scroll should result in horizontal(left-right) scroll.

## Proposed Fix: This PR

When Shift modifier is pressed, use the y dimension of the scroll and use it to scroll horizontally.

Tested with shift and without shift, the code changes produces expected behavior.

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -908,6 +908,9 @@ class Canvas(QtWidgets.QWidget):
                 # with Ctrl/Command key
                 # zoom
                 self.zoomRequest.emit(delta.y(), ev.pos())
+            elif QtCore.Qt.ShiftModifier == int(mods):
+                # Convert wheel y(up down) direction to horizontal scroll with Shift key
+                self.scrollRequest.emit(delta.y(), QtCore.Qt.Horizontal)
             else:
                 # scroll
                 self.scrollRequest.emit(delta.x(), QtCore.Qt.Horizontal)


ardiya:
similar issue: https://github.com/wkentaro/labelme/issues/1504

minewilliam (approved):
Tested. This fixes the side scrolling problem on my Linux machine.

wkentaro:
Thanks for the PR. This is resolved on `main` - `Canvas.wheelEvent` now remaps Shift+wheel to horizontal scroll:

https://github.com/wkentaro/labelme/blob/main/labelme/widgets/canvas.py#L1456-L1472

```python
elif int(mods) == int(Qt.ShiftModifier) and delta.x() == 0:
    # Shift+wheel scrolls horizontally. macOS swaps the axis for us,
    # but Linux/Windows deliver the delta on y and expect the app to
    # remap it.
    self.scroll_request.emit(delta.y(), Qt.Horizontal)
```

The extra `delta.x() == 0` guard avoids double-scrolling on macOS, where the OS already converts Shift+wheel into a horizontal delta. Closing as resolved.

PR #1550, opened 2025-03-10
Use Osam for AI-mask and AI-polygon

wkentaro:
## Why?

So that we don't need separate download, model hosting. Just extend Osam would be enough.

This is a stepping stone for Sam2 support.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined AI-assisted annotation workflows with a simplified model selection in the UI and enhanced drawing integration for polygon and mask creation.
  
- **Chores**
  - Removed legacy segmentation and annotation functionalities.
  - Cleaned up external dependencies to simplify maintenance and installation.


wkentaro:
Closing in favor of https://github.com/wkentaro/labelme/pull/1551

PR #1578, opened 2025-04-06
UI enhancement

iamfat:
1. show label near the very top-left points
2. show flags in polygon label list

### DIFF
--- a/labelme/_config/default_config.yaml
+++ b/labelme/_config/default_config.yaml
@@ -61,6 +61,10 @@ fit_to_content:
 epsilon: 10.0
 canvas:
   fill_drawing: true
+  # Show shape label text near the top-left of each shape on the canvas
+  shape_label:
+    show: false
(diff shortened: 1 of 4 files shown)


wkentaro:

Thanks for the contribution! Unfortunately this PR has merge conflicts with the current main branch, and a few code-quality issues that would need to be addressed before it could be merged: (1) please resolve the conflicts by rebasing onto current main; (2) the `# modified by jia.huang` / `# modified end` attribution comments should be removed — git history tracks authorship; (3) the flag-display logic in `_edit_label` and `addLabel` is duplicated and should be extracted into a helper; (4) the label-on-canvas rendering in `shape.py` always draws at a fixed 16pt bold font with no way for users to disable it — a config option or at least a guard using the existing `show_labels` setting would be needed. If you'd like to revise and reopen, a rebase + those cleanups would make this much easier to review.

PR #1614, opened 2025-08-11
Fix: Scaled polygon (point, rectangle, circle) labels (= no hidding labels)

Vlczech:
* for the same aspect ratios polygon labels are scaled in the same proportion to the change in size
* for different aspect ratios polygon labels keep absolute position if it does not exceed image size
* if polygon label exceed image size, polygon labels are moved to left top corner


Vlczech:
@wkentaro

wkentaro:
ah, so the issue is like:

1. keep_prev is on
2. you scale in one image
3. you go to next image
4. the copied shape is not correct scale

Vlczech:
Scenario:
1. yes, keep previous annotation is on and fit windows is activated also
2. - 4. as you wrote
5. maybe? In my words: if first image is much higher than second, then previous annotations are hidden "somewhere right down" - so you need to unzoom image and find an annotation somewhere. 
Previous behavior: image
New one if proportions of image are same:
image

wkentaro:
@Vlczech let me find time to reproduce the issue and review this.

wkentaro:
@Vlczech "Keep previous labels" is designed to annotate video. Example: https://github.com/wkentaro/labelme/tree/main/examples/video_annotation#video-annotation-example

And zooming doesn't seem to be a problem in this example:
https://github.com/user-attachments/assets/ca4e4d5a-99bc-48e6-a68b-09bafae20272

So I guess your problem would be with images whose dimensions are different, which is outside of the original intent.
What scenario do you use this and encounter a difficulty in your use case?

Vlczech:
@wkentaro : Yes, you are right- My problems are with images (more precisely, for example, the folder with images in it), as I showed on example in one of previous comment. But I'm convinced, that my fix for basic shapes does not interfere with video behavior in any way (because the video has the same aspect ratio, but this is also taken into account for images)

wkentaro:
I see. now I understand your use case. I'll try mimicking your use case (by cropping images in a video), and better understand what the code is intended to do.

Vlczech:
@wkentaro: any news about this?

wkentaro:

Thank you for this contribution! Unfortunately, since this PR was opened the codebase has been substantially refactored — `loadShapes` is now `_load_shapes`, `loadFile` has been restructured, and main already includes `keep_prev_scale` support (toggled via a menu action). This PR now has unresolvable merge conflicts and the approach (mutating Shape points in-place in `adjustScale`) no longer fits the current architecture. If you'd like to reopen this as a fresh PR against current main, I'd suggest coordinating first with the maintainer on the intended design for scaled shape propagation in the refactored codebase.

wkentaro:

Thanks for tackling the keep_prev scaling issue. The PR targets the old `shape.py`/`loadShapes` API (now `_shape.py` numpy dataclass / `_load_shapes`) and conflicts, and the original logic has a few edge cases (non-uniform aspect ratios, polygons/linestrips). I've captured the underlying request in #2187 for a clean reimplementation. Recommending close (your call to review or close).

PR #1622, opened 2025-08-29
feat: allows dragging to move around the canvas

juliusl:
This enables holding down the middle mouse button to drag the canvas in order to move the scroll wheels.

Also, adds additional space to min-size so that the edges of the image can be more centered. (Not a Qt expert, but setting a layout didn't seem to fix the issue and this was the least intrusive option)

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1611,6 +1611,9 @@ class MainWindow(QtWidgets.QMainWindow):
 
         self.actions.fitWidth.setChecked(self._zoom_mode == _ZoomMode.FIT_WIDTH)
         self.actions.fitWindow.setChecked(self._zoom_mode == _ZoomMode.FIT_WINDOW)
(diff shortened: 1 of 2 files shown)


wkentaro:
hmm, but the offset makes any newly open image offset by default:

image

even if we decide to run "Fit Window" after the offset in the beginning, just 500 offset means we can't move the image that freely.

image

juliusl:
@wkentaro - I was wondering if you had any ideas on how to better implement that? I noticed if there isn't some extra space around the edges, it's not possible to center a portion of the image on the edges to the center of the canvas.

spoonytripod:
The dragging feature is what I really wanted. Thanks! It works greatly. I hope this feature will be merged into the original branch.

juliusl:
@wkentaro  so I think I fixed the issue, could you give it a try and let me know what you think?

PR #1626, opened 2025-09-01
Clean up most `type: ignore` comments

akx:
This PR cleans up most `type: ignore` comments, either by just removing them where not needed, or by fixing the surrounding code to not need them (e.g. via an `assert x` to verify `x` is not None).


akx:
@wkentaro Sure thing 👍 Rebased and removed the "interesting" changes from here, will follow up with another PR.

wkentaro:
@akx thanks, do we still need this? I thought it won't happen if we run mypy in python3.9

image

akx:
@wkentaro Just a moment, I'll check!

akx:
@wkentaro Good catch, must've slipped in while rebasing/splitting :) Took that out, mypy is still clean.

PR #1631, opened 2025-09-02
chore: remove flake8 noqa comments from __init__.py files

wkentaro:
Close https://github.com/wkentaro/labelme/pull/1475

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,9 @@ select = ["E", "F", "I"]
 [tool.ruff.lint.isort]
 force-single-line = true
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -73,6 +73,9 @@ select = ["E", "F", "I"]
 [tool.ruff.lint.isort]
 force-single-line = true
 
+[tool.ruff.lint.per-file-ignores]
+"__init__.py" = ["F401"]
```
wkentaro, line 77: @akx let's do like this.

PR #1636, opened 2025-09-06
ci: check translation files are up-to-date

wkentaro:
for https://github.com/wkentaro/labelme/pull/1484#issuecomment-3261363676 /cc @akx


akx:
Hmmm... Now every time some code gets moved around (line numbers change), the language files get out of date and CI fails. (This happened in #1632.) I can see that being very annoying very quickly 😄

PR #1637, opened 2025-09-06
ci: use cache-apt-pkgs-action to speed up CI

wkentaro:
follow up https://github.com/wkentaro/labelme/pull/1636


wkentaro:
follow up https://github.com/wkentaro/labelme/pull/1636

PR #1638, opened 2025-09-08
Move language-updating commands to Makefile

akx:
So they're easy to run locally, too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,11 +23,8 @@ jobs:
         version: 1.0
     - name: check translation files are up-to-date
       run: |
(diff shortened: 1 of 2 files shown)


akx:
wdyt @wkentaro?

wkentaro:
if pyqt sets line number, shouldn't we respect that? (they use that information on their gui-based editor)

image

akx:
Does the line number add value for our uses here? I'd just rather avoid hundred-line translation diffs every time any code is changed :)

wkentaro:
Hmm, make sense. Let's adopt your proposal here, and see how it goes.

wkentaro:
I added fancy logging:

image

wkentaro (approved):

PR #1639, opened 2025-09-08
Don't use re-exported names 

akx:
This kind of follows up on #1631 to get rid of re-exports in `__init__.py` files, so there's only one canonical place to import a name from, and switches import styles to `from ...module import name` instead of `from ... import module` everywhere.

If you don't think this is a good idea, that's fine, feel free to close it out!


wkentaro:
I think the question is which APIs are exposed or not. Nowadays, I don't like to expose much of the APIs (e.g., utils, LabelFile), so I like the direction. But I don't think we need to do all of them and all at once.

wkentaro:
let's do this slowly.

PR #1640, opened 2025-09-09
Add score display support for shapes

safa0:
## Summary
This PR adds support for displaying scores/confidence values on shapes in the labelme annotation tool.

## Changes
- Added `_draw_score_text()` method to the `Shape` class
- Shapes can now display text scores from `other_data['score_text']` 
- Score text is rendered in the upper-left corner of the shape's bounding box
- Text has a semi-transparent white background for better readability
- Uses 10pt bold font with black text

## Use Case
This feature is useful for:
- Displaying confidence scores from ML model predictions
- Showing detection scores when importing pre-labeled data
- Adding any numerical or text metadata to be displayed on shapes

## Implementation Details
The score display:
- Only renders when `score_text` exists in the shape's `other_data` dictionary
- Positions text in upper-left corner with 3px padding
- Includes semi-transparent background (rgba: 255, 255, 255, 200)
- Scales properly with zoom level
- Restores original painter pen after drawing

## Screenshots
The score text appears as a small label in the upper-left corner of annotated shapes when `other_data['score_text']` is set.

### DIFF
--- a/labelme/shape.py
+++ b/labelme/shape.py
@@ -281,6 +281,57 @@ class Shape:
             painter.setPen(pen)
             painter.drawPath(negative_vrtx_path)
             painter.fillPath(negative_vrtx_path, QtGui.QColor(255, 0, 0, 255))
+        
+        # Draw score text for any shape that has it
+        if self.points and len(self.points) >= 2 and self.other_data.get('score_text'):
+            self._draw_score_text(painter)
+
+    def _draw_score_text(self, painter):
+        """Draw score text in the upper-left corner of shape's bounding box."""
+        score_text = self.other_data.get('score_text', '')
+        if not score_text:
+            return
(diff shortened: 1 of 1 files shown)


wkentaro:

Thank you for this contribution! Unfortunately this PR can no longer be merged as-is: `labelme/shape.py` has since been split into `labelme/_shape.py` (data model) and `labelme/widgets/_shape_render.py` (painting logic), so the diff has unresolvable conflicts and the approach no longer fits the current architecture.

If you'd like to resurrect this feature, the right place to add score-text rendering is in `labelme/widgets/_shape_render.py` alongside the existing `_paint_shape_*` helpers, reading from `shape.other_data`. Feel free to open a new PR against current `main` with that approach — happy to review it.

PR #1281, opened 2023-05-20
support yolo format annotation file generation

RunqiuBao:
Screenshot from 2023-05-20 16-25-42

PR #1641, opened 2025-09-09
feat: yolo import/export support

safa0:
Adds support for converting to yolov8 format:
 - annotations.txt
 - classes.txt
 
 and vice versa.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -2032,11 +2032,11 @@ class MainWindow(QtWidgets.QMainWindow):
         )
         if not output_dir:
             return
-        
+
         try:
             # Use the same conversion logic as CLI
             success = self._export_to_yolo(self.filename, output_dir)
-            
+
             if success:
                 self.informationMessage(
                     self.tr("Export Complete"),
@@ -2051,7 +2051,7 @@ class MainWindow(QtWidgets.QMainWindow):
                         "Please check that the file contains valid annotations."
                     ),
                 )
-            
+
         except Exception as e:
             self.errorMessage(
                 self.tr("Export Error"),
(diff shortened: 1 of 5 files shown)


wkentaro:

Thank you for the contribution! Unfortunately this PR has several blocking issues that need author-side fixes before it can be considered:

1. **Merge conflicts** — the branch no longer applies on `main`. Please rebase.
2. **CI has never run** — no checks are reported on this branch.
3. **Correctness bug in class-ID assignment** — in both `_export_to_yolo` (app.py) and `cli/export_yolo.py`, `class_names` is a `set()` that grows inside the loop, and `class_id = sorted(class_names).index(label)` is recomputed after each `add()`. Because earlier labels can shift position as new labels are inserted into the sorted order, the class IDs written to the `.txt` file can be inconsistent with the final `classes.txt`. Build the label-to-id map in a single pass after collecting all labels, then emit annotations.
4. **Undeclared dependency** — `cli/import_yolo.py` calls `from yolo2labelme import yolo2labelme` but `yolo2labelme` is not listed in `pyproject.toml` dependencies. Either add it or reimplement the import path without the external package.
5. **Duplicated logic** — the export conversion code is copy-pasted verbatim between `app.py` and `cli/export_yolo.py`. Extract a shared helper (e.g. `labelme/converters/yolo.py`) and call it from both.
6. **Scope overlap with PR #1281** — that older PR also adds YOLO export. It would help to clarify how this PR relates to or supersedes it.

Given the number of rework items (rebase, bug fix, dependency declaration, deduplication) and the absence of any CI signal, I'm recommending close rather than leaving it open indefinitely. Please feel free to reopen a fresh PR once the issues above are addressed.

PR #1642, opened 2025-09-10
ci: ignore line numbers in translation file check

wkentaro:
image

https://github.com/wkentaro/labelme/pull/1636#issuecomment-3264790957

@akx

### DIFF
--- a/tools/check_ts_files.sh
+++ b/tools/check_ts_files.sh
@@ -1,0 +1,44 @@
+#!/bin/bash -e
+
+TEMP_FILES=()
+
+cleanup() {
+  for file in "${TEMP_FILES[@]}"; do
+    if [[ -f "$file" ]]; then
+      rm "$file"
+    fi
+  done
+}
+
+trap cleanup EXIT
+
+function check_translation_file() {
+  if [ $# -ne 1 ]; then
+    echo "Usage: $0 <translation_file.ts>"
+    exit 1
+  fi
+  local ts_file=$1
+
+  echo "Checking '$ts_file' is up to date..."
+  cp "$ts_file" "${ts_file/.ts/.a.ts}"
+  TEMP_FILES+=("${ts_file/.ts/.a.ts}")
+  cp "$ts_file" "${ts_file/.ts/.b.ts}"
+  TEMP_FILES+=("${ts_file/.ts/.b.ts}")
+  #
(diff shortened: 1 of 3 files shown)


tools/check_ts_files.sh
```diff
@@ -0,0 +1,44 @@
...
+  local ts_file=$1
+
+  echo "Checking '$ts_file' is up to date..."
+  cp "$ts_file" "${ts_file/.ts/.a.ts}"
+  TEMP_FILES+=("${ts_file/.ts/.a.ts}")
+  cp "$ts_file" "${ts_file/.ts/.b.ts}"
+  TEMP_FILES+=("${ts_file/.ts/.b.ts}")
+  #
+  pylupdate5 "$LABELME_DIR/*.py" "$LABELME_DIR/widgets/*.py" -ts "${ts_file/.ts/.b.ts}"
+  diff <(sed -r 's/line="[0-9]+"//g' "${ts_file/.ts/.a.ts}") <(sed -r 's/line="[0-9]+"//g' "${ts_file/.ts/.b.ts}")
```
akx, line 29: I'm not sure I'm following what's happening here. Why is the pylupdate5 line twice here, aside from quoting?

tools/check_ts_files.sh
```diff
@@ -0,0 +1,44 @@
...
+
+  echo "Checking '${ts_file} -> .qm' is up to date..."
+  lrelease "$ts_file" -qm "${ts_file/.ts/.b.qm}" > /dev/null
+  TEMP_FILES+=("${ts_file/.ts/.b.qm}")
+  diff "${ts_file/.ts/.qm}" "${ts_file/.ts/.b.qm}"
+
+  echo "Finished checking '$ts_file'."
+}
```
akx, line 37: ... in fact, I'm not sure I understand what's going on in this file at all 😄 

What's the end goal to check here?

tools/check_ts_files.sh
wkentaro: is it twice? I am intended to call only once:

image

tools/check_ts_files.sh
wkentaro: so after this PR, you shouldn't need to change ts file for the change like: https://github.com/wkentaro/labelme/pull/1632

akx:
Honestly, it could be better to zero out the line numbers in all translation files (or see if `pylupdate5` has an option to not emit line numbers. Right now any change that moves lines around while adding a translation ends up with a huge, huge diff.

PR #1655, opened 2025-09-17
Fix brightness/contrast for PNG image

wkentaro:
## Why?

image

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1603,7 +1603,7 @@ class MainWindow(QtWidgets.QMainWindow):
 
     def brightnessContrast(self, value):
         dialog = BrightnessContrastDialog(
-            utils.img_data_to_pil(self.imageData),
+            utils.img_data_to_pil(self.imageData).convert("RGB"),
             self.onNewBrightnessContrast,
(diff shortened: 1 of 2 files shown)


labelme/app.py
```diff
@@ -1601,23 +1601,40 @@ def enableKeepPrevScale(self, enabled):
     def onNewBrightnessContrast(self, qimage):
         self.canvas.loadPixmap(QtGui.QPixmap.fromImage(qimage), clear_shapes=False)
 
-    def brightnessContrast(self, value):
+    def brightnessContrast(self, value: bool, is_initial_load: bool = False):
+        del value
+
         dialog = BrightnessContrastDialog(
-            utils.img_data_to_pil(self.imageData),
+            utils.img_data_to_pil(self.imageData).convert("RGB"),
```
akx, line 1608: This will lose transparency in the image. Wouldn't it be better to extract the RGB bands, run brightness/contrast on them, then splat in the original alpha?

wkentaro: yes, that sounds better. https://github.com/wkentaro/labelme/issues/1660

PR #1710, opened 2025-11-22
Clear previous selection on switch to prevent bugs

XuKang371322:


wkentaro:
can you share what issue specifically you've encountered?

XuKang371322:
> can you share what issue specifically you've encountered?

After I select  shapes on the current image （Select multiple items with the Ctrl key）and then click the shape of the next image , the application raises an error and crashes. To prevent this, I think it's essential to clear all shape selections when the image switch is triggered by clicking on a shape from the next image."

Erotemic:
This might be related to the bug I found: https://github.com/wkentaro/labelme/pull/1716

PR #1711, opened 2025-11-24
fix(app): use larger epsilon to hide scrollbar for fit-window

wkentaro:
happens on linux and large window


wkentaro:
the real issue is the fact that the image moves when we highlight shape on linux

PR #1716, opened 2025-11-26
Fix crash

Erotemic:
Had this occur when pressing the down arrow in the file list.  Got that the selected shape was not in the list not sure exactly why this happened the first time, but I can reproduce it.

To reproduce:

* Open folder with multiple images and polygon annotations. One polygon in the first image, no polygons in the second. 
* click a polygon
* select image in file list
* press down arrow
* press up arrow

This patch prevents the crash, but the up arrow does not respond. However, if you just click on another image the program recovers gracefully.


wkentaro:
reproduced, thanks:

image

PR #1778, opened 2026-01-20
feat: oriented rectangle

minewilliam:
# Feature
A rectangular labeling tool with support for rotation based on this [discussion](https://github.com/wkentaro/labelme/discussions/1374).

Screencast_20260311_172359

## Usage
- Draw an oriented rectangle using 3 points.
- The first line drawn determines the angle of the oriented rectangle.
- Drag a corner to move it.
- Adjacent corners are automatically projected during a corner move to maintain the rectangle.
- Drag edge midpoints to rotate the rectangle.
- Use the arrow visual in the center of the oriented rectangle as a reference for the direction of the shape.

- [ ] Support drag outside of canvas
- [ ] Orientation arrow svg
- [ ] Fix drawing mode where current line is overdrawn.

## Implementation Details
- Vertex moving is bounded to the canvas, like any other shape.
- The shape can be rotated beyond the bounds of the canvas. Blocking rotation would be user-unfriendly.
- Moving the shape keeps it bounded to the canvas.

# Not implemented
- This shape is not bounded by the canvas. Rotation behavior becomes complex and user-unfriendly. Vertex bounding checks are left in the code, but are not used for the time being.
- Click the orientation arrow to quickly "cycle" (i.e. flip) the direction of the shape.


wkentaro:
Hi @minewilliam, thanks for opening this and for the discussion in #1374. It's what got this feature moving.

The `oriented_rectangle` shape landed in #1980, so it's now on `main`. The final implementation took a different path from this branch, so I'll close this one rather than rebase. Thanks again.

PR #1782, opened 2026-01-22
Refactor canvas highlight function

minewilliam:
# Description
This PR aims to centralize the shape highlighting logic in the canvas.
To this end, the `self.update()` call has been moved outside of `setHighlight()`. The function will return `True` if the caller should call `self.update()`.

To prevent multiple calls to `self.update()`, the `deSelectShape()` function has also been modified in the same manner.

# Context
The upcoming [rectangle rotation](https://github.com/wkentaro/labelme/discussions/1374) feature ([PR](https://github.com/wkentaro/labelme/pull/1778)) will need to add a new kind of points to highlight; rotation points.

The upcoming logic for rotation points will add a parameter to `setHighlight` for the index of the highlighted rotation point (`hRotationPoint`), like such:
```python
def setHighlight(self, shape: Shape | None, hEdge = None, hVertex = None, hRotationPoint = None) -> bool:
    ...
    self.hRotationPoint = hRotationPoint
    ...
```

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -267,6 +267,7 @@ class Canvas(QtWidgets.QWidget):
         if self.hShape:
             self.hShape.highlightClear()
             need_update = True
+        # NOTE: Store last highlighted for adding/removing points.
         self._lasthShape = self.hShape if hShape is None else hShape
         self._lasthVertex = self.hVertex if hVertex is None else hVertex
         self._lasthEdge = self.hEdge if hEdge is None else hEdge


labelme/widgets/canvas.py
```diff
@@ -433,17 +440,12 @@ def mouseMoveEvent(self, a0: QtGui.QMouseEvent) -> None:
...
             if index is not None:
-                if self.selectedVertex() and self.hShape:
-                    self.hShape.highlightClear()
-                self.prevhVertex = self.hVertex = index
-                self.prevhShape = self.hShape = shape
-                self.prevhEdge = self.hEdge
-                self.hEdge = None
+                self.setHighlight(shape, hVertex=index)
+                self.prevhShape = shape
+                self.prevhVertex = index
```
minewilliam: Kept logic as-is. I assume we overwrite the previous hShape/hVertex/hEdge on purpose.

Note: Since assignations are applied from right to left, the following:
```python
self.prevhShape = self.hShape = shape
```
is the same as:
```python
self.hShape = shape
self.prevhShape = self.hShape
```

wkentaro: thanks, looks like it was a (almost harmless) bug before. let me fix it.

PR #1829, opened 2026-03-05
fix: normalize 16-bit grayscale to uint8 in export_json

### LINKED ISSUE #1528: json_to_dataset fails if image is mono 16-bit

with latest labelme_json_to_dataset.py script and the .exe as well

### What OS are you using?

windows 10

### Describe the Bug

script fails with 'data type must be np.uint8'
the problem is solved by adding this code before label2rgb():

import numpy as np

    img = imgviz.asgray(img)
# asgray does NOT convert mono (gray) 16-bits to 8-bits  
# so do it here:
    if img.dtype != np.uint8:
        img =  (img/(np.max(img)/256)).astype('uint8')
#
    lbl_viz = imgviz.label2rgb(
        lbl, img, label_names=label_names, loc="rb"
    )

### Expected Behavior

the script should be able to handle mono 16-bit images

### To Reproduce

run the script with a json file with a 16-bit image

wkentaro:
## Summary

When  processes 16-bit grayscale images,  returns a `uint16` array. Passing this directly to `imgviz.label2rgb()` causes a crash because the function expects `uint8` input.

## Fix

Normalize `img_gray` to `uint8` before calling `label2rgb` when the dtype is not already `uint8`:

```python
img_gray = imgviz.asgray(image)
if img_gray.dtype != np.uint8:
    img_gray = (img_gray / (img_gray.max() / 255.0)).astype(np.uint8)
lbl_viz = imgviz.label2rgb(lbl, img_gray, label_names=label_names, loc="rb")
```

## Changes

- `labelme/cli/export_json.py`: extract `asgray()` result, normalize to uint8 if needed

Fixes #1528

### DIFF
--- a/labelme/cli/export_json.py
+++ b/labelme/cli/export_json.py
@@ -46,9 +46,7 @@ def main():
     for name, value in label_name_to_value.items():
         label_names[value] = name
 
-    img_gray = imgviz.asgray(image)
-    if img_gray.dtype != np.uint8:
-        img_gray = (img_gray / (img_gray.max() / 255.0)).astype(np.uint8)
+    img_gray = imgviz.normalize(imgviz.asgray(image))
     lbl_viz = imgviz.label2rgb(lbl, img_gray, label_names=label_names, loc="rb")
 
     PIL.Image.fromarray(image).save(osp.join(out_dir, "img.png"))


wkentaro:
@claude what about using imgviz.normalize?

wkentaro:
@copilot what about using imgviz.normalize?

PR #1899, opened 2026-03-25
Show AI model download progress in GUI

kimchioverfit:
The AI polygon model download flow previously exposed progress only in the terminal, while the GUI remained in an indeterminate loading state without showing a percentage.

This change bridges gdown download progress into the Qt download dialog by emitting progress updates from the background worker and applying them to a determinate QProgressDialog. 

The dialog now shows percentage and transferred size in real time.

For models made up of multiple blobs, the dialog also indicates which blob is being downloaded. Auto-close and auto-reset were disabled so the dialog stays visible until the full download completes, rather than closing after an intermediate 100% update.

### DIFF
--- a/labelme/widgets/download.py
+++ b/labelme/widgets/download.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import contextlib
+import importlib
 import types
 
 import osam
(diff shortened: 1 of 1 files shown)


wkentaro:
Thank you for the proposal. Closing via https://github.com/wkentaro/labelme/pull/1948

PR #1949, opened 2026-04-15
fix: improve TIFF rendering for satellite imagery

wkentaro:
## Summary
- Route all TIFF files through `tifffile` directly, bypassing PIL which misinterprets extra bands (e.g. NIR) as alpha transparency
- Normalize display channels with shared percentile stretch (p2/p98) instead of per-channel min/max, preserving inter-channel color balance
- Skip normalization for uint8 TIFF data that is already in display range

## Why
PIL interprets 4-band uint8 TIFFs (common in satellite imagery like NAIP: R, G, B, NIR) as RGBA, treating the NIR band as alpha. This makes the image look "smoked out" / semi-transparent. Additionally, the previous per-channel independent min/max normalization destroyed relative brightness between channels, producing washed-out colors.

Ref: https://github.com/wkentaro/labelme-satellite-image-demo/issues/3

## Test plan
- [x] All 125 tests pass (`make test`)
- [x] Lint clean (`make lint`)
- [x] Type check clean (`uv run ty check`)
- [x] Verified with real NAIP GeoTIFF — outputs RGB JPEG instead of PNG-with-alpha
- [x] Visual check in labelme UI with multi-band GeoTIFF

### DIFF
--- a/labelme/_label_file.py
+++ b/labelme/_label_file.py
@@ -289,42 +289,42 @@ _DISPLAYABLE_MODES = {"1", "L", "P", "RGB", "RGBA", "LA", "PA"}
 
 
 def _imread(filename: str) -> PIL.Image.Image:
-    ext: str = osp.splitext(filename)[1].lower()
-    try:
-        image_pil = PIL.Image.open(filename)
-        if image_pil.mode not in _DISPLAYABLE_MODES:
-            raise PIL.UnidentifiedImageError
-        return image_pil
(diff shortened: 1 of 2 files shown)


wkentaro:
Closing — the NAIP 'smoked out' rendering is a data problem (ExtraSamples mislabels NIR as alpha), not a labelme bug. Labelme correctly respects the TIFF metadata.

PR #1961, opened 2026-04-16
feat: highlight AI buttons on hover and equalize toolbar widths

wkentaro:
## Summary
- Highlight AI-Points and AI-Box toolbar buttons with a yellow background and border when hovering the disabled AI-Assisted Annotation panel, guiding users to select the correct mode
- Tooltip and button highlight use matching `#FFFFCC` color for visual consistency
- Equalize button widths in the vertical toolbar for a cleaner layout

## Why
When the AI-Assisted Annotation widget is disabled, users need a visual hint about which toolbar buttons to click. The hover highlight draws attention to AI-Points and AI-Box buttons alongside the existing tooltip message.

## Test plan
- [x] Verified highlight color matches tooltip color visually
- [x] Verified no font size regression from stylesheet changes
- [x] Verified no flickering (transparent border applied when not highlighted)
- [x] Verified highlight triggers from header label area, not just combo boxes
- [x] Verified highlight clears on mouse leave and on mode switch

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -13,6 +13,7 @@ import time
 import webbrowser
 from collections.abc import Callable
 from pathlib import Path
+from typing import Final
 from typing import Literal
 from typing import NamedTuple
 from typing import TypeAlias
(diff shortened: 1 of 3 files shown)


wkentaro:
image

Issue #1968, opened 2026-04-20
When using in AI points ,I can not add more control Points in the existing polygon


ZhangDonglin12:
In Editing AI-Points model ,When I want do some change in the AI-Points , I can not create more points in the exiting AI polygon
  when i left click -twice to add new Control point in the AI-points polygon, it can not work

<img width="1066" height="572" alt="Image" src="https://github.com/user-attachments/assets/32987c54-377a-42ed-8b7b-14cf31dcc265" />

wkentaro:
Can you give me more detail to reproduce? Alt-click on the edge to add point is working for me.

https://github.com/user-attachments/assets/708796a7-e4c7-4133-95ed-9d4a55e92c6d

ZhangDonglin12:
> Can you give me more detail to reproduce? Alt-click on the edge to add point is working for me.
> 
>  export-1776688848887.mp4
Thank you so much,  I get it

for the old version ,I just left click the edge can add the points in the polygon.

wkentaro:
> for the old version ,I just left click the edge can add the points in the polygon.

yes that was a bug and fixed by https://github.com/wkentaro/labelme/pull/1496

wkentaro:
@ZhangDonglin12 fyi, https://github.com/wkentaro/labelme/pull/2029 will add the right-click menu

<img width="377" height="547" alt="Image" src="https://github.com/user-attachments/assets/24acbf4d-38c7-44c1-addd-76cec57816f6" />

PR #2029, opened 2026-04-28
feat(app): add right-click "Add Point to Edge" context menu action

wkentaro:
UI/UX like https://github.com/wkentaro/labelme/issues/1968

## Summary
- New right-click context-menu action **Add Point to Edge** in the canvas, enabled when the cursor is hovering over a polygon edge.
- New `Canvas.edge_selected` PyQt signal, mirroring the existing `vertex_selected` pattern that already drives **Remove Selected Point**.
- Action wired in `MainWindow._setup_actions` and inserted into `_actions.context_menu`. Existing ALT+Click shortcut still works.
- E2E test `test_add_point_via_context_menu_action` covers the disabled→enabled transition on hover and verifies the inserted point lands at the hovered edge midpoint.
- Translations filled for all 19 languages (.ts + recompiled .qm).

## Why
Closes #2022. Users could only add a point on an existing polygon via the ALT+Click hint shown in the status bar — easy to miss. The right-click menu is the conventional discoverability path.

## Test plan
- [x] `make lint` (ruff format + check, ty type-check, taplo, mdformat, yamlfix, typos, translation diff)
- [x] `make test` — 72 unit + 100 e2e passing, including the new test
- [x] Manual verification: hover edge → right-click → click **Add Point to Edge** inserts a vertex; cursor moving from canvas into the popup menu does not disable the action.


wkentaro:
image

PR #2035, opened 2026-04-28
feat(label-list): propagate hide/show across multi-selection

wkentaro:
## Summary
- Stacked on top of #2034.
- When several rows in the label list are selected, toggling one row's visibility checkbox now propagates the same state to every selected row, captured as a single undoable step.
- Adds a `mousePressEvent`/`mouseReleaseEvent` pair on `LabelListWidget` so multi-selection survives a checkbox click (Qt would otherwise collapse it to the clicked row).
- Adds e2e tests for both the propagation and the selection-preservation behavior.

## Test plan
- [x] `uv run pytest tests/e2e/visibility_test.py`
- [x] `make lint`

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1690,15 +1690,37 @@ class MainWindow(QtWidgets.QMainWindow):
                 self._canvas_widgets.canvas.update()
 
     def _on_label_item_changed(self, item: LabelListWidgetItem) -> None:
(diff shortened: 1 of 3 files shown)


wkentaro:
https://github.com/user-attachments/assets/6473313a-60ee-4a10-a783-bfa9fc07fdc4

PR #2049, opened 2026-05-05
refactor(shape,canvas): rework translate and drag-bound clamp

wkentaro:
## Summary
- Replace the `if/elif` chain in `_make_shape_path` with a `_PATH_BUILDERS` dispatch table (one builder per shape kind).
- Drop `Shape.to_path` wrapper; call `_make_shape_path` directly from `Shape.bounds`.
- Rename `Shape.move_by` to `Shape.translate` and use an `enumerate`-based in-place loop.
- Rename `Shape.bounding_rect` to `Shape.bounds` (matches the `QRectF` return type, no abbreviation).
- Replace `Canvas.calculate_offsets` / `Canvas.bounded_move_shapes` with `_record_drag_anchor` / `_drag_shapes`. The drag anchor is now `(rel_top_left, bounds: QRectF)`, and clamping is expressed as direct `max`/`min` on the bbox top-left rather than two corner offsets with `min(0, ...)`.

## Why
The previous shape API mixed naming styles (`bounding_rect`, `to_path`, `move_by`) and the drag-bound clamp encoded its constraint via a two-corner-offset trick that took mental decoding. The dispatch table makes it trivial to add new shape types, and the clamp now reads as "keep the bounds top-left inside the pixmap rectangle".

## Test plan
- [x] `make lint`
- [x] `make test` (201 passed)
- [x] Manual smoke: drag a shape near the pixmap edge in the GUI; bounded movement still works

### DIFF
--- a/labelme/shape.py
+++ b/labelme/shape.py
@@ -5,6 +5,7 @@ from collections.abc import Iterable
 from typing import Any
 from typing import Final
 from typing import NamedTuple
+from typing import Protocol
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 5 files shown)


wkentaro:
Splitting into two focused PRs (Shape API refactor + Canvas drag-bound clamp redesign).

PR #2055, opened 2026-05-05
refactor(config): rename display_label_popup → prompt_label_on_create

wkentaro:
## Summary
Rename the \`display_label_popup\` config key to \`prompt_label_on_create\`.

The new name reads better at the call site (\"prompt for a label when a shape is created\") and matches the surrounding boolean-config naming convention. Predicate operands at the use site are also reordered to put the \"first shape always gets a label\" rule first.

A migration shim is added in \`_migrate_config_from_file\` so existing \`~/.labelmerc\` files continue to work without user action.

## Test plan
- [x] \`make lint\` clean
- [x] \`make test\` (201 passed)
- [ ] Smoke-test: existing \`~/.labelmerc\` with the old key still works (migration logs the rename and the new key takes effect)

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -1742,7 +1742,7 @@ class MainWindow(QtWidgets.QMainWindow):
         flags = {}
         group_id = None
         description = ""
-        if self._config["display_label_popup"] or not text:
+        if not text or self._config["prompt_label_on_create"]:
             previous_text = self._label_dialog.edit.text()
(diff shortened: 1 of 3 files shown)


wkentaro:
Closing — re-evaluated cost/benefit: a config-key rename for one borderline boolean YAML entry plus a cosmetic operand-swap on the predicate isn't worth the migration-shim churn and the user-visible behavior change. Keeping the existing key name.

PR #2068, opened 2026-05-07
refactor(canvas): drop unused helpers and inline thin wrappers

wkentaro:
Tightens the Canvas API by dropping methods that were either unused outside `canvas.py` or thin one-line wrappers around an attribute / mode comparison.

Per commit:
- inline `Canvas.fill_drawing()` getter; `_fill_drawing` is read directly internally
- drop `Canvas.is_shape_visible(shape)`; callers read `shape.visible` directly
- inline `Canvas.drawing()`; callers compare `self.mode == CanvasMode.CREATE`
- inline `Canvas.editing()`; callers compare `self.mode == CanvasMode.EDIT`

No external callers (verified across `labelme/app.py` and tests). Public setter `set_fill_drawing` is unchanged.

## Test plan
- [x] `make lint`
- [x] `make test` (207 passed)

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -247,9 +247,6 @@ class Canvas(QtWidgets.QWidget):
         self._release_cursor()
         self._update_status()
 
-    def is_shape_visible(self, shape: Shape) -> bool:
-        return shape.visible
-
     def drawing(self) -> bool:
         return self.mode == CanvasMode.CREATE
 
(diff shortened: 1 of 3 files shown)


wkentaro (commented):
Self-review summary (cannot self-approve):

Decision: **would APPROVE** — clean, mechanical simplification.

**Verified**
- No external callers of `fill_drawing()`, `is_shape_visible()`, `drawing()`, `editing()` in `labelme/` or `tests/` (grep). Public setter `set_fill_drawing` is unchanged; `app.py:1055` references a different `drawing_polygon` signal, not the removed method.
- `make lint`: pass.
- `make test`: 207 passed.

**Notes (LOW, non-blocking)**
- `self.mode == CanvasMode.CREATE` / `EDIT` is now repeated ~14 times. Fine as-is, but if the comparison spreads further, a read-only property could justify itself again later — just flagging since the deletions here explicitly chose to drop those wrappers.
- `_fill_drawing` is now read directly inside the class (`preview.fill = self._fill_drawing`, etc.). Fine within the class body; the public setter is preserved.

PR #1980, opened 2026-04-25
feat: oriented rectangle shape

wkentaro:
Reimplementation of #1778 on top of latest `main`, adapted to the recent canvas refactor and split into reviewable commits.

A new `oriented_rectangle` shape drawn with three clicks: the first edge fixes the orientation, the third corner is projected at a right angle to keep the shape rectangular. Corners drag with parallelogram-completion; edge midpoints rotate the shape around its centre; a small arrow indicates orientation.

## Test plan

- [x] `make lint` clean
- [x] `make test` (229 passed) including new unit + e2e cases
- [x] Manual smoke: draw, drag corner, drag midpoint to rotate, save, reload

## Inherited follow-ups (out of scope)

- Drag past the canvas edge during creation/rotation
- Render orientation arrow as SVG
- Eliminate overdrawn current-line in drawing mode

PR #2078, opened 2026-05-09
feat: oriented_rectangle in AI-assisted annotation

wkentaro:
Adds `oriented_rectangle` as a fourth AI output format alongside polygon / mask / rectangle / circle.

The SAM mask is reduced to its convex hull, then a rotating-calipers pass picks the minimum-area enclosing rectangle (each hull edge tested as a candidate orientation). Sign disambiguation pins the long axis to the right (or lower) half-plane so the corner sequence is deterministic. When no mask is available the handler falls back to the axis-aligned bbox corners, matching the circle-from-bbox fallback.

Stacks on #1980 (the `oriented_rectangle` shape type itself).

## Test plan

- [x] `make lint` clean
- [x] `make test` (243 passed) — new unit tests cover empty / single-pixel / collinear guards, axis-aligned recovery (wider- and taller-than-tall), 30°-rotated long-axis recovery (within ±3°), and L-shape / square edge cases that catch the PCA failure mode; new e2e parameter `ai_box-oriented_rectangle` exercises the full SAM → min-area-rect pipeline
- [x] Manual: pick AI-Box mode, choose Oriented Rectangle output, draw on a tilted object, confirm orientation arrow points along the long side

### DIFF
--- a/labelme/_automation/_geometry.py
+++ b/labelme/_automation/_geometry.py
@@ -4,6 +4,7 @@ from typing import NamedTuple
 
 import imgviz
 import numpy as np
+import scipy.spatial
 import skimage
 from loguru import logger
 from numpy.typing import NDArray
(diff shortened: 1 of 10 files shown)


wkentaro:
https://github.com/user-attachments/assets/3a5cea04-b73d-449e-b950-b4236791ac1b

PR #2082, opened 2026-05-12
feat(ai-box): suppress duplicate detections from box prompts

wkentaro:
## Summary
Skip propagated AI-Box detections whose mask overlaps an existing shape (IoU >= 0.5).

## Why
Sam3 returns multiple detections from a single box prompt, so drawing around an already-annotated object would duplicate it. Mask IoU is used (not bbox IoU) so polygon outputs with similar bboxes but distinct geometry are not falsely treated as duplicates.

## Test plan
- [x] `make test` (247 passed)
- [x] `make lint`
- [ ] Manual: draw an AI-Box over an already-annotated object and confirm no duplicate is created


wkentaro:
Closing in favor of #2087, which bundles this change with mask-aware NMS over new detections (needed for non-object cases like trees).

PR #2087, opened 2026-05-12
feat(ai): mask-aware NMS for AI-assisted annotation

wkentaro:
## Summary
Supersedes #2082. The mask-NMS-among-new-detections part already landed via #2088, so this PR is now scoped to a single behaviour: skip a propagated AI detection whose mask IoU vs an already-annotated shape is >= 0.5.

Applied in the unified AI-box/AI-points canvas flow (`_shapes_from_ai_points`) after the existing `suppress_detections_greedy` pass.

## Why
Sam3 returns multiple detections from a single prompt. Drawing an AI box (or clicking points) over an already-annotated object would otherwise duplicate it.

Mask IoU (not bbox IoU) so polygon outputs with similar bboxes but distinct geometry are not falsely treated as duplicates.

## Implementation
- `shape_to_xyxy_bbox` moved from `labelme.app` into `labelme._automation._geometry` (now public via `_automation.shape_to_xyxy_bbox`); used both by the new suppression and by the existing AI-text NMS path.
- `suppress_detections_overlapping_existing_shapes` lives in `labelme._automation._suppression`, sharing `_LocalMask` / intersection helpers with `suppress_detections_greedy`.
- Shape rasterisation supports rectangle / polygon / oriented_rectangle / circle / mask.

## Test plan
- [x] `make test` (267 passed)
- [x] `make lint`
- [x] Manual: draw AI-box over an already-annotated object — no duplicate
- [x] Manual: AI-points click on an already-annotated object — no duplicate


wkentaro:
https://github.com/user-attachments/assets/6c48a554-180f-4461-a7e8-f17748269bef

PR #2088, opened 2026-05-13
feat(ai): suppress redundant mask-overlapping detections

wkentaro:
Greedy per-label mask NMS using both IoU and intersection-over-smaller (containment) thresholds, so the same region returned at multiple nested SAM granularities (e.g. tree-cluster vs single tree) deduplicates to one detection. Plain bbox NMS cannot deduplicate nested masks that share a bbox.

Applied in the AI-box canvas flow, the AI-points canvas flow, and the AI-text app flow.

Stacked on top of #2091.

## Test plan

- [x] `make lint`
- [x] `make test` (253 passed)

### DIFF
--- a/labelme/_automation/__init__.py
+++ b/labelme/_automation/__init__.py
@@ -1,6 +1,7 @@
 from ._osam_session import OsamSession
 from ._shape_builders import Detection
 from ._shape_builders import shapes_from_detections
+from ._suppression import suppress_detections_greedy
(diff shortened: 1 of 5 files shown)


wkentaro:
TODO: create a demo video, and merge.

wkentaro:
https://github.com/user-attachments/assets/7d534481-1e9a-4d56-8feb-52c40b5af338

PR #2101, opened 2026-05-25
fix(canvas): ramp grab-pan overscroll continuously past viewport

wkentaro:
The slack added to the canvas size for grab-pan padding was a step function: `0` while the image fit the viewport, then `viewport.width() // 2` the instant `scaled_w` crossed the viewport width. Because the image is centered in the canvas, its widget-coord offset jumped by `V/4` at that exact zoom step, visible as the image suddenly snapping sideways during zoom.

The fix ramps the slack linearly with the overflow, capped at `V/2`, and floored at `V/8`. The cap preserves the existing guarantee that any edge can ultimately be panned to the viewport center. The floor keeps middle-drag pan responsive at slight overflow: without it, the scroll range equals the overflow, so a 2-px-overflowing image feels locked under the cursor.

The floor still leaves a discontinuity at the threshold, but a much smaller one: the image shift at the crossover drops from `V/4` to `V/16` (4x smaller than before).

## Test plan

- [x] new parametrized unit test covers the threshold, the floor region, the linear ramp, and the cap saturation
- [x] existing `tests/e2e/middle_drag_scroll_test.py` still passes (depends on overscroll being present)
- [x] `make lint`

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -1386,12 +1386,11 @@ class Canvas(QtWidgets.QWidget):
         viewport = self._scroll_viewport()
         if viewport is None:
             return QtCore.QSize(scaled_w, scaled_h)
-        # Overscroll only along axes where the image actually overflows the
-        # viewport. Half a viewport of slack (split evenly around the centered
-        # image) lets each edge be panned a quarter-viewport past the viewport
-        # boundary, derived from the viewport rather than a fixed multiplier.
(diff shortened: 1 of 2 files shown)


wkentaro:
## Before -> After


https://github.com/user-attachments/assets/68b9824b-1fd4-471f-a102-75b88eaab80f


https://github.com/user-attachments/assets/95879d47-65ca-476b-bb83-0a253e7b628c

PR #2110, opened 2026-06-01
Add mask brush editing controls

kimchioverfit:
## Summary

- Add mask brush controls for edit mode
- Support circle and rectangle brush shapes with adjustable brush size
- Allow the erase brush to remove polygon points that fall within the brush area
- Keep brush actions disabled outside edit mode

## Tests

- `python -m pytest tests/unit/widgets/canvas_test.py -k "brush"`


kimchioverfit:
Currently, erasing a polygon with the mask brush only removes vertices that fall within the brush radius. 

This means edges passing through the brush area are unaffected, making the tool unreliable on low-density polygons.

A more robust approach would be to compute the boolean difference between the polygon and the brush footprint (circle or rect), replacing the affected edges with new intersection vertices. 

This is how CVAT and Supervisely handle brush-based polygon editing

wkentaro:

Thanks for this contribution, @kimchioverfit! The implementation looks solid — brush add/erase for masks and polygon point erasing is a useful feature, the code is clean, and all CI checks pass across platforms.

The only blocker is that the branch currently has merge conflicts with `main` (the PR's `mergeStateStatus` is `DIRTY`). Could you please rebase onto the latest `main` and force-push to resolve the conflicts? Once that's done, this should be ready for merge.

wkentaro:

Thanks for this, mask brush editing is a genuinely useful feature. Unfortunately the PR can't land as-is: it's now conflicting, and its canvas code and tests are written against the old `Shape` API (mutable `QPointF` point lists with `close()`/`add_point()`/`clear_highlight()`), whereas `Shape` is now a numpy-array-backed dataclass with none of those methods. The work would need a substantial rewrite against the current `_shape.py` model and the PySide6 canvas. I've captured the feature in #2184 for a fresh implementation. Recommending close (your call to review or close).

kimchioverfit:
Got it, thanks for the review and the clarification. I'll close this PR.

PR #2113, opened 2026-06-03
feat: add reopen-last-folder and language settings

wkentaro:
Part 3 of 3 of the in-app Settings stack. Stacked on #2112; the base of this PR is `settings/2-dialog`, so the diff here is the two new settings only. Merge after #2112.

## Summary
- Add "Reopen last folder on launch" (General): persist the last opened directory in the window-state store and reopen it on the next launch when enabled, unless a path is passed on the command line.
- Add "Language" (General): a picker listing the bundled translations by native name, persisting a locale code that the startup translator loads. It shows an inline "Takes effect after restart" caption, via a new optional per-setting note.
- Language is restart-to-apply on purpose: labelme installs the translator once at startup and has no runtime retranslation, so the note surfaces that inline rather than letting the change appear to do nothing.

## Test plan
- [x] tests added: language editor (default selects system, applies a locale code, unknown code falls back), reopen-on-launch e2e
- [x] `make lint`
- [x] `uv run pytest tests/` (378 passed)


wkentaro:

Most of this branch has already landed in main through other PRs: the comment-preserving config I/O, the schema-driven Settings dialog, live-apply, the settings ADR/context docs, and the Language setting are all present now, and `labelme/config/` has since been migrated to `labelme/_config/` with a new schema format. The only piece not yet in main is the `reopen_last_dir` feature, which now needs reimplementing against the migrated architecture. I have tracked that in #2193. Recommending close as superseded; keeping it as a draft for now.

PR #2119, opened 2026-06-06
chore: improve labelme maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/conftest.py, tests/e2e/conftest.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -67,12 +67,19 @@ def assert_labelfile_sanity(filename: str) -> None:
         data = json.load(f)
 
     assert "imagePath" in data
+    assert isinstance(data["imagePath"], str)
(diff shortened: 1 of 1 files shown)


wkentaro:

Thanks for the stronger label-file sanity assertions. The CLA check is failing (it needs to be signed before any contribution can be merged) and the lint check is also failing. The added type assertions are reasonable; a maintainer could fold the idea directly into `tests/conftest.py`. Recommending close (your call to review or close).

PR #2120, opened 2026-06-06
feat: add a minimal schema-driven Settings dialog

wkentaro:
## Summary
- Adds a Settings dialog driven by a declarative schema (`labelme/_config/_schema.py`): each `Setting` names a config key path, section, label, and editor kind (`bool` / `enum` / `str_list`).
- Edits apply immediately and are written back to `~/.labelmerc` as sparse, comment-preserving overrides via the ruamel.yaml writer from #2114 — no OK/Apply/Cancel.
- Replaces the "Preferences… (open YAML in a text editor)" menu item with "Settings…"; the dialog still offers "Open config file as text…". It is disabled when the config comes from a `--config` expression or per-session CLI overrides (nothing safe to write back to).
- `labels` live-applies by rebuilding the label dialog; the `validate_label` "exact" choice is gated on having predefined labels.

## Scope
Deliberately minimal first slice: prove the schema → dialog → immediate-apply → sparse-write loop end to end, then migrate the rest one-by-one. It exposes three config-only settings (`display_label_popup`, `labels`, `validate_label`), so it touches no existing menu actions, and uses stock Qt widgets.

#2112 (draft) carries the fuller version and stays as reference. Follow-ups land on top:
- migrate `auto_save` / `with_image_data` / `canvas.fill_drawing` from their menu actions into the dialog (each drops a menu action + regenerates translations)
- visual polish (custom stylesheet, animated toggle switch)

## Test plan
- [x] `tests/unit/_config/schema_test.py` — schema ↔ default-config contract
- [x] `tests/unit/widgets/settings_dialog_test.py` — all three editor kinds, validate-label gate, flush-on-close
- [x] `tests/e2e/settings_test.py` — open / disabled-with-overrides paths + persist-and-apply round-trip
- [x] `ruff check` + `ty check` clean


wkentaro:
B97C126A-8304-4C8C-930A-0C4FE04A3897

PR #2138, opened 2026-06-08
fix: keep image on canvas when deleting label file

### LINKED ISSUE #2126: Delete File clears the whole canvas (image disappears), not just the annotations

## Bug

**Delete File** removes only the current label file (its tooltip is *"Delete current label file"*, and it `unlink()`s the annotation `.json`, not the image). But after deleting, the entire canvas goes blank — the image disappears too, not just the annotations.

This was noticed while triaging #834 (whose crash is separately resolved).

### Steps to reproduce

1. Open a directory containing an image with an existing label file.
2. With the image loaded, trigger **File -> Delete File** and confirm.
3. The annotation file is deleted, and the canvas is now empty.

### Actual behavior

The image vanishes from the canvas; only an empty canvas remains.

### Expected behavior

Deleting the label file should clear the annotations (shapes / label list) but keep the current image displayed, since the image itself was not deleted.

### Cause

`delete_file()` calls `reset_state()`, which clears the image state (`image_path`, `image_data`) and resets the canvas pixmap to an empty `QPixmap`, in addition to clearing the shapes and label list. To keep the image visible, the image needs to remain loaded (or be reloaded) while only the annotations are cleared.

wkentaro:
**Delete File** deletes only the current label `.json`, but the canvas went blank afterwards because `delete_file()` called `reset_state()`, which also clears the image state and resets the canvas pixmap. Since the image itself is not deleted, it should stay on screen.

Fix: after unlinking the label file, clear only the shapes and label list and keep the loaded image (and its `_image_path` / `_annotation`, so a later re-annotation can still be saved).

## Test plan
- [x] `tests/e2e/file_operations_test.py::test_delete_label_file_keeps_image` — after delete, canvas pixmap is non-null and enabled, shapes/label list are empty, image state is retained
- [x] `make lint`
- [x] `make test` (358 passed)

Closes #2126.

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -2341,7 +2341,11 @@ class MainWindow(QtWidgets.QMainWindow):
         if item:
             item.setCheckState(Qt.Unchecked)
 
-        self.reset_state()
+        # Only the label file was deleted, not the image: clear the annotations
+        # but keep the image on the canvas.
+        self._docks.label_list.clear()
+        self._canvas_widgets.canvas.load_shapes(shapes=[], replace=True)
+        self.mark_clean()
 
     @property
     def _is_settings_editable(self) -> bool:
--- a/tests/e2e/file_operations_test.py
+++ b/tests/e2e/file_operations_test.py
@@ -68,3 +68,40 @@ def test_delete_label_file(
     assert item.checkState() == Qt.Unchecked
 
     close_or_pause(qtbot=qtbot, widget=win, pause=pause)
+
+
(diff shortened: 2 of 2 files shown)


wkentaro:

Finalized: rebased on `main` (already up to date), CI green. The new `test_delete_label_file_keeps_image` GUI test runs in CI (Xvfb) and verifies the image stays on the canvas while annotations are cleared. Undrafted and recommending merge.

PR #2139, opened 2026-06-08
fix: fall back to image basename when label file's imagePath fails to load

### LINKED ISSUE #2125: Fall back to image basename when label file's imagePath fails to load

Tracking issue for the image-path basename fallback originally proposed in #1035 by @drjasonharrison. That PR's diff predates a refactor of the label-file reader and no longer applies, so the fix is re-specified here for a fresh implementation. Credit for the original idea goes to @drjasonharrison.

## Agent Brief

**Category:** bug
**Summary:** When a label file's recorded `imagePath` can't be loaded, fall back to the image's basename next to the label file, so datasets that move between machines still open.

**Context / why this is re-specified:**
The original patch (#1035) targets `labelme/label_file.py`, which has since been refactored away; the image-loading logic now lives in the module-level label-file reader. The patch no longer applies. The *idea* is still valid, so this should be implemented fresh against the current code.

Note for whoever picks this up: labelme's own save path has written a **relative** `imagePath` (via `os.path.relpath`) since 2018, so labelme itself does not emit absolute paths. The failing files this fix helps are produced by old labelme versions, external tooling, or manual edits — that's the real source, not the current write path. Do **not** change the write side.

**Current behavior:**
When a label file has no embedded image data, the reader resolves `imagePath` against the label file's own directory and loads the image bytes. If `imagePath` is absolute, the absolute path takes precedence and is attempted. If that single attempt fails, loading raises the label-file read error with no fallback. A dataset annotated on one machine (absolute path, or a relative path with a now-stale subdirectory) and copied to another therefore fails to open even when the image sits right next to its `.json`.

**Desired behavior:**
When the primary resolution fails to load the image, retry once using the label file's directory joined with **only the basename** of `imagePath`. If that also fails, raise the existing read error, and have its message reference both attempted paths. The happy path (primary resolution succeeds) must be unchanged. Label files that embed image data (`imageData` present) must be unaffected — the fallback applies only to on-disk image loads.

**Key interfaces:**
- The module-level label-file reader that resolves `imagePath` relative to the label file's directory and reads the image bytes — this is where the fallback belongs.
- The image-read helper it calls — decide whether the fallback lives in the reader or the helper; keep the existing read-error type as the failure signal.
- No change to the save/write path or to `imagePath` serialization.

**Acceptance criteria:**
- [ ] A label file whose `imagePath` resolves to an existing image loads with no behavioral change.
- [ ] A label file with an absolute `imagePath` that does not exist, but whose basename exists alongside the `.json`, loads successfully via the basename fallback.
- [ ] A label file with a relative `imagePath` containing a stale subdirectory (e.g. `../old/img.jpg`) whose basename exists next to the `.json` loads via the fallback.
- [ ] When neither the primary path nor the basename path resolves, loading raises the existing read-error type and the message names both attempted paths.
- [ ] Label files with embedded `imageData` are unaffected (no fallback path executed).
- [ ] Tests cover the happy path, the absolute-path fallback, the stale-relative fallback, and the both-fail error.

**Out of scope:**
- Changing the write/save side or `imagePath` serialization.
- Searching anywhere other than the label file's own directory (no recursive/parent-tree search).

wkentaro:
A label file with no embedded image data resolves its `imagePath` against the label file's directory. When a dataset moves between machines, that path can be stale (an absolute path, or a relative path with a now-missing subdirectory) even though the image still sits next to its `.json`, and the load failed outright.

Fix: when the primary resolution fails, retry once using the label file's directory joined to **only the basename** of `imagePath`. If that also fails, raise the existing `LabelFileReadError`, whose message now names both attempted paths. The happy path, the write/serialization side, and label files with embedded `imageData` are all unchanged.

Re-implements the idea from #1035 by @drjasonharrison against the current module-level reader (the original patch no longer applied after a refactor).

## Test plan
- [x] `tests/unit/_label_file_test.py` — absolute-path fallback, stale-relative fallback, both-fail error names both paths, and embedded-`imageData` left untouched
- [x] existing happy-path read test unchanged
- [x] `make lint`
- [x] `make test` (361 passed)

Closes #2125.

### DIFF
--- a/labelme/_label_file.py
+++ b/labelme/_label_file.py
@@ -220,6 +220,26 @@ def _check_image_dimensions(
         )
 
 
+def _read_image_with_basename_fallback(*, label_dir: Path, image_path: str) -> bytes:
+    # imagePath can be stale when a dataset moves between machines (absolute
+    # path, or a relative path with a now-missing subdirectory). The image often
+    # still sits next to its label file, so retry once with just the basename.
+    primary = label_dir / image_path
+    try:
+        return read_image_file(filename=str(primary))
+    except OSError as primary_error:
+        fallback = label_dir / Path(image_path).name
+        if fallback == primary:
+            raise
+        try:
+            return read_image_file(filename=str(fallback))
+        except OSError as fallback_error:
+            raise OSError(
+                f"failed to load image from {str(primary)!r} "
+                f"or {str(fallback)!r}: {fallback_error}"
+            ) from primary_error
+
+
 def read_label_file(filename: str) -> Annotation:
     try:
         with open(filename, encoding="utf-8") as f:
@@ -228,8 +248,8 @@ def read_label_file(filename: str) -> Annotation:
         if raw["imageData"] is not None:
             image_data = base64.b64decode(raw["imageData"])
         else:
-            image_data = read_image_file(
-                filename=str(Path(filename).parent / image_path)
+            image_data = _read_image_with_basename_fallback(
+                label_dir=Path(filename).parent, image_path=image_path
             )
         _check_image_dimensions(
             image_data=image_data,
--- a/tests/unit/_label_file_test.py
+++ b/tests/unit/_label_file_test.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import base64
 import json
 import shutil
 from collections.abc import Callable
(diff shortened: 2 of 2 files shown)


wkentaro:

Finalized: rebased on `main` (already up to date), CI green across all platforms (including the new `_label_file_test.py` cases, which run under Xvfb on Ubuntu). The basename-fallback logic chains errors correctly and is covered by tests for absolute-path, stale-relative-path, both-missing, and embedded-`imageData` cases. Undrafted and recommending merge.

wkentaro:

Superseded by #2162, which re-specifies the same #2125 fix more completely: it raises `LabelFileReadError` instead of a bare `OSError`, logs a warning on basename substitution, and adds tests for the warning and the no-fallback bare-basename miss. Recommending this be closed in favor of #2162 (your call).

PR #2163, opened 2026-06-13
fix: round linestrip mask joints to remove notches at turning points

wkentaro:
`shape_to_mask` rendered a `linestrip` as a single wide `draw.line`. PIL's wide-line rasterizer draws each segment as its own rectangle and does not fill the wedge where two segments meet, so the generated mask had an unfilled notch at every interior vertex (most visible at acute angles, as in the screenshot on #760/#2124).

Passing `joint="curve"` makes PIL round each joint to the line half-width, leaving the joint solid. Other `shape_type` branches are untouched, and a collinear linestrip is unchanged (no real joint to round).

This re-implements just the notch fix from the stale PR #761 by @nico-zck, without the unrelated changes that PR bundled.

Note: `joint="curve"` was added in Pillow 5.3 (2018) and is a no-op for `line_width <= 4`. The declared `pillow>=2.8` floor is already understated (the codebase uses `PIL.UnidentifiedImageError`, Pillow 7.0+), so the feature is satisfied in practice; I left the stale floor as a separate pre-existing concern rather than bump it here.

## Test plan
- [x] tests added: `tests/unit/utils/shape_test.py` - an acute-V linestrip has its notch filled (fails on `main`), and adding a collinear vertex does not change the mask
- [x] `uv run pytest tests/unit` (232 passed)
- [x] `uv run ruff check` and `uv run ty check` clean

### DIFF
--- a/labelme/utils/shape.py
+++ b/labelme/utils/shape.py
@@ -41,7 +41,8 @@ def shape_to_mask(
         assert len(xy) == 2, "Shape of shape_type=line must have 2 points"
         draw.line(xy=xy, fill=1, width=line_width)  # ty: ignore[invalid-argument-type]
     elif shape_type == "linestrip":
-        draw.line(xy=xy, fill=1, width=line_width)  # ty: ignore[invalid-argument-type]
+        # joint="curve" rounds the joints so wide lines have no notch at turns.
+        draw.line(xy=xy, fill=1, width=line_width, joint="curve")  # ty: ignore[invalid-argument-type]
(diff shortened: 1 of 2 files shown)


wkentaro:

Finalized: mergeable on `main`, CI green across all platforms. The fix is a one-line `joint="curve"` on the `linestrip` branch (other shape types untouched), covered by a notch-fill test that fails on `main` and a collinear-invariance test. Undrafted and recommending merge.

PR #2164, opened 2026-06-13
feat: update predefined labels in LabelDialog in place, preserving session history

wkentaro:
## Summary
- `LabelDialog` tracked predefined labels (`config["labels"]`) and session label-history (accumulated via `add_label_history`) in one merged `QListWidget` with no way to distinguish them. Live-applying a labels edit (from the Settings dialog, PR #2120) therefore rebuilt the whole dialog and dropped session history.
- Track the two sources separately (`_predefined_labels`, `_label_history`) and add `set_predefined_labels()`, which rebuilds the visible list in place from their union. Mutating the existing list widget (rather than reconstructing the dialog) keeps the completer's model binding intact.
- The main window now calls `set_predefined_labels()` instead of reconstructing the dialog: removing a predefined label drops it from suggestions unless it was used this session, and session-used labels survive the edit.
- Guard `_on_label_selected` against the `None` that `QListWidget.clear()` emits through `currentItemChanged` (would otherwise raise in the slot during the in-place rebuild).

## Test plan
- [x] unit tests added (`tests/unit/widgets/label_dialog_test.py`): live add, live remove, history preserved across an edit, removed-but-session-used label kept, empty predefined list keeps history, completer model stays bound, and no crash when an item is selected during the update
- [x] updated `tests/e2e/settings_test.py` to assert the dialog is updated in place (not reconstructed) while still preserving history
- [x] `uv run pytest tests/unit tests/e2e/settings_test.py` (all green)

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -2380,20 +2380,11 @@ class MainWindow(QtWidgets.QMainWindow):
 
     def _apply_to_live_widgets(self, key_path: tuple[str, ...]) -> None:
         if key_path[0] == "labels":
-            # Rebuilding from config alone would drop labels accumulated via
-            # add_label_history (learned from loaded/created shapes), so carry the
-            # existing dialog's labels into the new one. Like the dock below, this
-            # keeps removed predefined labels until restart.
-            label_list = self._label_dialog.label_list
-            previous_labels = [
-                cast(QtWidgets.QListWidgetItem, label_list.item(i)).text()
-                for i in range(label_list.count())
-            ]
-            old_label_dialog = self._label_dialog
-            self._label_dialog = self._make_label_dialog()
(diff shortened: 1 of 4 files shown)


wkentaro:

Finalized: mergeable on `main`, CI green across all platforms. Reviewed the source change: `set_predefined_labels()` mutates the existing `QListWidget` in place (union of predefined + session history, order-preserving dedup via `dict.fromkeys`) so the completer model binding survives, `add_label_history` now tracks history separately, and `_on_label_selected` guards the `None` that `clear()` emits via `currentItemChanged`. Covered by the new unit tests (live add/remove, history preservation, completer-stays-bound, None-selection) and the updated e2e test. Undrafted and recommending merge.

PR #2165, opened 2026-06-13
feat: give the zoom widget 0.1% precision for smoother stepping

wkentaro:
## Summary
- The zoom level widget was a `QSpinBox` limited to integer percent, so on large images a 1% increment (whether stepping via zoom in/out or typing a value) was a perceptibly large jump.
- Switch `ZoomWidget` to `QDoubleSpinBox` with one decimal of precision (e.g. `150.5 %`), addressing both stepping and direct-entry granularity.
- Widen the zoom value through `_set_zoom` / `_zoom_values` from `int` to `float`, and drop the `math.ceil` / `math.floor` clamping in `_add_zoom`: multiplicative stepping now keeps sub-percent precision, and the `QDoubleSpinBox` rounds to its configured decimals. The now-unused `import math` is removed.

## Note for review
With `setDecimals(1)`, round zoom levels render as `100.0 %` rather than `100 %`. This matches the issue's acceptance criterion ("displays one decimal place of precision"). If you'd prefer to suppress the trailing `.0` for whole-number zoom levels, that's a small `textFromValue` override I can add as a follow-up.

## Test plan
- [x] unit tests added (`tests/unit/widgets/zoom_widget_test.py`): default 100, accepts one decimal (`150.5`), rounds to one decimal, clamps to range
- [x] e2e test added (`tests/e2e/zoom_test.py`): a multiplicative step keeps fractional precision (`105 * 1.1 = 115.5`, where the old integer widget clamped to `116`)
- [x] existing `tests/e2e/zoom_test.py` cases still pass
- [x] `uv run pytest tests/unit` (234 passed) and the zoom e2e suite

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import enum
 import functools
 import json
-import math
 import os
 import platform
 import re
@@ -192,7 +191,7 @@ class MainWindow(QtWidgets.QMainWindow):
     _label_file_path: str | None
     _image_path: str | None
     _prev_image_path: str | None
-    _zoom_values: dict[str, tuple[_ZoomMode, int]]
+    _zoom_values: dict[str, tuple[_ZoomMode, float]]
     _brightness_contrast_values: dict[str, tuple[int | None, int | None]]
     _scroll_values: dict[Qt.Orientation, dict[str, float]]
     _default_state: QtCore.QByteArray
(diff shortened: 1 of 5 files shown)


wkentaro:

Finalized: mergeable on `main`, CI green across all platforms. Reviewed the source: `ZoomWidget` becomes a `QDoubleSpinBox` (1 decimal), the zoom value types widen `int`→`float` through `_set_zoom`/`_zoom_values`, the `math.ceil`/`math.floor` clamping in `_add_zoom` is dropped (the spinbox rounds to its decimals), and the now-unused `import math` is removed (ruff confirms). Covered by new unit tests (default/one-decimal/rounding/clamp) and an e2e test asserting a multiplicative step keeps fractional precision. Undrafted and recommending merge.

Issue #2123, opened 2026-06-06
Re-add flags to the Settings dialog with in-place flag dock refresh


wkentaro:
## What to build

The Settings dialog (PR #2120) live-applies every setting it exposes, but the
`flags` setting was excluded because the flag dock is rendered once and only
repopulates on the next image load, so editing flags would leave the visible
dock stale until the user navigates. Add `flags` back to the dialog once the
flag dock refreshes in place: when the `flags` config changes, rebuild the flag
dock contents immediately against the current image, preserving any flags the
current image already has checked.

This is the same shape as the `labels` live-apply path, which rebuilds its
widget in `_apply_to_live_widgets` on change. `flags` should follow suit instead
of waiting for the next image load.

## Acceptance criteria

- [ ] Editing Predefined image flags in the Settings dialog updates the flag
      dock for the currently loaded image without navigating away
- [ ] Flags already checked on the current image stay checked after the update
- [ ] The `flags` Setting is restored to `schema.SETTINGS` under the Labels section
- [ ] `_apply_to_live_widgets` handles the `flags` key path with an in-place dock refresh

## Blocked by

None - can start immediately

PR #2169, opened 2026-06-13
feat: re-add image flags to the Settings dialog with live flag-dock refresh

wkentaro:
## Summary
- The `flags` setting was excluded from the live-apply Settings dialog (PR #2120) because the flag dock only repopulates on the next image load, so editing flags left the visible dock stale.
- Add a `flags` (Predefined image flags) `str_list` Setting under the Labels section. The dialog's generic `str_list` editor already wires the apply path, so no `settings_dialog.py` change is needed.
- Handle the `"flags"` key in `_apply_to_live_widgets`: refresh the flag dock against the current image **additively** — add newly predefined flags (unchecked) and keep every flag already in the dock with its checked state. Extract `_read_flag_dock_states` to share the dock-reading logic with `save_labels`.

## Note for review
The refresh is intentionally **additive**: a flag *removed* from the predefined list lingers in the dock until the next image load (mirroring the existing label docks). This is deliberate so a settings edit never silently drops a flag the current image actually carries (the save path reads the dock, so dropping an unchecked flag would lose it from the `.json`). If you'd prefer removed-and-unchecked flags to disappear immediately, that's a small change, but it trades away that data-safety property.

## Test plan
- [x] e2e test added (`tests/e2e/settings_test.py`): editing the flags setting updates the dock live (no navigation), a newly added flag appears, a checked flag stays checked across a further edit, and the settings-driven refresh does not mark the image dirty (`not win._is_changed`)
- [x] `uv run pytest tests/unit tests/e2e/settings_test.py tests/e2e/annotation_test.py` (266 passed)
- [x] schema parametrized tests still pass; `uv run ruff check`, `uv run ty check` clean


wkentaro:

Finalized. CI was failing only on missing translations for the new `Predefined image flags` setting string (lint regenerates `.ts`/`.qm` and rejects `unfinished`). I filled translations for all 20 locales (matching each files existing `Predefined …` terminology) and regenerated; force-pushed.

The implementation reads well: the `str_list` editor already wires the apply path, and the additive flag-dock refresh (keep dock flags + their checked state, add newly-predefined unchecked) is the right call since the save path reads the dock, so a non-additive refresh could silently drop a flag the current image carries. CI green all platforms, mergeable. Undrafted and recommending merge.

wkentaro:

Rebased onto latest main. Resolved conflicts from the PySide6 migration: `save_labels` now uses the extracted `_read_flag_dock_states()`, the new `Qt.Checked` usages were updated to `Qt.CheckState.Checked`, the test imports moved to PySide6, the schema `label=` was wrapped in `cast(str, ...)`, and translations were regenerated. Full `make lint` and the settings e2e tests (8) pass; CI is green (closes #2123). Recommending merge.

PR #2170, opened 2026-06-13
refactor(app): drop unused queue_event and has_labels methods

wkentaro:
Two `MainWindow` methods have no callers anywhere in `labelme/`, `tests/`, or
`examples/`:

- `queue_event` — a one-line wrapper over `QtCore.QTimer.singleShot(0, ...)`
  left behind after its call site was removed.
- `has_labels` — the "you must label at least one object to save" guard, dead
  since `6a9913af` ("allow saving file without any labels") deliberately removed
  it from the save path.

Removing `queue_event` also orphaned `from collections.abc import Callable`
(its only use), and the `# Message Dialogs. #` section comment headed only the
removed `has_labels`, so both go too. No behavior change: the methods were
unreferenced, including via dynamic dispatch and Qt signal connections.

## Test plan
- [x] `grep` confirms zero references to either method across the repo
- [x] `uv run python -c "import labelme.app"` imports cleanly
- [x] `uv run pytest tests/unit` (239 passed)
- [x] `uv run ruff check` and `uv run ty check` clean

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -10,7 +10,6 @@ import re
 import subprocess
 import time
 import webbrowser
-from collections.abc import Callable
 from pathlib import Path
 from typing import Final
 from typing import Literal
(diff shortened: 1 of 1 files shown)


wkentaro:

Finalized: mergeable on `main`, CI green all platforms. Independently verified both methods are dead: a repo-wide grep finds only their `def` sites, no callers (including Qt signal connections / dynamic dispatch). Removing `queue_event` correctly orphans the `Callable` import and the lone `# Message Dialogs.` section comment, both dropped. Undrafted and recommending merge.

Issue #2167, opened 2026-06-13
Cleanup follow-ups for LabelDialog.set_predefined_labels (from #2164)


wkentaro:
Minor follow-ups noted during review of #2164 (merged). All scoped to `labelme/widgets/label_dialog.py`. None are correctness bugs.

### 1. Dead `_predefined_labels` initialization (actionable)
`__init__` sets `self._predefined_labels = list(labels or [])`, but the only reader is `set_predefined_labels`, which reassigns the field *before* reading it. The constructor value is therefore never observed.

Pick one:
- Drop the `__init__` assignment and have `set_predefined_labels` read its `labels` argument directly (no longer needs the field), or
- Keep it as deliberate explicit-initialization and leave a one-line note.

### 2. `list(labels)` → `labels[:]` (actionable, trivial)
In `set_predefined_labels`, `labels` is typed `list[str]`, so copy with `labels[:]` per the repo style (reserve `list(...)` for iterable→list conversion). The `__init__` `list(labels or [])` is the documented `T | None` exception and should stay as-is.

A drafted one-line change for this existed in the PR worktree but never made it into the merged commit.

### 3. Manual reorder lost on update when `sort_labels=False` (open question)
When sorting is disabled the list uses `InternalMove` drag-drop; `set_predefined_labels` does `clear()` + `addItems(merged)`, discarding any user reordering. This is **not a regression** (the previous dialog-reconstruction approach also lost it), and with the default `sort_labels=True` it never triggers. Decide whether preserving reorder is worth the complexity, or close as wontfix.

PR #2172, opened 2026-06-13
refactor: drop the write-only _predefined_labels field from LabelDialog

wkentaro:
Cleanup follow-up from the review of #2164 (merged).

`LabelDialog._predefined_labels` was assigned in `__init__` and then reassigned by `set_predefined_labels` before it was ever read, so the constructor value was never observed - the field was effectively write-only. Remove it and have `set_predefined_labels` use its `labels` argument directly. The `labels + self._label_history` concatenation already produces a fresh list, so the previous `list(labels)` copy is unnecessary. No behavior change; the existing `LabelDialog` unit tests and the settings e2e tests cover it unchanged.

This addresses items 1 and 2 of #2167. Item 3 (preserving manual drag-reorder when `sort_labels=False` across a predefined-labels edit) is left as **wontfix**: it only applies to the non-default `sort_labels=False` mode, was already lost by the previous dialog-reconstruction approach (not a regression), and preserving it would add real complexity for a narrow case. Happy to revisit if desired.

## Test plan
- [x] `uv run pytest tests/unit/widgets/label_dialog_test.py tests/e2e/settings_test.py` (14 passed, behavior unchanged)

### DIFF
--- a/labelme/widgets/label_dialog.py
+++ b/labelme/widgets/label_dialog.py
@@ -44,7 +44,6 @@ class LabelDialog(QtWidgets.QDialog):
         self._fit_to_content = fit_to_content
         self._sort_labels = sort_labels
         self._flags = flags or {}
-        self._predefined_labels: list[str] = list(labels or [])
         self._label_history: list[str] = []
 
         self.edit = self._build_label_edit(placeholder=text, has_flags=bool(flags))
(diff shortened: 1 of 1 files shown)


wkentaro:

Finalized: mergeable on `main`, CI green all platforms. Confirmed the field was genuinely write-only: the `__init__` assignment was always overwritten by `set_predefined_labels` before any read, and the in-method read was redundant since `labels + self._label_history` already produces a fresh list (green ty/tests confirm no other readers). No behavior change. Undrafted and recommending merge.

PR #2174, opened 2026-06-13
fix: repaint the canvas immediately after adding or removing a polygon point

### LINKED ISSUE #890: [BUG] Missing `self.update()` function in the end of `canvas.addPointToEdge` and `canvas.removeSelectedPoint`
**Describe the bug**
When using the shortcut of `add_point_to_line` (Ctrl+Shift+P), the polygon not update immediately until have mousemove event.

**Expected behavior**
Fixed by adding `self.update()` in the end of both functions:
```python
# labelme\widgets\canvas.py:
# line: 304
    def addPointToEdge(self):
        shape = self.prevhShape
        index = self.prevhEdge
        point = self.prevMovePoint
        if shape is None or index is None or point is None:
            return
        shape.insertPoint(index, point)
        shape.highlightVertex(index, shape.MOVE_VERTEX)
        self.hShape = shape
        self.hVertex = index
        self.hEdge = None
        self.movingShape = True

        +++++++++++++++++
        + self.update() +
        +++++++++++++++++

    def removeSelectedPoint(self):
        shape = self.prevhShape
        point = self.prevMovePoint
        if shape is None or point is None:
            return
        index = shape.nearestVertex(point, self.epsilon)
        shape.removePoint(index)
        # shape.highlightVertex(index, shape.MOVE_VERTEX)
        self.hShape = shape
        self.hVertex = None
        self.hEdge = None
        self.movingShape = True  # Save changes

        +++++++++++++++++
        + self.update() +
        +++++++++++++++++
```

wkentaro:
`add_point_to_edge` and `remove_selected_point` mutated the polygon geometry but never called `self.update()`, so the edit was invisible until the next mouse-move event happened to repaint the canvas - the behavior reported in #890 (point added via the shortcut/action doesn't show until you move the mouse).

Call `self.update()` at the end of both methods so they repaint themselves, matching every other geometry-mutating method on `Canvas`. This also makes the explicit `canvas.update()` in `MainWindow.remove_selected_point` redundant (Qt defers the paint to after the whole synchronous handler, so the post-delete state still renders), so it's dropped.

## Test plan
- [x] unit tests added (`tests/unit/widgets/canvas_test.py`): adding/removing a point calls `update()` exactly once (spied via `Mock`) and mutates the point count; both verified to fail without the fix
- [x] `uv run pytest tests/unit` (241 passed) and `tests/e2e/canvas_interaction_test.py tests/e2e/oriented_rectangle_test.py` (28 passed)

Closes #890

### DIFF
--- a/labelme/app.py
+++ b/labelme/app.py
@@ -2537,7 +2537,6 @@ class MainWindow(QtWidgets.QMainWindow):
 
     def remove_selected_point(self) -> None:
         self._canvas_widgets.canvas.remove_selected_point()
-        self._canvas_widgets.canvas.update()
         if (
             self._canvas_widgets.canvas.hovered_shape
             and len(self._canvas_widgets.canvas.hovered_shape.points) == 0
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -890,6 +890,8 @@ class Canvas(QtWidgets.QWidget):
         self._hovered_vertex = index
         self._hovered_edge = None
         self._is_moving_shape = True
+        # Repaint now; otherwise the edit is invisible until the next mouse move.
+        self.update()
 
     def remove_selected_point(self) -> None:
         shape = self._last_hovered_shape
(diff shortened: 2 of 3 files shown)


wkentaro:

Rebased onto latest main (clean); `make lint` and the canvas unit tests (66) pass, CI green. Repaints the canvas immediately after add/remove polygon point so the edit is visible without a mouse move (closes #890), and drops the now-redundant explicit `canvas.update()` in MainWindow. Undrafted and recommending merge.

Note: this and #2175 both touch `remove_selected_point`; whichever merges second may need a trivial rebase.

PR #2175, opened 2026-06-13
fix: deselect the vertex after removing a polygon point

### LINKED ISSUE #968: [BUG]
**Describe the bug**
When deleting a point by pressing shift and clicking on the point, the adjacent point gets auto-selected. Moving the mouse even a little causes the adjacent point to snap to the cursor position. This effectively behaves as if the adjacent point was deleted and not the one that was clicked on.

**To Reproduce**
1. Create a polygon'
2. Press and hold the shift key
3. Click on a point to delete it. The point will be deleted. Note how the adjacent point gets auto-selected.
4. While keeping the shift key held down, move the mouse a little bit. Note how the adjacent point was moved to the current position.

**Expected behavior**
After deleting a point, all points should remain deselected and not snap to the mouse when it is moved.

**Desktop (please complete the following information):**
- OS: Windows 10 64-bit
- Labelme Version: 4.6.0

wkentaro:
`remove_selected_point` cleared `_last_hovered_vertex` but left `_hovered_vertex` pointing at the just-removed index. Because `_is_vertex_selected()` reads `_hovered_vertex` and the left-button drag path (`_continue_left_button_drag` -> `_drag_hovered_vertex`) drags `vertex_index=self._hovered_vertex`, a drag continued right after deleting a point moved the *adjacent* point (indices shifted by the removal) instead of leaving everything deselected - the behavior reported in #968.

Clear `_hovered_vertex` in `remove_selected_point` so the vertex is fully deselected after a removal.

Note: this touches the same method as #2174 (which adds `self.update()` there); whichever merges second will need a trivial one-line rebase.

## Test plan
- [x] unit test added (`tests/unit/widgets/canvas_test.py`): after `remove_selected_point`, the point is gone and `_is_vertex_selected()` is False - verified to fail without the fix
- [x] `uv run pytest tests/unit/widgets/canvas_test.py` (65 passed) and `tests/e2e/canvas_interaction_test.py` (25 passed)

Closes #968

### DIFF
--- a/labelme/widgets/canvas.py
+++ b/labelme/widgets/canvas.py
@@ -902,6 +902,8 @@ class Canvas(QtWidgets.QWidget):
         self._clear_highlight_state()
         self.hovered_shape = shape
         self._last_hovered_vertex = None
+        # Deselect the vertex; its index now aliases the adjacent point (#968).
+        self._hovered_vertex = None
         self._is_moving_shape = True  # Save changes
         # Repaint now; otherwise the edit is invisible until the next mouse move.
         self.update()
--- a/tests/unit/widgets/canvas_test.py
+++ b/tests/unit/widgets/canvas_test.py
@@ -656,3 +656,18 @@ def test_remove_selected_point_repaints(
 
     assert len(shape.points) == n_before - 1
     update.assert_called_once()  # repaint now, not only on the next mouse move (#890)
+
+
+@pytest.mark.gui
+def test_remove_selected_point_deselects_vertex(canvas: Canvas) -> None:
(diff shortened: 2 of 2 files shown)


wkentaro:

Rebased onto latest main (clean); `make lint` and the canvas unit tests (65) pass, CI green. Clears `_hovered_vertex` after removing a polygon point so a continued drag no longer moves the now-adjacent point (closes #968). Undrafted and recommending merge.

Note: this and #2174 both touch `remove_selected_point`; whichever merges second may need a trivial rebase.

PR #2178, opened 2026-06-14
fix(test): use non-deprecated QMouseEvent constructor in canvas_test

### LINKED ISSUE #2177: Use the non-deprecated QMouseEvent constructor in canvas_test.py

## What to build

`tests/unit/widgets/canvas_test.py` (lines ~301 and ~338) constructs `QMouseEvent` with the Qt 6-deprecated 5-arg overload, producing DeprecationWarnings during the test run. Switch to the supported 6-arg form by adding the `globalPos` argument:

```python
QMouseEvent(type, localPos, globalPos, button, buttons, modifiers)
```

The event is fed directly to `canvas._extend_current_shape`, which does not read the global position, so any `QPointF` works (e.g. reuse `localPos`). Pattern reference: `drag_canvas` in `tests/e2e/conftest.py` already uses the non-deprecated form.

## Acceptance criteria

- [ ] No `QMouseEvent` DeprecationWarning when running `pytest tests/unit/widgets/canvas_test.py`
- [ ] Both affected tests still pass

## Context

Surfaced during the PySide6 migration (#2144 / PR #2158).

wkentaro:
Switch the two `QMouseEvent` constructions in `canvas_test.py` from the Qt6-deprecated 5-arg overload to the supported 6-arg form by adding the `globalPos` argument. The event is fed to `canvas._extend_current_shape`, which does not read the global position, so `localPos` is reused. Matches the pattern already used by `drag_canvas` in `tests/e2e/conftest.py`.

Closes #2177.

## Test plan
- [x] `uv run pytest tests/unit/widgets/canvas_test.py -W error::DeprecationWarning` (64 passed, no QMouseEvent DeprecationWarning)

### DIFF
--- a/tests/unit/widgets/canvas_test.py
+++ b/tests/unit/widgets/canvas_test.py
@@ -301,6 +301,7 @@ def test_extend_after_mode_switch_finalizes_at_last_cursor(
     event = QtGui.QMouseEvent(
         QtCore.QEvent.Type.MouseButtonPress,
         QPointF(50, 30),
+        QPointF(50, 30),
         Qt.MouseButton.LeftButton,
         Qt.MouseButton.LeftButton,
         Qt.KeyboardModifier.NoModifier,
@@ -337,6 +338,7 @@ def test_extend_after_mode_switch_grows_partial_at_last_cursor(
 
     event = QtGui.QMouseEvent(
         QtCore.QEvent.Type.MouseButtonPress,
+        QPointF(50, 30),
         QPointF(50, 30),
         Qt.MouseButton.LeftButton,
         Qt.MouseButton.LeftButton,


wkentaro:

**Verdict: recommend-merge**

Targeted test-only fix. Adds the `globalPos` argument to the two `QMouseEvent` constructions in `canvas_test.py`, migrating from the Qt6-deprecated 5-arg overload to the supported 6-arg form.

- Argument order is correct (`globalPos` inserted as the 3rd positional arg).
- Reusing `localPos` for `globalPos` is sound here since `_extend_current_shape` ignores global position. This intentionally differs from `drag_canvas` in `tests/e2e/conftest.py`, which computes the real `mapToGlobal(end)` because it pumps a fuller event simulation; these unit tests invoke the handler directly, so an accurate global mapping buys nothing.
- Two added lines, no deletions, no behavioral change to assertions.

PR #2196, opened 2026-06-16
fix: warn when text-to-annotation model lacks mask output

wkentaro:
AI Text-to-Annotation silently produced nothing when YOLO-World was paired with a mask-based output format (Polygon/Mask). YOLO-World is a detector that returns bounding boxes only, so `_shape_from_detection` dropped every detection that needed a mask. This warns the user and points them to SAM3 (smart) or Rectangle output instead of failing silently.

The warning only fires when the model actually returned detections (`len(boxes) > 0`). A mask-capable model that matches nothing returns `masks is None` too, so without this guard a SAM3 query with zero matches would have shown a spurious "model can't create masks" warning.

## Test plan
- [x] YOLO-World + Polygon (with detections) -> shows "Mask Output Unavailable" warning instead of doing nothing
- [x] SAM3 + Polygon with zero matches -> no spurious warning
- [x] YOLO-World + Rectangle, and SAM3 + Polygon -> annotations still created
- [x] `uv run ruff check` and `uv run ty check labelme/app.py`


wkentaro:

**Fixed the failing `lint` check.** The two new `tr()` strings (`"Mask Output Unavailable"` and the body message) left `labelme/translate/*.ts` out of date, so `make lint` failed on both `git diff --exit-code labelme/translate` and the `type="unfinished"` gate.

Pushed `776cce79 chore(translate): regenerate translations`:
- Ran `tools/update_translate.py` and filled the two new strings for all 20 locales (kept `%s`, `'SAM3 (smart)'`, `'Rectangle'`, and the `AI Text-to-Annotation` feature name as literals, matching how existing tokens like `'AI-Points'` are left untranslated).
- Regenerated the `.qm` files; verified `make lint` is green locally and `update_translate.py` is idempotent (no churn on a second run).

A medium-effort code review of the guard found it **correct**. Two optional, non-blocking polish notes for your call (not changed, since they touch user-facing wording):
1. The body shows the raw lowercase `shape_type` (e.g. `'polygon'`) while the output-format combo displays `"Polygon"`.
2. If the active model is already `SAM3 (smart)` but returns box-only detections, the "switch to SAM3" suggestion reads oddly (a pre-existing limitation of the detection API, not introduced here).

Awaiting CI; will set the verdict once checks are green.

wkentaro:

**Verdict: `recommend-merge`.** All 8 checks are green on `776cce79` (`lint` now passes, all `test` matrix jobs pass). The code review found the guard correct, and the only changes since the last red build were the regenerated translations. Branch is up to date with `main` (fast-forward). Ready for your review and merge.

PR #2197, opened 2026-06-16
fix: default delete confirmation to cancel with verb buttons

wkentaro:
The delete confirmation defaulted to its destructive button, so pressing
Enter (or any accidental default activation) permanently deleted shapes with
no recovery. Make Cancel the default instead, and label the buttons with verbs
(Delete / Cancel) rather than Yes / No, following macOS alert conventions. A
single `_confirm_deletion` helper now backs both `delete_file` and
`delete_selected_shapes`.

## Test plan
- [x] `ruff check` and `ty check` clean on the touched files
- [x] New `tests/e2e/delete_confirmation_test.py`: the default button carries
      the reject role (Enter cancels), clicking Delete returns `True`, and
      clicking Cancel returns `False`
- [x] Existing delete e2e tests updated to the new confirmation seam
- [x] Full e2e suite green (`pytest tests/e2e`, 150 passed)


wkentaro:

**Verdict: `recommend-merge`.** Branch was rebased onto current `main` (`0ae9ef3a`); all 8 checks are green (`lint` now passes after the new `Delete`/`Cancel` button strings were translated for all 20 locales and folded into the feature commit). mergeState is `CLEAN`.

Code review found the change **clean**:
- `_confirm_deletion` correctly wires the Delete (DestructiveRole) / Cancel (RejectRole) buttons, defaults to Cancel, and returns via `clickedButton() is delete_button` (the Escape/close `None` case safely returns False).
- Behavior parity is exact in both `delete_file` and `delete_selected_shapes`; the only semantic change is the intended default-to-Cancel.
- The 3 new tests plus the migrated e2e tests exercise the helper directly. I ran the delete-confirmation e2e suite locally: 11 passed.

PR #2200, opened 2026-06-16
test: cover masks_to_bboxes happy path, guards, and empty mask

wkentaro:
`labelme.utils.masks_to_bboxes` is part of the public utils API but had no test
coverage. These tests pin its `(y1, x1, y2, x2)` return order, both input
validation guards (`ndim` and `dtype`), and the all-False-mask path.

The all-False case currently raises a numpy-internal `ValueError` (no foreground
pixels, so no bbox is defined); the test characterizes that behavior without
deciding whether an explicit guard or a sentinel return is preferable. That
contract is left for a follow-up.

## Test plan
- [x] `uv run pytest tests/unit/utils/shape_test.py` (11 passed)
- [x] `uv run ruff check` / `ruff format --check` clean

### DIFF
--- a/tests/unit/utils/shape_test.py
+++ b/tests/unit/utils/shape_test.py
@@ -111,3 +111,36 @@ def test_shape_to_mask_rectangle_reversed_coords() -> None:
     assert np.array_equal(mask_tl_br, mask_tr_bl)
     assert np.array_equal(mask_tl_br, mask_bl_tr)
     assert mask_tl_br.sum() > 0
+
+
+def test_masks_to_bboxes_returns_yxyx_bboxes() -> None:
+    masks = np.zeros((2, 10, 10), dtype=bool)
+    masks[0, 2:5, 3:7] = True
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: `recommend-merge`.** Test-only change, all 8 checks green, mergeState `CLEAN`.

Reviewed each test against `labelme/utils/shape.py::masks_to_bboxes`:
- `returns_yxyx_bboxes`: matches the impl (`where.min(0)`, `where.max(0) + 1` → `(y1, x1, y2, x2)`), incl. `float32` dtype and `(N, 4)` shape.
- `raises_for_wrong_ndim` / `raises_for_non_bool_dtype`: the `match=` patterns line up with the guard messages.
- `raises_for_all_false_mask`: correctly pins numpy's empty-reduction `ValueError` without coupling to its message.

Good coverage of the happy path and all guards. Ready for review and merge.

PR #2201, opened 2026-06-16
refactor: derive mask-required shape types from a single source

### LINKED ISSUE #2198: Derive mask-required shape types from a single source instead of duplicating them

Follow-up from #2196.

The set of AI Text-to-Annotation output formats that **require a mask** from the model is currently encoded in two independent places:

- In the AI-prompt submission path (`_submit_ai_prompt`), a local `MASK_REQUIRED_SHAPE_TYPES = ("polygon", "mask")` gates the "Mask Output Unavailable" warning shown when a bbox-only model is paired with a mask-based output.
- In the shape-building path (`_shape_from_detection` / `shapes_from_detections`), each output format implicitly decides this by returning `None` when `detection.mask is None`. Today `polygon` and `mask` hard-require a mask, while `circle` and `oriented_rectangle` gracefully fall back to the bounding box.

These two encodings are not synchronized. If a new mask-only output format is added to the shape builders but not to the warning's tuple, the warning silently won't fire and the format will drop every detection with no user feedback (the exact failure #2196 set out to prevent).

This is a small maintainability/DRY cleanup, not a user-visible defect today.

wkentaro:
The set of AI Text-to-Annotation output formats that require a mask was
encoded in two independent places: a local tuple in `_submit_ai_prompt`
gated the "Mask Output Unavailable" warning, while the shape builders
independently encoded the same set by returning `None` when a detection has
no mask. A new mask-only format added to the builders but not the warning
tuple would silently drop every detection with no user feedback (the failure
#2196 set out to prevent).

This derives `MASK_REQUIRED_SHAPE_TYPES` from `_shape_from_detection` itself,
exports it from the `_automation` package, and consumes it in the warning
path. The probe mirrors the runtime warning condition (a box but no mask), so
the set cannot drift from builder behavior.

## Test plan
- [x] Added `test_mask_required_shape_types_is_polygon_and_mask` pinning the
      derived set to `{"polygon", "mask"}`
- [x] `tests/e2e/ai_text_to_annotation_test.py` (exercises the warning path)
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run ty check`
- [x] Full suite: `uv run pytest tests/` (434 passed)

Closes #2198

### DIFF
--- a/labelme/_automation/__init__.py
+++ b/labelme/_automation/__init__.py
@@ -1,6 +1,7 @@
 from ._ai_assist import AiAssistSession
 from ._geometry import shape_to_xyxy_bbox
 from ._osam_session import OsamSession
+from ._shape_builders import MASK_REQUIRED_SHAPE_TYPES
 from ._shape_builders import Detection
 from ._shape_builders import shapes_from_detections
 from ._suppression import suppress_detections_greedy
--- a/labelme/_automation/_shape_builders.py
+++ b/labelme/_automation/_shape_builders.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from typing import Final
+from typing import get_args
 
 import numpy as np
 from numpy.typing import ArrayLike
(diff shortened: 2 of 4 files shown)


wkentaro:

**Verdict: `recommend-merge`.** All 8 checks green, mergeState `CLEAN`.

Reviewed the refactor as behavior-preserving:
- `MASK_REQUIRED_SHAPE_TYPES` is now derived from `_shape_from_detection` (the single source of truth) by probing which output formats drop a maskless detection, replacing the hardcoded `("polygon", "mask")` tuple in `app.py` that could drift.
- Confirmed `_shape_from_detection` returns `None` exactly for `polygon`/`mask` when `mask is None`, while `rectangle` builds from the bbox, so the import-time probe yields `{"polygon", "mask"}` (pinned by the new test). The probe uses a degenerate `bbox=(0, 0, 1, 1)` and does not crash at import (CI import succeeds).
- Constant is module-level in `_shape_builders.py` and exported via `_automation/__init__.py`; justified since it is now shared with `app.py`.

Closes the drift risk noted when the original guard landed in #2196. Ready for review and merge.

PR #2212, opened 2026-06-16
test: cover apply_exif_orientation transforms and no-exif guards

wkentaro:
`labelme.utils.image.apply_exif_orientation` sits in the production image-load
path (`_label_file.py`) but had no test coverage. This parametrizes all eight
EXIF orientations, asserting the output size and the transformed marker position
(tracked via the brightest-red pixel, with JPEG quality pinned so the marker
stays dominant), plus both no-EXIF guard branches (a non-JPEG `fromarray` image
and a plain JPEG with no EXIF header).

## Test plan
- [x] `uv run pytest tests/unit/utils/image_test.py` (13 passed)

### DIFF
--- a/tests/unit/utils/image_test.py
+++ b/tests/unit/utils/image_test.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import io
+from typing import Final
 
 import numpy as np
 import PIL.Image
+import pytest
 
 from labelme.utils import image as image_module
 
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: `recommend-merge`.** Test-only change, all 8 checks green, mergeState `CLEAN`.

Strong coverage of `apply_exif_orientation`: parametrized over all 8 EXIF Orientation values, asserting both the output size and the migrated position of an asymmetric red marker (so an x/y swap or wrong flip is caught), plus two pass-through guards (a non-JPEG image without EXIF, and a JPEG without an Orientation tag returning the same object identity). The JPEG round-trip with `quality=95, subsampling=0` keeps the marker the brightest pixel, which the assertions rely on. Ready for review and merge.

PR #2219, opened 2026-06-17
test: cover compute_polygon_from_mask

wkentaro:
`labelme._automation._geometry.compute_polygon_from_mask` (mask -> polygon via
skimage contouring) had no test coverage. This pins three behaviors: an empty
mask returns an empty `(0, 2)` array; a solid rectangle's polygon extent is
correct and xy-ordered (the max assertion catches a yx/xy swap); and with two
blobs the longer-perimeter contour is selected. Values were verified against the
implementation.

These characterize current behavior, including the function's padded-space
coordinate offset and the empty-vs-non-empty dtype split (the annotation says
float32 but non-empty masks return float64); both are left as-is here.

## Test plan
- [x] `uv run pytest tests/unit/_automation/_geometry_test.py` (19 passed)

### DIFF
--- a/tests/unit/_automation/_geometry_test.py
+++ b/tests/unit/_automation/_geometry_test.py
@@ -8,6 +8,7 @@ from numpy.typing import NDArray
 
 from labelme._automation._geometry import compute_circle_from_mask
 from labelme._automation._geometry import compute_oriented_rectangle_from_mask
+from labelme._automation._geometry import compute_polygon_from_mask
 from labelme._automation._geometry import shape_to_xyxy_bbox
 from labelme._shape import Shape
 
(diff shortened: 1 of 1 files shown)


wkentaro:

Rebased onto current `main` and re-ran CI on the updated matrix (Python 3.11/3.14; `main` had dropped 3.10, so the prior green run was on a stale matrix). All 8 checks green. `/review-fix` (5 reviewers) found nothing to change in the test diff; `ruff`/`ty`/`pytest` clean locally.

Test-only PR characterizing `compute_polygon_from_mask` (empty mask, xy-ordering, largest-contour selection). The tests intentionally pin current behavior including the padded-space coordinate offset.

follow-ups: #2224 (the ~1px coordinate offset from the uncorrected `pad_width`, surfaced during review; does not block this PR).

PR #2221, opened 2026-06-17
test: cover oriented_rectangle_arrow_points

wkentaro:
`labelme._shape.oriented_rectangle_arrow_points` computes the direction-arrow
overlay for an oriented rectangle (rotate a fixed template by the rectangle's
angle, then translate to its center) but had no coverage. This pins three cases:
axis-aligned (identity), a 90-degree rotation (tip moves +x -> +y), and a
non-cardinal direction (8, 6) chosen so an `arctan2` argument swap would change
the result. Pure numpy.

## Test plan
- [x] `uv run pytest tests/unit/_shape_test.py` (47 passed)

### DIFF
--- a/tests/unit/_shape_test.py
+++ b/tests/unit/_shape_test.py
@@ -10,14 +10,16 @@ from labelme._shape import Shape
 from labelme._shape import ShapeType
 
 
-def _make_axis_aligned_oriented_rectangle() -> Shape:
+def _make_oriented_rectangle(points: list[tuple[float, float]]) -> Shape:
     return Shape(
         shape_type="oriented_rectangle",
-        points=np.array(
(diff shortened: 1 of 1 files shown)


wkentaro:

Rebased onto current `main`; CI green (8/8) on the updated Python 3.11/3.14 matrix. `/review-fix` (5 reviewers) found the diff correct; applied one cleanup: extracted the repeated `Shape(shape_type="oriented_rectangle", ..., closed=True)` construction into a `_make_oriented_rectangle(points)` helper (the existing axis-aligned factory now delegates to it), removing the copy-pasted setup across the three tests. `ruff`/`ty`/`pytest` clean.

Test-only PR: covers `oriented_rectangle_arrow_points` for axis-aligned, 90-degree rotated, and non-cardinal directions (the non-square/non-cardinal cases guard against a yx/xy swap and arctan2 argument-order bugs).

PR #2223, opened 2026-06-17
feat: allow annotation points outside the image boundary

### LINKED ISSUE #2116: Allow placing annotation points outside the image boundary for partially visible objects

## Problem

It is currently impossible to annotate objects that are only partially visible in the image (partially outside the field of view). Whenever a point would land outside the image, the canvas snaps it back to the image border:

- While drawing a new shape, the cursor position is clamped to the image edge (`labelme/widgets/canvas.py` — `_compute_intersection_edges_image`).
- While editing, `_bounded_move_vertex` clamps the dragged vertex to the image edge.
- Dragging a whole shape is blocked when the cursor leaves the image.

As a result, a rectangle/polygon for a partially visible object always ends at the image border, even when the annotator knows the object's true extent (e.g., the physical size of the object is known, or annotations are converted from existing detection results).

## Proposal

Allow annotation points to go outside the image boundary, so shapes can represent the true extent of partially visible objects:

- Add a config option (e.g., `allow_out_of_bounds_points: true`) that disables the snap-to-border behavior for drawing, vertex editing, and shape dragging. Default stays as today (clamped) to preserve current behavior.
- The JSON format already stores points as plain floats, so negative or larger-than-image coordinates serialize fine. Loading such files should keep the points where they are instead of clamping them on edit.
- Rasterization utilities (e.g., `shape_to_mask`) naturally clip to the image canvas, so downstream mask generation keeps working.

## Use case

Creating ground truth for object detection where items at the edge of the camera's field of view are cut off. The visible part alone is not the correct ground truth box; the annotator wants to place corners at the object's actual extent, slightly outside the image.

wkentaro:
Closes #2116.

Annotating partially-visible objects (cut off by the camera's field of view) was impossible: the canvas snapped every point back to the image edge, so a shape could never represent the object's true extent. This adds an opt-in `canvas.allow_out_of_bounds_points` setting (default `false`, preserving today's clamping) that relaxes the snap.

image

## Summary
- New setting `canvas.allow_out_of_bounds_points`, threaded through `default_config.yaml` → `app.py` → `Canvas`, exposed in the Settings dialog and applied live.
- When enabled, points may extend past the image while drawing, editing a vertex, dragging a whole shape, reprojecting an oriented rectangle, and the crosshair tracks the cursor across the full viewport.
- Default behavior, the on-disk format, and mask/COCO/VOC export are unchanged: coordinates already serialize as plain floats, and rasterization still clips to the image.
- Settings-dialog helper text now renders under the label (title+subtitle layout) rather than under the control.
- Rationale recorded in `docs/adr/0004-out-of-bounds-points.md` (the canvas no longer guarantees in-image points when the setting is on; do not re-add a blanket clamp).

## Test plan
- [x] Unit tests added (`tests/unit/widgets/canvas_test.py`): vertex edit, whole-shape drag, oriented-rectangle reprojection, and crosshair, each covering default-clamp vs flag-on.
- [x] `uv run ruff check` + `ruff format --check` + `uv run ty check` clean.
- [x] Full suite green: `475 passed` (incl. canvas paint snapshot and settings/config e2e).
- [x] Manual GUI validation (the acceptance gate, can't be automated): draw/drag/edit past each image edge with the setting on, then save+reload to confirm coordinates persist.


wkentaro:

Finalized this branch:

- **Fixed the failing lint** (the only red check): the PR added two new
  Settings strings but the translations were stale. Regenerated all 20
  `.ts`/`.qm` locales and filled in both strings in every language (lint
  rejects `type="unfinished"` entries).
- **Re-rebased onto `main`** after the cryptography bump (#2222) merged, which
  cleared a spurious `uv.lock` diff.
- **Ran the review loop and applied the meaningful findings:**
  - extracted a `_should_constrain_to_pixmap()` predicate, replacing four
    duplicated `not self._allow_out_of_bounds_points and self.is_out_of_pixmap(...)`
    guards plus the equivalent open-coded check in `_should_draw_crosshair`, so
    the bounds policy now lives in one place;
  - made `allow_out_of_bounds` a required parameter on
    `_reproject_oriented_rectangle_corners` (dropping the `= False` default that
    would silently clamp a future caller that forgot the flag);
  - tightened `test_bounded_move_vertex_clamps_to_image_by_default` to assert the
    exact clamped point `(75, 50)` instead of a loose `<= bounds` bound;
  - removed an em dash from ADR-0004.
- **Verified behavior:** with the flag off a dragged vertex clamps to `(75, 50)`
  on-image; with the flag on it keeps `(150, 80)` outside the boundary.

All checks green (lint + tests on ubuntu/macOS/windows × 3.11/3.14). Commit
sequence is clean: docs(ADR) → feat → test → chore(translate). Ready for your
review and merge.

PR #2236, opened 2026-06-18
Modernize LabelDialog for Qt 6

### LINKED ISSUE #2147: Modernize LabelDialog
## What to build

LabelDialog mixes several concerns (text completion, flags, color handling) and carries long-stale TODO comments. Restructure it to current code style with clear separation of those concerns, resolving or removing the stale TODOs.

## Acceptance criteria

- [ ] Popup flow identical: prefilled text, completion list, move-to-cursor, validation, OK/cancel semantics
- [ ] Per-label flags continue to work
- [ ] Test suite passes

## Blocked by

- #2144 (PySide6 migration)

wkentaro:
Reimplements the `LabelDialog` label-entry widget for Qt 6 / PySide6, and reworks its tests to behavior-level (black-box) characterization.

- Characterization tests pin observable behavior through the public API, then the widget is reimplemented to satisfy them.
- Full test suite green.

Closes #2147

### DIFF
--- a/labelme/widgets/label_dialog.py
+++ b/labelme/widgets/label_dialog.py
@@ -260,7 +260,7 @@ class LabelDialog(QtWidgets.QDialog):
             self._update_flags(self.edit.text())
 
         matches = self.label_list.findItems(
-            self.edit.text(), QtCore.Qt.MatchFlag.MatchExactly
+            self.edit.text(), QtCore.Qt.MatchFlag.MatchFixedString
         )
         if matches:
             self.label_list.setCurrentItem(matches[0])
--- a/labelme/widgets/label_dialog.py
+++ b/labelme/widgets/label_dialog.py
@@ -1,34 +1,44 @@
 from __future__ import annotations
 
 import re
-from typing import cast
+from typing import Final
 
-from loguru import logger
 from PySide6 import QtCore
 from PySide6 import QtGui
 from PySide6 import QtWidgets
 
-import labelme.utils
+from labelme.utils import label_validator
 
-# TODO(unknown):
-# - Calculate optimal position so as not to go out of screen area.
+_PLACEHOLDER_TEXT: Final[str] = "Enter object label"
+_GROUP_ID_PLACEHOLDER: Final[str] = "Group ID"
+_DESCRIPTION_PLACEHOLDER: Final[str] = "Description"
 
 
 class LabelQLineEdit(QtWidgets.QLineEdit):
+    """QLineEdit that forwards Up/Down key events to a paired list widget."""
+
+    def __init__(self, parent: QtWidgets.QWidget | None = None) -> None:
+        super().__init__(parent)
+        self.list_widget: QtWidgets.QListWidget | None = None
+
     def set_list_widget(self, list_widget: QtWidgets.QListWidget) -> None:
         self.list_widget = list_widget
 
-    def keyPressEvent(self, a0: QtGui.QKeyEvent) -> None:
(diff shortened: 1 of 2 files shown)


wkentaro:
> *This was generated by AI.*

Updated the branch (force-pushed) with two follow-up fixes on top of the reimplementation:

- **fix: disable flag checkboxes when `flags_disabled` is set** — `popup()` accepted `flags_disabled` but never acted on it, so flag checkboxes stayed editable even when the caller asked to lock them (e.g. editing multiple shapes whose flags differ). The disable is now applied as the final step in `popup()`, after the last `setText()` that could rebuild the checkboxes, so it sticks. Checkboxes are also orphaned from their parent before `deleteLater()` so the disabled state is observable. Covered by updated/new characterization tests.
- **fix: tighten spacing between flag checkboxes** — the flags layout used the default `QVBoxLayout` spacing, leaving a visible gap between checkboxes; set to 0 so they sit flush.

Full suite green; ruff and ty clean.

wkentaro:

**Verdict: `recommend-merge`** — sound rewrite, CI green (8/8), public API fully preserved.

**Finalize checks**
- **Rebase:** skipped — `main` gained 0 commits since the merge-base and merge state is `CLEAN`, so the green CI already reflects the merged result.
- **Review (`/code-review`, medium):** no correctness findings after verification (details below). Skipped the full `/review-fix` loop.
- **History:** already clean and logically ordered (`test: characterize` → `feat: reimplement` → two documented `fix:` commits), each independently valid. No reshaping needed.
- **Compatibility:** traced every call site (`app.py` constructor + both `popup()` calls, `add_label_history`, `set_predefined_labels`, and direct `.edit`/`.label_list`/`.edit_group_id`/`.edit_description` access, plus e2e tests). All satisfied by the new public API; return tuple shape unchanged.

**Refuted finding (worth recording).** A plausible-looking concern that `popup(flags=...)` would be wiped when the label matches a predefined item — via `setCurrentItem` → `_on_label_selected` → `setText` → `_update_flags` — does **not** occur. `setText` only emits `textChanged` when the value actually changes, and the matched item always equals the current edit text exactly, so `_update_flags` never re-fires there. Verified at runtime across empty-spec, extra-key, in-spec, and changed-prior-text cases: passed flags are preserved in all four.

**Two minor behavioral differences vs. the old impl (non-blocking, for your call as author).** Neither is pinned by the characterization tests:
- Popup highlight matching changed from `MatchFixedString` (case-insensitive) to `MatchExactly` (case-sensitive): editing a shape whose label differs only in case from a predefined entry (e.g. `cat` vs `Cat`) no longer pre-selects the list item. Arguably more correct, since labelme treats labels case-sensitively.
- `popup(text=None)` no longer calls `selectAll()` on the edit field (old impl always did). On the new-shape path (`app.py:1829`) the leftover label is retained but not pre-selected, so typing appends rather than overtypes. A mild papercut if you want to restore it; otherwise harmless.

CI: all 8 checks green. Recommending merge.

wkentaro:

Addressed the two behavioral differences flagged in the review by restoring the old impl behavior, each with a characterization test that fails without the fix:

- **fix: highlight predefined label case-insensitively on popup** — reverted the popup highlight match from `MatchExactly` back to `MatchFixedString`, so editing a shape whose label differs only in case from a predefined entry (`cat` vs `Cat`) pre-selects the list item again.
- **fix: select existing edit text when popup opens without explicit text** — `popup(text=None)` now calls `selectAll()` on the edit field as the old impl did, so on the new-shape path the retained label is pre-selected and typing overtypes rather than appends.

Full unit suite green locally (466 passed); ruff and ty clean. Removed the stale `recommend-merge` label; will re-verdict once CI confirms green on the new commits.

PR #2238, opened 2026-06-19
fix: guard AI inference so model errors surface instead of crashing

### LINKED ISSUE #2202: Guard AI inference run so model errors don't silently crash the app
## What to build

The AI-assisted annotation inference run can throw (model/runtime errors, frozen-exe DLL issues, malformed model output) and the exception is currently unhandled, so the app dies silently — the long-standing "click Create AI-polygon, window flashes back, no error message" symptom. Make a failed inference a recoverable, visible error instead of a crash: catch exceptions around the inference call, log the traceback, and surface a non-fatal message (dialog or status bar) so the user can keep working. The mouse-move preview path must not repeatedly crash on hover.

## Acceptance criteria

- [ ] An exception raised during AI point/box inference does not terminate the application
- [ ] The user sees a clear, non-fatal error (dialog or status message) and the traceback is logged
- [ ] The hover/preview inference path is guarded and/or throttled so a failing model does not spam-crash
- [ ] Manual repro: pointing the AI tool at a broken/missing model surfaces an error instead of closing the app

## Related

Addresses the most-reported crash cluster: #1300, #1334, #1335, #1459, #1473. Do not close those automatically; link on merge.

## Blocked by

None - can start immediately

wkentaro:
Closes #2202

AI-assisted inference (AI-Points / AI-Box / Text-to-Annotation) could throw on
model load, frozen-exe DLL issues, or malformed model output, and the exception
was unhandled. The points preview path runs inference inside `paintEvent` on
every repaint, so a failure there killed the process silently: the long-standing
"click Create AI-Polygon, window flashes back, no error" symptom.

## Summary
- Guard the single inference choke point (`Canvas._shapes_from_ai_points`) and
  the text-prompt path: catch the exception, log the traceback, surface a
  non-fatal status-bar message, and return no shapes instead of crashing.
- Throttle the hover/preview path with a per-exception-type latch so a
  persistently failing model reports once rather than spamming the log and
  status bar on every repaint frame.
- Reset the latch whenever the inference context changes (successful inference,
  a commit click, loading a new image, or re-entering an AI mode) so each fresh
  attempt surfaces its own first failure.
- `_finalize` suppresses the misleading "produced no new annotation" notice when
  the empty result came from an error.

## Test plan
- [x] tests added: `tests/unit/widgets/canvas_test.py` (error is swallowed and
  emitted, dedup throttles repeats, latch resets on commit / image-load /
  AI-mode entry, `_finalize` suppresses the no-shapes notice on error)
- [x] headless smoke: a real exception from `propose_shapes` is caught, the app
  stays alive, and the status bar shows `AI inference failed: ...`
- [x] full suite: `uv run pytest tests/` (570 passed)
- [x] `make lint`


wkentaro:

Finalized in a `/process-prs` tick.

- **Divergence:** branch is fully current with `main` (merge-base = `main` tip, 0 commits behind, no file overlap). No rebase needed.
- **Code review:** ran a medium-effort review over the non-generated diff (`app.py`, `canvas.py`, `canvas_test.py`). No meaningful findings. The `except Exception` guard correctly spares `KeyboardInterrupt`/`SystemExit`; the per-exception-type latch resets on new pixmap, AI-mode entry, and at the top of `_finalize`; `_finalize`'s suppression of the "no new annotation" notice on error is correct; text inference reports via `_on_inference_failed` (no double-report). Tests cover dedup, recovery-after-success, suppression, commit-path re-report, and both reset paths.
- **Commits:** already a clean 2-commit sequence (fix + translation regen). No reshaping.
- **CI:** green on all platforms (lint, cla, test on ubuntu/macos/windows × py3.11/3.14).
- **mergeState:** CLEAN.

Recommend merge. Final merge decision remains the maintainer's.

wkentaro:
let's try again.

PR #2239, opened 2026-06-19
fix: align compute_polygon_from_mask coordinates with image space

### LINKED ISSUE #2224: compute_polygon_from_mask returns coordinates offset ~1px by uncorrected pad_width

While reviewing #2219 (which adds characterization tests for
`labelme._automation._geometry.compute_polygon_from_mask`), I noticed the
returned polygon coordinates appear to carry a systematic ~1px offset.

`compute_polygon_from_mask` pads the mask before contouring:

```python
contours = skimage.measure.find_contours(np.pad(mask, pad_width=1))
...
polygon = np.clip(polygon, (0, 0), (mask.shape[0] - 1, mask.shape[1] - 1))
...
return polygon[:, ::-1]  # yx -> xy
```

Two things stand out:

1. The `pad_width=1` shifts every contour coordinate by +1 in both axes, but
   the offset is never subtracted before returning.
2. The clip ceiling is `mask.shape - 1`, applied in the padded coordinate
   space.

Worked example: a `(5, 9)` mask with `mask[1:4, 1:8] = True` (filled cols 1-7,
rows 1-3). The half-pixel boundary of that region in original-image space is
`x in [0.5, 7.5]`, `y in [0.5, 3.5]`. The function returns `min = [1.5, 1.5]`,
`max = [8.0, 4.0]` (xy) instead, i.e. roughly a +1px diagonal shift on the
min corner with the max clamped by the `shape - 1` clip.

This affects polygons produced by AI-assisted / automation annotation. The
characterization tests in #2219 intentionally pin the *current* behavior
(the PR body documents the padded-space offset and leaves it as-is), so this
issue tracks whether the offset should be corrected.

Suggested investigation: subtract the pad width (and reconsider the clip
bound, e.g. clamp to `mask.shape` rather than `mask.shape - 1`) so returned
coordinates align with the original image, then update the #2219 tests to the
corrected expectations.

Originating PR: #2219

wkentaro:
Closes #2224

`compute_polygon_from_mask` pads the mask by 1px before `find_contours` (so a
region touching the image border still has a background ring to close its
contour against), but the resulting contour coordinates were never shifted back.
Every returned polygon therefore carried a systematic ~1px diagonal offset. The
clip ceiling compounded it: it used `mask.shape - 1` in that padded space,
pulling the far edge in by another pixel.

This affects polygons produced by AI-assisted / automation annotation
(`_shape_builders` consumes this helper).

## Summary
- Subtract the pad width after contouring to return coordinates to image space.
- Clamp to `mask.shape` (not `mask.shape - 1`): `np.clip` is inclusive and a
  half-pixel boundary can reach `shape - 0.5`, so the old ceiling pulled
  edge-touching contours inward by a pixel.
- Update the #2219 characterization tests (which pinned the offset behavior) to
  the corrected expectations, and add an edge-touching regression test.

For the issue's worked example (a `(5, 9)` mask with `mask[1:4, 1:8]`), the
function now returns `min=[0.5, 0.5]`, `max=[7.5, 3.5]` (xy) — the region's true
half-pixel boundary in image space — instead of the previous `[1.5, 1.5]` /
`[8.0, 4.0]`.

## Test plan
- [x] updated characterization tests + new edge-touching case in
  `tests/unit/_automation/_geometry_test.py`
- [x] empirical check: worked example -> `[0.5,0.5]..[7.5,3.5]`; fully-set mask
  keeps far-edge boundary `..[8.5,4.5]`
- [x] full suite: `uv run pytest tests/` (564 passed)
- [x] `make lint`

### DIFF
--- a/labelme/_automation/_geometry.py
+++ b/labelme/_automation/_geometry.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from typing import Final
 from typing import NamedTuple
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


wkentaro:

Finalized in a `/process-prs` tick.

- **Divergence:** branch fully current with `main` (merge-base = `main` tip, 0 commits behind, no file overlap). No rebase needed.
- **Code review:** medium-effort review of `_geometry.py` + test. No meaningful findings. The `-PAD` shift correctly undoes the `np.pad(mask, pad_width=1)` offset (every contour vertex was 1 row/col too large), and the clip ceiling change from `shape - 1` to `shape` is correct because `skimage` places contour vertices at half-pixel boundaries (`H-0.5`/`W-0.5`) for border-touching regions, which the old ceiling clamped destructively. Tests cover interior, far-border, and near-zero cases.
- **Commits:** single clean commit. No reshaping.
- **CI:** green on all platforms (lint, cla, test on ubuntu/macos/windows × py3.11/3.14).
- **mergeState:** CLEAN.

Recommend merge. Final merge decision remains the maintainer's.

Issue #2240, opened 2026-06-19
flags_disabled has no effect: read-only flag checkboxes stay editable in LabelDialog


wkentaro:
> *This was generated by AI.*

## Summary

`LabelDialog.popup(..., flags_disabled=True)` does not make the flag
checkboxes read-only. The flags remain editable even when the caller asks
for them to be locked. This is a long-standing latent bug: it is broken in
the current `main` implementation **and** was already broken before the Qt 6
reimplementation, so it is not a regression.

## Where it comes from

`app.py` passes `flags_disabled=not edit_flags` when opening the dialog, with
the intent that flag checkboxes should be non-editable when the user is not
allowed to edit flags (e.g. multi-shape / view contexts):

```python
text, flags, group_id, description = self._label_dialog.popup(
    text=...,
    flags=first_shape.flags if edit_flags else None,
    ...
    flags_disabled=not edit_flags,
)
```

But `popup()` never acts on `flags_disabled`, so the checkboxes stay enabled
and any toggles are returned by `_collect_flags()`, silently overwriting flags
that were meant to be read-only.

## Root cause

- **Old code (pre-rewrite):** `_apply_dialog_state` did disable the checkboxes,
  but then called `self.edit.setText(text)` afterwards. `setText` fires
  `textChanged`, which rebuilds the flag checkboxes from scratch (enabled),
  undoing the disable.
- **Current `main` (Qt 6 reimplementation):** `popup()` accepts the
  `flags_disabled` parameter but never reads it at all.

The existing characterization test
`test_popup_flags_disabled_checkboxes_remain_enabled` documents this no-op
behavior on purpose (it pins the old behavior), so a fix must update that test
to assert the intended behavior.

## Expected behavior

When `flags_disabled=True`, all flag checkboxes shown in the dialog should be
disabled (non-interactive), and toggling should be impossible, so read-only
flags cannot be mutated.

## Suggested fix

After building the flag checkboxes in `popup()` (both the `_show_popup_flags`
and `_update_flags` paths), disable them when `flags_disabled` is set, e.g.:

```python
if flags_disabled:
    for checkbox in self._flag_checkboxes():
        checkbox.setEnabled(False)
```

Care is needed so that a later `textChanged` rebuild (when the label text is
editable but flags are not) does not re-enable them; the disabled state may
need to be tracked and re-applied in `_update_flags`.

Update `test_popup_flags_disabled_checkboxes_remain_enabled` accordingly
[...]

PR #2243, opened 2026-06-20
fix: keep flag checkboxes disabled across textChanged rebuild

wkentaro:
`LabelDialog.popup(..., flags_disabled=True)` disabled the flag checkboxes once after building them, but editing the label text fires `textChanged`, which runs `_update_flags` and rebuilds the checkboxes enabled, re-enabling read-only flags. The original fix (#2240) covered the initial render but not this rebuild path, which the issue flagged as a residual gap.

This tracks the disabled state on the dialog and applies it in both flag-building paths (`_show_popup_flags` and `_update_flags`), so any rebuild preserves it. The now-redundant post-build loop in `popup()` is removed.

## Test plan
- [x] new `test_flags_disabled_survives_text_edit_rebuild` (edits text mid-popup, asserts checkboxes stay disabled); confirmed it fails without the `_update_flags` change
- [x] `uv run pytest tests/unit/widgets/label_dialog_characterize_test.py` (53 passed)
- [x] `uv run ruff check` / `uv run ty check` clean

### DIFF
--- a/labelme/widgets/label_dialog.py
+++ b/labelme/widgets/label_dialog.py
@@ -52,6 +52,7 @@ class LabelDialog(QtWidgets.QDialog):
         self._sort_labels = sort_labels
         self._flags_spec: dict[str, list[str]] = flags or {}
         self._label_history: list[str] = []
+        self._flags_disabled = False
 
         if fit_to_content is None:
             fit_to_content = {"row": False, "column": True}
@@ -212,6 +213,7 @@ class LabelDialog(QtWidgets.QDialog):
                     checkbox = QtWidgets.QCheckBox(key)
                     if key in current_states:
                         checkbox.setChecked(current_states[key])
+                    checkbox.setEnabled(not self._flags_disabled)
                     self._flags_layout.addWidget(checkbox)
 
     def add_label_history(self, label: str) -> None:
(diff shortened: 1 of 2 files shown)


wkentaro:

Reviewed the diff and read `labelme/widgets/label_dialog.py` on the branch.

- The fix is correct and minimal: `flags_disabled` is promoted to `self._flags_disabled` state and reapplied at construction time in every checkbox-build path (`_update_flags`, `_show_popup_flags`), replacing the old post-hoc `setEnabled(False)` loop that a later `textChanged` rebuild would silently drop. That rebuild-on-text-edit case is exactly #2240, and it's now covered by `test_flags_disabled_survives_text_edit_rebuild`.
- Reset-between-popups still holds (`flags_disabled` defaults to `False`, assigned unconditionally each `popup()`), guarded by the existing `test_flags_disabled_resets_between_popups`.
- No dead code left behind; `_flag_checkboxes()` remains in use by `_update_flags` and `_collect_flags`.
- CI green across all 8 checks (ubuntu/macos/windows × 3.11/3.14, lint, cla). Branch is even with `main` (0 commits behind), so the green reflects the merged result.

Non-blocking nit (not a defect, no action needed): `self.edit.setText(text)` at the top of `popup()` triggers `textChanged` → `_update_flags` once with the stale `self._flags_disabled` before line `self._flags_disabled = flags_disabled` runs. It's harmless because the explicit rebuild right after always overwrites it, but moving the `self._flags_disabled = flags_disabled` assignment above the `setText` call would remove the ordering dependency entirely. Optional cleanup; not worth a re-push on its own.

PR #2244, opened 2026-06-20
feat(examples): make examples/ self-contained with a vendored JSON reader

### LINKED ISSUE #2231: Make examples/ self-contained: add examples/utils.py, drop labelme imports
Part of the v7 effort to make labelme an application with **no public Python API**. Reading the labelme JSON format is the user's job (the PyTorch-`Dataset` model: the framework defines the format; the user writes the loading). The `examples/` scripts must stop importing labelme internals and instead carry their own copy-and-adapt helper.

## What to build

Add a single self-contained `examples/utils.py` at the examples root that reads the labelme JSON format and provides the helpers the example scripts need, depending **only** on `json` / `numpy` / `PIL` — never on `labelme`. Repoint every example script to import from it, so that `examples/` contains zero `labelme.*` imports. Add a CI smoke test that runs each example end-to-end on a tiny fixture, so the file users actually copy is known to work.

`examples/utils.py` carries its own copy of the rasterization helpers (`shape_to_mask`, `shapes_to_label`, `img_data_to_arr`, plus JSON load that decodes `imageData`). This duplicates ~150 lines from labelme's internal `_utils`/shape code **deliberately**: the two copies have different owners and lifecycles (the internal copy serves the GUI and stays free to refactor; the examples copy is frozen reference code users vendor). That decoupling is the point.

Example scripts to repoint: `examples/instance_segmentation/labelme2coco.py`, `examples/instance_segmentation/labelme2voc.py`, `examples/bbox_detection/labelme2voc.py`, `examples/tutorial/export_json.py`, `examples/tutorial/draw_json.py`. They currently use `labelme.LabelFile(filename=...)` (read `.image_data`, `.shapes`) and `labelme.utils.{img_data_to_arr, shape_to_mask, shapes_to_label}`.

## Acceptance criteria

- [ ] `examples/utils.py` exists at the examples root and imports nothing from `labelme`
- [ ] `examples/utils.py` reads the labelme JSON format and exposes equivalents of the loading + `img_data_to_arr` + `shape_to_mask` + `shapes_to_label` helpers the scripts need
- [ ] Every script under `examples/` imports those helpers from `examples/utils.py`; `grep -rn "import labelme\|from labelme" examples/` returns nothing
- [ ] A CI smoke test runs each example script end-to-end on a small fixture and passes
- [ ] Existing test suite (unit + e2e) stays green

## Blocked by

None - can start immediately

wkentaro:
Closes #2231

Part of the v7 effort to make labelme an application with no public Python API. The `examples/` scripts no longer import labelme internals; they read the labelme JSON format themselves via a vendored helper, the way a PyTorch `Dataset` reads whatever format its data lives in.

## Summary
- Add `examples/utils.py`: a self-contained reader for the labelme JSON format (`load_label_file`, `img_data_to_arr`, `shape_to_mask`, `shapes_to_label`) depending only on `json`/`numpy`/`PIL`, never on `labelme`. It deliberately carries its own copy of the rasterization helpers so the examples and labelme's internals can evolve independently.
- Repoint all five scripts (`instance_segmentation/labelme2coco.py`, `instance_segmentation/labelme2voc.py`, `bbox_detection/labelme2voc.py`, `tutorial/export_json.py`, `tutorial/draw_json.py`) onto `examples/utils.py`. `grep -rn "import labelme\|from labelme" examples/` now returns nothing.
- Guard it in CI: the existing "Run examples" job (which already runs every script end-to-end) now also fails if a `labelme` import is reintroduced into `examples/`.

## Test plan
- [x] every example script runs end-to-end via the CI "Run examples" job (incl. `primitives.json`, which exercises every shape type including `mask`)
- [x] vendored helpers produce byte-identical output to labelme's across all annotated fixtures
- [x] `grep -rnE "import labelme|from labelme" examples/` returns nothing
- [x] full suite green: `uv run pytest tests/` (617 passed)
- [x] `make lint` (ruff format/check, ty, taplo)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,8 @@ jobs:
         run: |
           labelme --help
           labelme --version
+          # examples/ must stay self-contained: no labelme imports (#2231)
+          ! grep -rnE "import labelme|from labelme" examples/
           (cd examples/primitives && ../tutorial/export_json.py primitives.json && rm -rf primitives)
           (cd examples/tutorial && rm -rf apc2016_obj3_json && ./export_json.py apc2016_obj3.json && python load_label_png.py && git checkout -- .)
           (cd examples/semantic_segmentation && rm -rf data_dataset_voc && ./labelme2voc.py data_annotated data_dataset_voc --labels labels.txt && git checkout -- .)
--- a/examples/bbox_detection/labelme2voc.py
+++ b/examples/bbox_detection/labelme2voc.py
@@ -7,14 +7,15 @@ from pathlib import Path
 
 import imgviz
 
-import labelme
-
 try:
     import lxml.builder  # type: ignore
     import lxml.etree  # type: ignore
 except ImportError:
     print("Please install lxml:\n\n    pip install lxml\n")
     sys.exit(1)
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+import utils  # noqa: E402  # examples/utils.py, vendored alongside this script
 
 
 def main() -> None:
@@ -60,7 +61,7 @@ def main() -> None:
     for path in sorted(Path(args.input_dir).glob("*.json")):
         print("Generating dataset from:", path)
 
-        label_file = labelme.LabelFile(filename=str(path))
+        label_file = utils.load_label_file(str(path))
 
         base = path.stem
         out_img_file = output_dir / "JPEGImages" / f"{base}.jpg"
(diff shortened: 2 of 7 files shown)


wkentaro:

Reviewed the diff across correctness, removed-behavior, cross-file, reuse, simplification, and convention angles. No actionable findings.

- Vendored `shape_to_mask` / `shapes_to_label` in `examples/utils.py` are byte-for-byte copies of `labelme/utils/shape.py`; the deliberate duplication is the point of the PR (independent lifecycles).
- Every `utils.*` call site maps to a vendored function with a matching signature (positional and kwargs both correct), and all `shape[...]` accesses are covered by the keys `load_label_file` emits.
- `sys`/`Path` are imported before each `sys.path.insert`, and dropping `assert image_data is not None` is safe because `LabeledImage.image_data` is non-optional `bytes` (the loader raises if neither `imageData` nor a readable `imagePath` exists).
- Smoke-verified `export_json.py` end-to-end on `primitives.json` (null `imageData` + every shape type incl. `mask`); all outputs produced. All 8 CI checks green, merge state CLEAN.

Minor note (not a blocker): `load_label_file` reads external images via `read_bytes()`, skipping the EXIF-orientation re-encoding / TIFF normalization that labelme's own `read_image_file` applies. Byte-identical holds for every fixture CI exercises, but the property would not generalize to an EXIF-rotated external JPEG; worth a one-line docstring caveat at most.

PR #2251, opened 2026-06-23
build: move update_translate from lint to the release gate

wkentaro:
`make lint` ran `update_translate` on every PR, regenerating translations via `pyside6-lupdate`/`-lrelease` over every source file and all 22 languages, which is slow. Translations only need to be current at release, not on every lint, so this moves the regeneration into the release gate.

`update_translate` is now a prerequisite of `check_translate` (run in `publish.yml`) instead of `lint`. The now-redundant `git diff --exit-code labelme/translate` is dropped from lint since nothing regenerates there anymore.

Encoding it as a Make dependency keeps the regenerate-before-grep ordering in one place. As a bonus it closes a gap in the old release gate: `check_translate` greps committed `.ts` for `type="unfinished"`, so a brand-new untranslated string that was never run through `update_translate` was simply absent from `.ts` and slipped past. Now `check_translate` regenerates first, surfacing such strings as `unfinished` and blocking the release.

## Test plan
- [x] `make lint` no longer regenerates translations (no `pyside6-lupdate` invocation)
- [x] `make check_translate` regenerates `.ts`/`.qm` then fails on any `type="unfinished"`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ help:
 setup:  # Setup the development environment
 	$(call exec,uv sync)
 
-lint: update_translate  # Lint code
+lint:  # Lint code
 	$(call exec,uv run ruff format --check)
 	$(call exec,uv run ruff check)
 	$(call exec,uv run ty check --no-progress)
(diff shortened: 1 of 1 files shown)


wkentaro:

Reviewed at medium effort; no meaningful findings.

- Moves \`update_translate\` out of \`lint\` and makes it a prerequisite of \`check_translate\` instead. Regenerating translations is a release-time concern, not a read-only lint step, so this is the right altitude.
- The translation-completeness gate is not dropped, only relocated: \`check_translate\` is invoked by \`.github/workflows/publish.yml\`, so unfinished translations still block a release.
- All CI checks green; branch sits on current \`main\` (CLEAN, 0 divergence), so the existing green is authoritative.

PR #2255, opened 2026-06-25
fix(tiff): normalize TIFF display over finite pixels only

wkentaro:
A float TIFF with a single non-finite pixel loaded as an all-black image: `np.nanmax` returned `inf`, so every finite pixel divided by `inf` and collapsed to 0. The existing NaN guard shows the function already meant to handle non-finite TIFFs, but `inf` slipped through.

`_normalize_to_uint8` now computes the min/max over finite pixels only, and saturates non-finite pixels deterministically: `+inf -> 255`, `-inf -> 0`, `nan -> 0` (previously the `nan` path also hit an undefined `astype(uint8)` cast and a `RuntimeWarning`).

## Test plan
- [x] Added unit tests for `_normalize_to_uint8` (finite range, constant, all-non-finite, and the mixed `+inf`/`-inf`/`nan` regression): `tests/unit/_label_file_test.py`
- [x] Smoke: a float TIFF with an `inf` pixel now renders a full 0-255 gradient through the real `_imread` path instead of black
- [x] `uv run pytest tests/` (680 passed), `ruff check`, `ruff format --check`, `ty check`

### DIFF
--- a/labelme/_label_file.py
+++ b/labelme/_label_file.py
@@ -333,9 +333,13 @@ def _imread_tiff(filename: str) -> PIL.Image.Image:
 
 def _normalize_to_uint8(arr: NDArray) -> NDArray[np.uint8]:
     arr = arr.astype(np.float64)
-    min_val = np.nanmin(arr)
-    max_val = np.nanmax(arr)
-    if np.isnan(min_val) or np.isnan(max_val) or max_val - min_val == 0:
+    finite = arr[np.isfinite(arr)]
+    if finite.size == 0:
+        return np.zeros(arr.shape, dtype=np.uint8)
+    min_val = finite.min()
+    max_val = finite.max()
+    if max_val - min_val == 0:
         return np.zeros(arr.shape, dtype=np.uint8)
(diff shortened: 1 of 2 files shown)


wkentaro:

Reviewed the change to `_normalize_to_uint8`. The fix is sound: it computes min/max over finite pixels only, then lets `np.clip` saturate `±inf` (`+inf -> 255`, `-inf -> 0`) and `np.nan_to_num` zero out `nan`, replacing the old path where a single `inf` pixel made `nanmax` return `inf` and collapsed every finite pixel to black.

- All CI checks green (lint, cla, tests across ubuntu/macos/windows × 3.11/3.14).
- Unit tests cover the finite range, constant array, all-non-finite, and the mixed `+inf`/`-inf`/`nan` regression; 24 passed locally.
- No rebase needed: branch is current with `main` (0 commits behind, no file overlap).

Non-blocking note (not a regression in this PR, present in the old code too): if a float TIFF spans more than ~1e308 in finite range, `max - min` overflows to `inf` and the image collapses to black. Pathological for real images; out of scope here.

PR #2256, opened 2026-06-25
fix(text-detection): raise ValueError instead of asserting a missing mask

wkentaro:
`get_bboxes_from_texts` collected per-annotation masks behind a bare `assert annotation.mask is not None`. That guard raises `AssertionError` rather than the `ValueError` the rest of the function already raises for malformed responses (missing bounding box, unexpected text), and `assert` is stripped entirely under `python -O`, so a response mixing masked and unmasked annotations could slip a `None` into the mask list. This matches the surrounding malformed-response handling by raising `ValueError` instead.

## Test plan
- [x] `pytest tests/unit/_automation/_text_detection_test.py` — new tests characterize all three mask cases (all present, none present, a later one missing -> `ValueError`)
- [x] `make lint`

### DIFF
--- a/labelme/_automation/_text_detection.py
+++ b/labelme/_automation/_text_detection.py
@@ -51,7 +51,8 @@ def get_bboxes_from_texts(
     if response.annotations and response.annotations[0].mask is not None:
         masks = []
         for annotation in response.annotations:
-            assert annotation.mask is not None
+            if annotation.mask is None:
+                raise ValueError("Mask is missing in the annotation.")
             masks.append(annotation.mask)
 
     return boxes, scores, labels, masks
--- a/tests/unit/_automation/_text_detection_test.py
+++ b/tests/unit/_automation/_text_detection_test.py
@@ -1,0 +1,79 @@
+from __future__ import annotations
+
+import numpy as np
+import osam
+import pytest
+from numpy.typing import NDArray
+
+from labelme._automation._text_detection import get_bboxes_from_texts
+
+
+class _FakeOsamSession:
+    def __init__(self, response: osam.types.GenerateResponse) -> None:
+        self.model_name = "stub"
+        self._response = response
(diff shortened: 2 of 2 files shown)


wkentaro:

Reviewed the diff: the change replaces a bare `assert annotation.mask is not None` with a `raise ValueError("Mask is missing in the annotation.")`, which is the right call (asserts are stripped under `python -O`) and matches the existing convention in the same function (missing bounding box, unexpected text both raise `ValueError`). The new test file covers all three branches: every annotation has a mask, none do (returns `None`), and a later annotation is missing one (raises). `masks` is correctly initialized to `None`.

- Code review (correctness + cleanup): no meaningful findings.
- Commit history: single, clean logical commit (fix + its tests). No `/recommit` needed.
- CI: all 8 checks green; merge state `CLEAN`.

Mergeable as-is, for the maintainer to merge.

wkentaro:

Reviewed the change (assert → `raise ValueError` on a missing mask, plus a new test file covering all three branches of the mask-collection logic).

- The conversion is correct and consistent with the existing style in the same function (line 36 already raises `ValueError` for a missing bounding box). An `assert` would be stripped under `python -O`, so this hardens the error path.
- The new tests exercise the all-masks, no-masks, and mixed-mask (raises) cases. The `_FakeOsamSession` stub matches the signature already used by the sibling `_ai_assist_test.py`.
- CI is green across the full matrix (lint + ubuntu/macos/windows × 3.11/3.14); merge state is CLEAN and the branch is up to date with `main`.

A medium-effort `/code-review` surfaced only nits already matching repo convention and one pre-existing osam-contract edge case (when the *first* annotation lacks a mask but a later one has one, masks are silently dropped rather than raising). That predates this PR and is out of scope; not blocking.

Ready for maintainer merge.

PR #2257, opened 2026-06-25
test(shape): cover move_vertex and translate mutation methods

wkentaro:
`Shape.insert_point` and `remove_point` are unit-tested, but their sibling
mutation methods `move_vertex` (drag a vertex) and `translate` (drag a whole
shape) had no direct coverage despite being core editing primitives. These tests
close that gap: they pin sub-pixel float precision, assert `move_vertex` changes
only the target vertex, and assert `translate` shifts every point by the offset
without changing the point count.

## Test plan
- [x] `uv run pytest tests/unit/_shape_test.py -q` (49 passed)
- [x] `uv run ruff check` / `uv run ty check` on the changed file

### DIFF
--- a/tests/unit/_shape_test.py
+++ b/tests/unit/_shape_test.py
@@ -243,6 +243,29 @@ def test_remove_point_is_noop_for_non_polyline() -> None:
     assert len(shape.point_labels) == 2
 
 
+def test_move_vertex_replaces_only_target_point() -> None:
+    shape = _make_square_polygon()
+
+    shape.move_vertex(i=1, pos=(3.5, -1.25))
+
+    assert len(shape.points) == 4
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding coverage for `Shape.move_vertex` and `Shape.translate`. Reviewed:

- `move_vertex(i=1, ...)` correctly asserts only `points[1]` changes and the other three vertices are untouched.
- `translate(offset=...)` correctly asserts all four vertices shift by the offset (verified arithmetic against the `(0,0)-(10,10)` square).
- Both tests pass locally; names, kwargs, the `_make_square_polygon` helper, and `pytest.approx` usage match repo conventions.

CI is green across all 8 matrix jobs; branch is current with `main` (0 commits behind) and `MERGEABLE`. No findings.

PR #2262, opened 2026-06-30
fix: reset label flags on each new-shape popup

wkentaro:
Checking a flag on one shape left it checked when the label dialog reopened for the next new shape, until you manually unchecked it. The checkboxes from the previous popup were never cleared, and the no-explicit-flags path re-applies the checked state of whatever boxes still exist. On a fresh popup the dialog now clears the leftover flag checkboxes before rebuilding them, so each new shape starts with all flags unchecked. The carry-over that live typing relies on is unchanged.

## Test plan
- [x] `pytest tests/unit/widgets/label_dialog_characterize_test.py` — two new tests pin that checked flags (set interactively, and via an edited shape's flags dict) do not leak into the next new-shape popup; both fail without the fix
- [x] `ruff check` and `ty check` clean

### DIFF
--- a/labelme/_widgets/label_dialog.py
+++ b/labelme/_widgets/label_dialog.py
@@ -260,6 +260,9 @@ class LabelDialog(QtWidgets.QDialog):
         if flags is not None:
             self._show_popup_flags(flags)
         else:
+            # _update_flags keeps the checked state of boxes that already exist
+            # (wanted while typing); clear them so a fresh popup starts unchecked.
+            self._clear_flags_layout()
             self._update_flags(self.edit.text())
 
         matches = self.label_list.findItems(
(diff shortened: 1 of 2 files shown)


wkentaro:

Finalized this tick:

- **Rebase:** not needed. The branch is even with `main` (no commits gained since the merge-base, no file overlap), so the green CI already reflects the merged result.
- **Review:** ran a workflow-backed code review. One confirmed observation and three refuted. The confirmed one is design-altitude, not a defect: `popup()` pre-clears the layout before `_update_flags()`, which snapshots `current_states` and then clears again — so on this path the snapshot is always empty and the carry-over branch is dead. Behavior is correct and the added comment documents the coupling, so this does not block merge. *(Optional, non-blocking: an explicit `preserve_state=False` argument to `_update_flags()` would express the intent without the redundant double-clear — left to your judgment.)*
- **Tests:** `tests/unit/widgets/label_dialog_characterize_test.py` passes locally (53 passed), including the two new leak-guard tests; CI is green across all 6 platform/version combos.
- **CI:** green. No push this tick, so the existing checks are authoritative.

Recommending merge. Merging stays your call.

PR #2263, opened 2026-06-30
fix: keep label dialog fully on screen when a label has many flags

wkentaro:
A user reported that when a label has many flags the dialog grew taller
than the screen, so the description field and lower flags fell off the
bottom with no way to drag them back; the cursor-anchored popup could
also open half off-screen near a display edge.

The flags now live in a scroll area sized to `min(content, 150px)`: the
dialog stays compact, scrolls past 150px, and the flag rows keep
identical spacing whether or not the scrollbar is shown. The popup is
also clamped to the available geometry of the screen under the cursor so
it always opens fully on-screen.

## Test plan
- [x] `tests/unit/widgets/label_dialog_screen_fit_test.py`: flag area caps at 150px and scrolls, hugs content below the cap, spacing identical with and without the scrollbar, popup stays on screen from all four corners
- [x] `ruff` + `ty` clean; `tests/unit/widgets` + `tests/e2e/label_dialog_flags_test.py` green (248 passed)

### DIFF
--- a/labelme/_widgets/label_dialog.py
+++ b/labelme/_widgets/label_dialog.py
@@ -285,7 +285,11 @@ class LabelDialog(QtWidgets.QDialog):
         self.edit.setFocus(QtCore.Qt.FocusReason.PopupFocusReason)
 
         if move:
-            self.move(position if position is not None else QtGui.QCursor.pos())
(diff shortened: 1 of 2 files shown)


wkentaro:
image

PR #2273, opened 2026-07-01
fix: keep annotation label text visible on styles that collapse the text sub-rect

### LINKED ISSUE #1228: Incorrectly clipped polygon labels item text

/usr/bin/python
Python 3.10.9
labelme 2b225b2b384bdb25fb37678a9d28b8e545732473

### What OS are you using?

ArchLinux

### Describe the Bug

On my GNOME desktop, the item text of Polygon Labels entries is practically invisible:

Screenshot from 2023-01-12 00-28-04

The problem is caused by this code:

https://github.com/wkentaro/labelme/blob/2b225b2b384bdb25fb37678a9d28b8e545732473/labelme/widgets/label_list_widget.py#L46-L57

Since the delegate does not actually set any text, the reported text width is too narrow. `textRect` must account for the width of the text rendered by the `QTextDocument` instance.

The following patch fixes the problem:

```patch
diff --git a/labelme/widgets/label_list_widget.py b/labelme/widgets/label_list_widget.py
index 6164ef9..ace2b6e 100644
--- a/labelme/widgets/label_list_widget.py
+++ b/labelme/widgets/label_list_widget.py
@@ -4,6 +4,7 @@ from qtpy import QtGui
 from qtpy.QtGui import QPalette
 from qtpy import QtWidgets
 from qtpy.QtWidgets import QStyle
+import math
 
 
 # https://stackoverflow.com/a/2039745/4158863
@@ -44,6 +45,7 @@ class HTMLDelegate(QtWidgets.QStyledItemDelegate):
             )
 
         textRect = style.subElementRect(QStyle.SE_ItemViewItemText, options)
+        textRect.setWidth(int(math.ceil(self.doc.idealWidth())))
 
         if index.column() != 0:
             textRect.adjust(5, 0, 0, 0)
```
which sets the clip region correctly:
Screenshot from 2023-01-12 00-40-15

### Expected Behavior

Item text is visible.

### To Reproduce

Run ArchLinux with GNOME desktop and packages installed using `pacman`.

wkentaro:
`HTMLDelegate.paint` empties `opt.text` before asking the style for the text sub-rect via `subElementRect(SE_ItemViewItemText, ...)`, so some desktop styles (e.g. GNOME/Adwaita) return a rect too narrow for the rendered HTML and clip the annotation label — on some setups making the text practically invisible. The fix widens the clip rect to the document's ideal width (`max(width, ceil(doc.idealWidth()))`), so it can only grow, never shrink a rect a well-behaved style already sized correctly.

The bug is style-dependent and can't be reproduced on this platform, so the regression test forces the same condition deterministically: it renders the delegate with a collapsed text sub-rect and asserts the label ink is drawn past the collapsed width instead of being clipped away (red on `main`, green here).

Closes #1228

## Test plan
- [x] `tests/unit/widgets/label_list_widget_test.py` — new test, verified failing on `main` and passing here
- [x] `uv run pytest tests/` — full suite green (703 passed)
- [x] `uv run ruff format --check && uv run ruff check && uv run ty check` — clean
- [x] Real-widget render smoke: `LabelListWidget` populated with real shape labels renders without clipping or crash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed annotation label text being clipped or invisible in the label list on some desktop styles (e.g. GNOME/Adwaita) by widening the delegate's text clip rect to the rendered width ([#2273](https://github.com/wkentaro/labelme/pull/2273))
 - Fixed `line` and two-point `linestrip` shapes being unselectable by click ([#2107](https://github.com/wkentaro/labelme/pull/2107))
 - Fixed the canvas going blank after "Delete File" (Delete File now removes only the label JSON and keeps the image on screen) ([#2138](https://github.com/wkentaro/labelme/pull/2138))
 - Fixed notch artifacts at interior vertices of linestrip masks by using PIL's `joint="curve"` rasterization ([#2163](https://github.com/wkentaro/labelme/pull/2163))
--- a/labelme/_widgets/label_list_widget.py
+++ b/labelme/_widgets/label_list_widget.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import html
+import math
 from collections.abc import Iterator
 from typing import NamedTuple
 from typing import cast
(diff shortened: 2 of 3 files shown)


wkentaro:

Finalized this tick:

- **Rebase:** not needed — branch sits exactly at current `main` (zero divergence), so the green CI already reflects the merged result.
- **Review gate:** `/code-review` (medium) over the diff found no meaningful issues. The one substantive line widens the delegate's text sub-rect to `ceil(doc.idealWidth())` *before* it becomes the clip rect, so styles (e.g. Adwaita/GNOME) that return a collapsed sub-rect no longer clip the label; `max()` leaves normal styles untouched. Single-column list, so no adjacent-column overflow.
- **History:** single, well-formed commit — nothing to reshape.
- **CI:** all 8 test matrix jobs + lint green. No push this tick, so the existing checks are authoritative.
- **Test:** the new characterization test (`label_list_widget_test.py`) reproduces the collapsed-sub-rect case and passes locally.

Ready for maintainer merge.

PR #2276, opened 2026-07-01
fix: make a shape hidden while hovered inert to the pointer

### LINKED ISSUE #2250: Hidden shape stays hover-interactive if it was hovered when hidden
> *This was generated by AI.*

## Summary

A shape that is hidden *while it is the hovered shape* stays interactive: its hover highlight, cursor change, status hints, and click-to-select/drag all keep working even though the shape is not drawn. Hiding a shape should make it inert to pointer interaction.

## Steps to reproduce

1. Open an image and create or load a polygon.
2. In edit mode, move the cursor over the polygon (so it becomes the hovered shape — the cursor changes to the open-hand / pointing-hand cursor).
3. **Without moving the mouse**, hide that shape (uncheck its visibility in the Polygon Labels list).
4. Move the cursor slightly.

**Observed:** the cursor still switches to the hover/grab cursor over the now-invisible shape, the vertex/edge/body highlight still appears, and the shape can still be selected and dragged.

**Expected:** an invisible shape is not hover-highlighted, does not change the cursor, and cannot be selected or dragged via the canvas.

## Cause

The canvas hit-test examines the previously-hovered shape *first* and prepends it to the candidate list unconditionally — the `.visible` filter is only applied to the remaining shapes. Combined with `set_shape_visible` not clearing the hovered shape, a shape hidden while hovered remains a hit-test candidate until the pointer moves off it.

This is long-standing behavior (it predates the recent canvas-interaction refactor and is preserved by it), so it is a pre-existing bug rather than a regression.

## Suggested fix

Exclude the priority/hovered shape from hit-testing when it is not visible (or clear the hovered shape when its visibility is toggled off).

wkentaro:
The canvas hover hit-test (`find_hover_target` → `_build_candidates`) prepended the previously-hovered shape as a priority candidate **without** checking visibility, while every other shape was already visibility-filtered. So a shape hidden while it was the hovered shape stayed hover-interactive — its highlight, cursor change, and click-to-select kept working on the now-invisible shape until the pointer moved off it. The fix applies the same `.visible` check to the priority slot.

The regression test drives the real `Canvas` through actual mouse events (hover a shape body → hide it → move slightly) and asserts it becomes inert (no hovered shape, no override cursor). Verified red on `main`, green here.

Closes #2250

## Test plan
- [x] `tests/unit/widgets/canvas_interaction_characterization_test.py::test_shape_hidden_while_hovered_stops_being_hover_interactive` — new test, verified failing on `main` and passing here
- [x] `uv run pytest tests/` — 705 passed
- [x] `uv run ruff format --check && uv run ruff check && uv run ty check` — clean

## Deferred to a follow-up

Scoped to the reported hover path. Review surfaced sibling cases where a hidden shape stays **movable** via *selection* rather than hover — all reachable only by hiding a shape through a keyboard shortcut while an interaction is active (the Polygon Labels checkbox path is covered here, since clicking the list forces a hit-test-refreshing mouse move):

1. A body drag continues to move the shape if it is hidden mid-drag (`_continue_left_button_drag` → `_drag_selected_shapes` is driven by `selected_shapes`, not hover).
2. A vertex press-drag with no intervening mouse move (stale `_hovered_vertex`).
3. Arrow-key nudging of a selected-but-hidden shape (`_move_by_keyboard` reads `selected_shapes` with no visibility check).

These are **selection**-inertness rather than hover, and fixing them cleanly needs a product decision (should hiding a shape also deselect it / cancel an in-progress drag?), so they are intentionally left out of this bugfix and warrant a separate tracking issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a hidden shape staying hover-interactive (highlight, cursor, click-to-select) when it was the hovered shape at the moment it was hidden ([#2276](https://github.com/wkentaro/labelme/pull/2276))
 - Fixed annotation label text being clipped or invisible in the label list on some desktop styles (e.g. GNOME/Adwaita) by widening the delegate's text clip rect to the rendered width ([#2273](https://github.com/wkentaro/labelme/pull/2273))
 - Fixed `line` and two-point `linestrip` shapes being unselectable by click ([#2107](https://github.com/wkentaro/labelme/pull/2107))
 - Fixed the canvas going blank after "Delete File" (Delete File now removes only the label JSON and keeps the image on screen) ([#2138](https://github.com/wkentaro/labelme/pull/2138))
--- a/labelme/_widgets/_canvas_interaction.py
+++ b/labelme/_widgets/_canvas_interaction.py
@@ -89,7 +89,7 @@ def _build_candidates(
     priority_shape: Shape | None,
 ) -> list[Shape]:
     candidates: list[Shape] = []
-    if priority_shape is not None:
+    if priority_shape is not None and priority_shape.visible:
         candidates.append(priority_shape)
     for shape in reversed(shapes):
         if not shape.visible:
--- a/tests/unit/widgets/canvas_interaction_characterization_test.py
+++ b/tests/unit/widgets/canvas_interaction_characterization_test.py
@@ -183,6 +183,44 @@ def test_cursor_is_open_hand_when_hovering_shape_body_in_edit_mode(
 
 
 # ---------------------------------------------------------------------------
+# Cursor shape -- edit mode, shape hidden while hovered (#2250)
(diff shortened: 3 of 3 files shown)


wkentaro:

Finalized this tick:

- **Rebase:** not needed — branch sits exactly at current `main` (zero divergence), so the green CI already reflects the merged result.
- **Review gate:** `/code-review` (medium) found no meaningful issues. The one-line fix adds `and priority_shape.visible` to the priority-candidate guard in `_build_candidates`, so a shape hidden while it was the hovered shape is no longer re-offered to the hit-test — consistent with the loop's own `if not shape.visible: continue` skip a few lines down. Correct and minimal.
- **History:** single, well-formed commit — nothing to reshape.
- **Test:** the new `test_shape_hidden_while_hovered_stops_being_hover_interactive` reproduces #2250 (hover shape, hide it, nudge pointer, assert `hovered_shape is None` and no override cursor); all 19 canvas-interaction characterization tests pass locally.
- **CI:** all 8 test matrix jobs + lint green. No push this tick, so the existing checks are authoritative.

Ready for maintainer merge.

PR #2288, opened 2026-07-03
fix: mark file dirty on undo so auto-save records it

### LINKED ISSUE #2287: ctrl +z 的操作无法被自动保存所记录

windows +V6.31

### What OS are you using?

windows

### Describe the Bug

Image
比如我把标注放到旁边，然后使用ctrl+z 进行回正，我开的是自动保存模式，当我ctrl +z回正后切换下一张 再切回来的时候，发现这张图片的标注还是偏的。

### Expected Behavior

_No response_

### To Reproduce

_No response_

wkentaro:
Fixes #2287. With auto-save on, moving a shape is saved to disk, but pressing Undo (Ctrl+Z) restored the shape only in memory: `undo_shape_edit` never marked the file dirty, so auto-save never wrote the reverted state back. Switching to another image and returning showed the pre-undo (offset) annotation.

`undo_shape_edit` now routes through `mark_dirty`, which already refreshes the undo action's enabled state and, with auto-save on, writes the restored shapes to disk (and otherwise flags the file dirty / enables Save). The bug predates v7 (present in v6.3.1), so it gets a CHANGELOG entry.

## Test plan
- [x] `test_auto_save_on_undo` added: move a shape (auto-saved), Undo, assert the on-disk label matches the original points. Fails on `main` (saved x off by 5.0), passes with the fix.
- [x] `uv run pytest tests/e2e/auto_save_test.py`
- [x] `uv run ruff check` / `uv run ty check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed Undo (Ctrl+Z) not triggering auto-save; undoing a shape edit now marks the file dirty, so with auto-save on the restored state is written to disk instead of leaving the previous edit saved ([#2288](https://github.com/wkentaro/labelme/pull/2288))
+
 ## [7.0.1] - 2026-07-03
 
 ### Changed
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -1451,7 +1451,7 @@ class MainWindow(QtWidgets.QMainWindow):
         self._canvas_widgets.canvas.restore_last_shape()
         self._docks.label_list.clear()
         self._load_shapes(self._canvas_widgets.canvas.shapes)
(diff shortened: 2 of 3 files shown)


wkentaro:

**Verdict: recommend-merge**

Correct, minimal fix at the right altitude. Routing `undo_shape_edit` through `mark_dirty` re-establishes the exact `undo.setEnabled(...)` behavior it replaces and reuses the existing dirty/auto-save mechanism (matching the `_load_shapes(...)` → `mark_dirty()` pattern in `_insert_shapes`) rather than special-casing a save call on the undo path. The label list is repopulated before `mark_dirty`, so `save_labels` persists the restored shapes rather than stale state. Regression test fails on `main` and passes with the fix, and the CHANGELOG entry is justified (bug predates v7).

Non-blocking nit: the test invokes `undo_shape_edit()` directly rather than simulating Ctrl+Z, so it doesn't cover the action wiring, but that matches the file's existing style and the method-level behavior is the point of the fix.

PR #2293, opened 2026-07-06
fix(logging): substitute values in loguru warnings via brace-style placeholders

wkentaro:
Two `logger.warning` calls used printf-style `%r`/`%d` placeholders, but loguru
renders messages with `str.format`, which silently ignores unreferenced
positional args. So the no-op point-removal warning (`Shape.remove_point`) and
the empty save-path warning both logged their literal placeholder text with the
diagnostic values dropped. Both are switched to loguru's brace-style
placeholders, matching every other `logger.*` call in the codebase.

A grep across `labelme/` confirms these were the only two remaining printf-style
logger placeholders.

## Test plan
- [x] Added `test_remove_point_noop_logs_substituted_values`, which captures the
  rendered message via a loguru sink and asserts the substituted values appear
- [x] `uv run pytest tests/unit -q` (556 passed)
- [x] `uv run ruff check` / `uv run ty check --no-progress` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed two warning log messages rendering literal `%r`/`%d` placeholders instead of their values (the no-op point-removal warning and the empty save-path warning); loguru formats with brace-style placeholders, so the diagnostic values were silently dropped ([#2293](https://github.com/wkentaro/labelme/pull/2293))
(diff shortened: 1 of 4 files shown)


wkentaro:

Reviewed at the own-PR finalize tier.

- **Correctness**: The fix is right. loguru formats messages with brace-style (`str.format`) placeholders, so the previous `%r`/`%d` were emitted literally and the diagnostic values (`label_path`, `shape_type`, `len(points)`) were silently dropped. Switching to `{!r}` / `{:d}` restores substitution.
- **Coverage**: `git grep` across the branch confirms no remaining `%r`/`%s`/`%d` in `logger.*` calls, so the two sites here are the complete set. The added `test_remove_point_noop_logs_substituted_values` asserts the substituted output.
- **Rebase**: not needed. Branch is directly on current `main` (0 commits gained since merge-base, no file overlap, merge state CLEAN).
- **CI**: green across all 9 test jobs (ubuntu/macos/windows × 3.12/3.13/3.14) plus lint and cla. No push made this tick, so these checks are authoritative.

Single clean, self-contained commit. Recommending merge; the merge call stays with the maintainer.

PR #2294, opened 2026-07-06
refactor: build ToolBar actions via the shared add_actions helper

wkentaro:
`ToolBar.__init__` hand-rolled the same `None → addSeparator / else → addAction`
dispatch that `labelme._utils.add_actions` already provides and that the rest of
the app uses for every menu and toolbar. This was the one place left reinventing
it. Route `ToolBar` through the canonical helper so the dispatch lives in one
place.

`ToolBar.addAction` stays overridden, so `add_actions` still wraps each `QAction`
in a `QToolButton` exactly as before, and the `QWidgetAction` bypass is preserved.
The helper's extra `QMenu` branch is unreachable here (`actions` is typed
`list[QAction | None]`), so behavior is unchanged.

## Test plan
- [x] Drove the real `ToolBar` under `QT_QPA_PLATFORM=offscreen` in both
      orientations; the rendered tool buttons, labels, and separators are
      byte-identical to `main`.
- [x] `tests/unit/widgets/tool_bar_test.py` (21) and the full suite (709) pass
- [x] `ruff check`, `ruff format --check`, `ty check` clean

### DIFF
--- a/labelme/_widgets/tool_bar.py
+++ b/labelme/_widgets/tool_bar.py
@@ -6,6 +6,8 @@ from PySide6 import QtGui
 from PySide6 import QtWidgets
 from PySide6.QtCore import Qt
 
+from .._utils import add_actions
+
 _OBJECT_NAME_SUFFIX: Final = "ToolBar"
 _FONT_SCALE_FACTOR: Final = 0.8
 _VERTICAL_SEPARATOR_STYLE: Final = (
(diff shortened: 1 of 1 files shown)


wkentaro:

A clean, minimal reuse refactor: the 5-line separator/action loop in `ToolBar.__init__` is replaced by the already-exported `labelme._utils.add_actions` helper.

- **Behaviorally equivalent** — `add_actions` handles `None` → `addSeparator()` and otherwise `addAction()`, exactly as the removed loop did. Its extra `QMenu` branch is unreachable here since `ToolBar` actions are typed `QAction | None`.
- **No circular import** — `add_actions` lives in `labelme/_utils/qt.py`, which does not import from `labelme/_widgets`.
- **Green** — full CI matrix (lint + tests across macOS/Ubuntu/Windows × 3.12–3.14) passes; merge state is `CLEAN` and the branch is even with `main`.

History is a single, well-scoped commit. Ready for maintainer merge.

PR #2295, opened 2026-07-06
fix(config): tolerate empty config sections instead of crashing at startup

wkentaro:
A bare `shortcuts:` or `ai:` section in the config file parses as YAML `null`. `_migrate_config_from_file` then called `.pop()` / `.get()` on it, raising an uncaught `AttributeError` that crashed labelme at startup — `_load_config` only catches `ValueError`, so it never reached the config-error dialog. `_update_dict` also overwrote the default section with `None`, which would crash the app downstream where sections are indexed as plain dicts (`config["ai"]["default"]`, `config["shortcuts"][...]`).

Empty sections now keep their defaults: migration guards each section with `or {}`, and the merge skips a `None` override of a dict-typed section. The `_update_dict` guard is general — it covers every dict-typed section (`shape`, `ai`, `canvas`, the docks, `shortcuts`, ...), not just the two that triggered the report — and matches the existing `raw.get("flags") or {}` tolerance in `_label_file.py`.

## Out of scope

A non-`None`, non-dict override of a dict-typed section (e.g. `shortcuts: "oops"`) still fails uncaught, and crashes even earlier — inside `_migrate_config_from_file`, before the merge is reached. Handling that malformed-config class is a separate, larger change (harden both migration and merge, decide raise-vs-tolerate) and is left for a follow-up.

## Test plan

- [x] `tests/unit/_config_test.py`: added `test_migrate_tolerates_empty_section` and `test_load_config_empty_section_keeps_defaults` (parametrized over `shortcuts` / `ai`); both fail with the original `AttributeError` before the fix.
- [x] Verified end-to-end via the e2e harness: launching the app with a `shortcuts:`/`ai:` empty-section config crashes at startup before the fix, and boots keeping defaults after.
- [x] `make lint` (ruff format + ruff + ty) and the full `pytest` suite (713) pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed two warning log messages rendering literal `%r`/`%d` placeholders instead of their values (the no-op point-removal warning and the empty save-path warning); loguru formats with brace-style placeholders, so the diagnostic values were silently dropped ([#2293](https://github.com/wkentaro/labelme/pull/2293))
+- Fixed a startup crash when the config file contains an empty section such as a bare `shortcuts:` or `ai:`; empty sections now keep their defaults instead of raising an uncaught `AttributeError` ([#2295](https://github.com/wkentaro/labelme/pull/2295))
 - Fixed noisy Qt log lines flooding the terminal (notably the macOS per-keypress `qt.qpa.keymapper: Mismatch between Cocoa and Carbon` warning); Qt logging is now routed through the app logger with these harmless lines filtered out while genuine Qt warnings still surface ([#2292](https://github.com/wkentaro/labelme/pull/2292))
 
 ## [7.0.2] - 2026-07-03
--- a/labelme/_config/__init__.py
+++ b/labelme/_config/__init__.py
@@ -24,7 +24,12 @@ def _update_dict(
             validate_item(key, value)
         if key not in target_dict:
             raise ValueError(f"Unexpected key in config: {key}")
-        if isinstance(target_dict[key], dict) and isinstance(value, dict):
+        target_is_section = isinstance(target_dict[key], dict)
+        if target_is_section and value is None:
+            # An empty section (e.g. a bare `shortcuts:`) parses as None; keep the
+            # defaults instead of wiping the whole section.
+            continue
+        if target_is_section and isinstance(value, dict):
             _update_dict(
                 cast(dict[str, object], target_dict[key]),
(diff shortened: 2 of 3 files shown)


wkentaro:

Reviewed at the own-PR finalize tier.

- **Fix is correct and minimal.** A bare `shortcuts:`/`ai:` section parses as YAML `None`; the old code called `.pop()`/`.get()` on it in `_migrate_config_from_file` and overwrote the default section with `None` in `_update_dict`, crashing at startup (only `ValueError` was caught, so it never reached the config-error dialog). The fix guards both paths: `get(...) or {}` in migration and a `value is None` skip in the merge. The `_update_dict` guard is general across every dict-typed section, not a special-case bandaid.
- **Behavior preserved.** The `shortcuts` variable reuse keeps the same in-place mutation of `config_from_yaml["shortcuts"]` as the old two-call form across all cases (real dict / empty / `None` / missing); non-section `None` overrides still pass through unchanged; `validate_item` still runs for every key.
- **Tests.** Two new parametrized tests cover both the migration and the full `load_config` path for `shortcuts`/`ai`; both fail with the original `AttributeError` before the fix. Full suite green.
- **CI green** across all 11 checks (lint + test on ubuntu/macos/windows × py3.12-3.14). Branch is up to date with `main` (0 commits behind), no rebase needed.
- Verified end-to-end: `load_config` with a `shortcuts:`/`ai:` empty-section config now boots keeping defaults (`Ctrl+S`, `Sam2 (balanced)`) instead of crashing.

The documented out-of-scope case (a non-`None`, non-dict override like `shortcuts: "oops"`) is a reasonable, larger follow-up and does not block this fix.

wkentaro:

follow-ups: #2296 (the documented out-of-scope non-`None`, non-dict section override — does not block this merge).

wkentaro:

**Verdict: `recommend-merge`** (re-verdict after the amended commit `9e4fbbd`)

- Core fix is correct and well-scoped: an empty (`None`) config section such as a bare `shortcuts:` or `ai:` now keeps its defaults instead of raising an uncaught `AttributeError` at startup. Verified end-to-end by loading a config with empty `shortcuts:`/`ai:` sections; defaults are preserved and no crash occurs.
- Tests cover both the migration path (`_migrate_config_from_file`) and the full `load_config` path for each section.
- CI is green across all 11 jobs (ubuntu/macos/windows × py3.12-3.14) on the current head; no push this tick, so the existing checks are authoritative.

Scope note: a section set to a non-`None`, non-dict *scalar* (e.g. `shortcuts: false`) still crashes at startup, but that was already the case before this PR and is explicitly captured as follow-up #2296. It does not block this merge.

follow-ups: #2296

PR #2297, opened 2026-07-07
test(label-file): cover malformed shape-field validation branches

wkentaro:
`_load_shape_json_obj` in `labelme/_label_file.py` guards `read_label_file`
against malformed hand-edited annotation files, raising a specific
`TypeError`/`ValueError` per bad field. Only the `points is required` branch was
exercised; the label / shape_type / flags / description / group_id / mask checks
and the individual `points` sub-checks had no regression coverage.

This adds a dedicated parametrized `test_read_label_file_raises_on_malformed_shape_field`
that breaks exactly one field of an otherwise-valid shape per case (via the same
mutator-lambda idiom the sibling malformed-document test already uses), and moves
the one pre-existing shape-level case out of `test_read_label_file_raises_read_error_on_malformed`
so the two tests split cleanly: document-level keys vs. per-shape fields.

Test-only; no production code changes.

## Test plan
- [x] `uv run pytest tests/unit/_label_file_test.py -q` (36 passed)
- [x] `uv run pytest tests/unit -q` (567 passed)
- [x] `uv run ruff check` / `ruff format --check` / `uv run ty check` clean on the touched file

### DIFF
--- a/tests/unit/_label_file_test.py
+++ b/tests/unit/_label_file_test.py
@@ -91,7 +91,6 @@ def test_read_label_file_extracts_other_data(
         (lambda raw: raw.pop("imagePath"), "imagePath"),
         (lambda raw: raw.pop("imageData"), "imageData"),
         (lambda raw: raw.pop("shapes"), "shapes"),
-        (lambda raw: raw["shapes"].append({"label": "x"}), "points is required"),
         (lambda raw: raw.update({"imageHeight": 1}), "imageHeight mismatch"),
         (lambda raw: raw.update({"imageWidth": 1}), "imageWidth mismatch"),
     ],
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only change: adds a dedicated parametrized test covering every malformed shape-field validation branch in `read_label_file`, and removes the now-redundant loose `invalid_shape` case from the sibling test.

- All 13 `error_match` strings correspond exactly to the validation messages in `labelme/_label_file.py` (label/points/shape_type/flags/description/group_id/mask).
- Behavior preserved: the removed `points is required` case is subsumed by the new comprehensive `points_missing` case.
- CI green across all matrix jobs (ubuntu/macos/windows × 3.12–3.14), lint, cla.
- Branch is current with `main` (0 commits behind base, merge state CLEAN); no rebase needed.

Ready for maintainer merge.

PR #2298, opened 2026-07-07
fix(ai): drop degenerate 2-point polygon from thin detections

wkentaro:
AI Assist / AI Text Prompt in polygon output mode built a Shape directly from `compute_polygon_from_mask`, which can legitimately collapse a thin, near-collinear detection (a wire, crack, or hairline object) to just 2 points. The builder only rejected results with fewer than 2 points, so a degenerate 2-point "polygon" was emitted. That is not a valid polygon: it has no area, and it later crashed mask conversion, since `shape_to_mask` asserts a polygon has more than 2 points.

This raises the guard to drop results with fewer than 3 points, matching the polygon minimum already enforced everywhere else in the codebase, and consistent with the file's existing silent-drop convention for degenerate detections (e.g. the all-False mask guard in the same function).

## Test plan
- [x] Added `test_shapes_from_detections_polygon_drops_degenerate_thin_mask` (regression: a 1px-thick sliver mask is dropped) and `test_shapes_from_detections_polygon_with_mask_traces_contour` (a normal mask still yields a valid polygon)
- [x] `uv run pytest tests/ -q` (724 passed under Xvfb)
- [x] `make lint` (ruff format + ruff check + ty) clean on the tracked source

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed AI Assist / AI Text Prompt in polygon output mode emitting a degenerate 2-point "polygon" for thin, near-collinear detections; such a shape is not a valid polygon and later crashed mask conversion (`shape_to_mask` asserts more than 2 points), so it is now dropped like other empty detections ([#2298](https://github.com/wkentaro/labelme/pull/2298))
 - Fixed two warning log messages rendering literal `%r`/`%d` placeholders instead of their values (the no-op point-removal warning and the empty save-path warning); loguru formats with brace-style placeholders, so the diagnostic values were silently dropped ([#2293](https://github.com/wkentaro/labelme/pull/2293))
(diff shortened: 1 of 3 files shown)


wkentaro:

Reviewed the change end-to-end:

- **Fix is correct.** `compute_polygon_from_mask` can return fewer than 3 points for thin / near-collinear masks (`approximate_polygon` collapses the near-collinear contour, then the duplicate-last-point drop) or 0 points when no contour is found. The old `len(polygon) < 2` guard let a degenerate 2-point "polygon" through, which is not a valid polygon and later crashes `shape_to_mask` (asserts `> 2` points). Tightening to `< 3` drops it, consistent with how other empty detections are handled.
- **Well-tested.** New tests cover both the valid path (`>= 3` points) and the degenerate thin-mask path (`shapes == []`). Local run: 14 passed.
- **CHANGELOG** entry is in the right `### Fixed` subsection with the PR linked, per repo convention.
- **CI:** all 11 checks green; merge state CLEAN, no rebase needed (main gained 0 commits since the merge-base).

A report-only `/code-review` at medium effort surfaced no correctness, reuse, simplification, efficiency, altitude, or convention findings. Ready for maintainer review and merge.

PR #2300, opened 2026-07-07
fix(config): report non-mapping config sections instead of crashing

wkentaro:
Follow-up to #2295, which made an *empty* config section (a bare `shortcuts:` / `ai:`, parsed as `None`) keep its defaults. This handles the sibling case that #2295 explicitly left out of scope: a section set to a *non-mapping* value, e.g. `shortcuts: oops`.

That case crashed the app at startup with an uncaught `AttributeError` — `_migrate_config_from_file` called `.pop()`/`.get()` on the scalar, and only `ValueError` reaches the config-error dialog. Even past migration, `_update_dict` would overwrite the default section dict with the scalar, moving the crash downstream where sections are indexed as dicts (`config["shortcuts"][...]`).

## Approach

Make the merge the single validation gate:

- `_migrate_config_from_file` skips non-dict sections (via `isinstance` guards), leaving the malformed value in place instead of crashing on it.
- `_update_dict` raises a `ValueError` naming the offending section when a dict-typed section gets a non-`None`, non-mapping override. That flows into the existing "Configuration Errors" dialog (`_app.py` `_load_config`, already the handler for `validate_label`/`shape_color`/`labels` errors), which offers Ignore-and-use-defaults.

Empty sections keep the tolerate-and-default behavior from #2295; only non-`None`, non-mapping overrides are rejected.

## Test plan

- [x] `tests/unit/_config_test.py`: added `test_migrate_leaves_malformed_section_for_merge_to_report` and `test_load_config_malformed_section_raises` (both parametrized over `shortcuts` / `ai`); both fail on the pre-fix code (`AttributeError` in migration; silent overwrite in the merge).
- [x] `make lint` and the full `pytest` suite (730) pass.

Stacked on #2295 (base branch `fix/config-empty-section-crash`); GitHub will retarget this to `main` once #2295 merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed two warning log messages rendering literal `%r`/`%d` placeholders instead of their values (the no-op point-removal warning and the empty save-path warning); loguru formats with brace-style placeholders, so the diagnostic values were silently dropped ([#2293](https://github.com/wkentaro/labelme/pull/2293))
 - Fixed a startup crash when the config file contains an empty section such as a bare `shortcuts:` or `ai:`; empty sections now keep their defaults instead of raising an uncaught `AttributeError` ([#2295](https://github.com/wkentaro/labelme/pull/2295))
+- Fixed a startup crash when a config section is set to a non-mapping value such as `shortcuts: oops`; the malformed section now surfaces in the configuration-error dialog (with an Ignore-and-use-defaults option) instead of crashing ([#2300](https://github.com/wkentaro/labelme/pull/2300))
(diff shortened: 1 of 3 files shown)


wkentaro:

Finalized and endorsed. This makes a config section set to a non-mapping scalar (e.g. `shortcuts: oops`) surface in the "Configuration Errors" dialog (with Ignore-and-use-defaults) instead of crashing at startup with an uncaught `AttributeError`.

Review notes:
- `_update_dict` now rejects a non-`None`, non-mapping override for a section with a named `ValueError` (`"Config section 'shortcuts' must be a mapping, ..."`), making the merge the single validation gate. `_migrate_config_from_file` leaves malformed sections untouched so that gate reports them, while valid dict sections still migrate in place (renames, `ai.default` rewrite) with no regression.
- Cheap code-review pass (medium): no meaningful findings.
- Smoke-verified end-to-end: `shortcuts: oops` raises the named `ValueError`; an empty `shortcuts:` still keeps its defaults; a valid override still loads.
- CI: green (11/11), merge state CLEAN.

Note for merge order: this branch is **stacked on #2295** (it contains that PR's commit), so merge #2295 first, then this. No follow-up issues.

PR #2301, opened 2026-07-07
test(geometry): cover rectangle/mask/oriented_rectangle bbox branches

wkentaro:
`shape_to_xyxy_bbox` had direct tests only for its circle and polygon paths. The
rectangle, mask, and oriented_rectangle branches (the shared min/max path that
backs NMS suppression), plus their too-few-points `None` guards, were untested.
This adds that coverage; the rectangle and mask cases use points that cross on
an axis so a naive "point0 is the min corner" regression would be caught.

## Test plan
- [x] `uv run python -m pytest tests/unit/_automation/_geometry_test.py` (26 passed)
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run ty check` clean

### DIFF
--- a/tests/unit/_automation/_geometry_test.py
+++ b/tests/unit/_automation/_geometry_test.py
@@ -150,6 +150,63 @@ def test_shape_to_xyxy_bbox_returns_none_when_circle_has_only_center() -> None:
     assert shape_to_xyxy_bbox(shape=shape) is None
 
 
+def test_shape_to_xyxy_bbox_rectangle() -> None:
+    shape = Shape(
+        shape_type="rectangle",
+        points=np.array([(10, 4), (1, 12)], dtype=np.float64),
+    )
+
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding branch coverage for `shape_to_xyxy_bbox` (rectangle, mask, oriented_rectangle happy paths plus the too-few-points `None` branch).

- Reviewed the diff: all bbox expectations verified against the source (`min`/`max` over points; `None` when `len(points) < minimum`). Correct.
- Conventions: plain `test_` functions, `-> None` annotations, keyword args, mirrors source layout. All satisfied.
- No rebase needed: `main` advanced 4 commits in files disjoint from this test file, merge state CLEAN, so the existing green still reflects the merged result.
- CI: all 11 checks green (9-way test matrix + lint + cla).
- History: already a single, well-formed conventional commit.

PR #2302, opened 2026-07-07
refactor(shape): extract shared nearest-within-epsilon index helper

wkentaro:
`nearest_vertex_index`, `nearest_edge_index`, and `nearest_rotation_point_index`
each ended with the identical "argmin, then null out if the closest distance
exceeds `epsilon`" tail. Extract it into a private
`_nearest_index_within_epsilon` helper so the selection rule lives in one place;
each function now just computes its own `distances` and delegates.

No behavior change: the extracted body is byte-for-byte the removed logic, and
every caller keeps its own empty/shape-type guard before computing `distances`,
so the helper never sees an empty array.

## Test plan
- [x] `uv run pytest tests/unit/_shape_test.py` (50 tests covering all three functions, incl. within-epsilon / far / empty-shape)
- [x] `uv run pytest tests/` (732 passed)
- [x] `ruff check` / `ruff format --check` / `ty check`

### DIFF
--- a/labelme/_shape.py
+++ b/labelme/_shape.py
@@ -91,6 +91,15 @@ class Shape:
         return copy.deepcopy(self)
 
 
+def _nearest_index_within_epsilon(
+    *, distances: npt.NDArray[np.float64], epsilon: float
+) -> int | None:
+    nearest = int(np.argmin(distances))
+    if distances[nearest] > epsilon:
+        return None
+    return nearest
+
+
 def nearest_vertex_index(
     *,
     shape: Shape,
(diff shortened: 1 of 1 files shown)


wkentaro:

Faithful extraction of the identical `argmin`/epsilon-threshold tail shared by `nearest_vertex_index`, `nearest_edge_index`, and `nearest_rotation_point_index` into `_nearest_index_within_epsilon`. Behavior is byte-for-byte unchanged, and all three call sites still guard against empty `points` before computing `distances`, so `np.argmin` never sees an empty array.

- `/code-review` (medium): clean, no meaningful findings.
- CI: all 11 checks green (lint + test on macOS/Ubuntu/Windows × 3.12–3.14).
- Single well-described commit; no rebase needed (up to date with `main`).

PR #2303, opened 2026-07-07
test(shape-clipboard): cover deep-copy paste and availability signal

wkentaro:
Backfills a unit test file for `ShapeClipboard`, previously exercised only
through the slow e2e suite. The class's `paste()` must hand out deep copies
(a shared-reference regression is called out in
`tests/e2e/shape_editing_test.py`), and `availability_changed` should fire
only when the buffer crosses the empty/non-empty boundary. Both are now
pinned by fast, Qt-display-free unit tests.

Covers:
- `paste()` returns independent copies on each call, and `store()` snapshots
  its input so later source mutation does not leak into the buffer
- multi-shape stores preserve order and per-shape copy independence
- `availability_changed` emits only on emptiness transitions (parametrized
  truth table, including a non-empty -> non-empty case that stays silent)

## Test plan
- [x] `uv run pytest tests/unit/_shape_clipboard_test.py -q` (9 passed)
- [x] `uv run pytest tests/unit/ -q` (585+9 passed)
- [x] `make lint` clean (ruff format/check, ty)

### DIFF
--- a/tests/unit/_shape_clipboard_test.py
+++ b/tests/unit/_shape_clipboard_test.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from labelme._shape import Shape
+from labelme._shape_clipboard import ShapeClipboard
+
+
+def _make_polygon() -> Shape:
+    return Shape(
+        shape_type="polygon",
+        points=np.array([(0.0, 0.0), (10.0, 0.0), (10.0, 10.0)], dtype=np.float64),
+        closed=True,
+    )
+
+
+def test_paste_without_store_returns_empty() -> None:
+    assert ShapeClipboard().paste() == []
+
+
+def test_paste_returns_independent_copies_each_call() -> None:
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding `tests/unit/_shape_clipboard_test.py` for `ShapeClipboard`.

- **Correctness:** verified by mutation testing the source — breaking copy-on-paste, copy-on-store, and the `availability_changed` emission logic each independently fails the relevant assertion, so the tests are load-bearing rather than false-green.
- **Conventions:** naming/location (`_shape_clipboard_test.py`, flat in `tests/unit/`) matches siblings; plain `test_` functions, full type annotations including `-> None`. `ruff` and `ty` clean.
- **CI:** all 11 checks green (lint + test across ubuntu/macos/windows × py3.12–3.14).
- **Rebase:** not needed; `main` gained 0 commits since the merge-base, so the green run reflects the merged result.

No follow-ups. Ready for your review and merge.

PR #2304, opened 2026-07-08
test(_config): cover the unknown-key rejection path of load_config

wkentaro:
`_update_dict` raises `Unexpected key in config: <key>` when a config override
carries a key absent from the default config, but no test exercised it. Extend
`test_load_config_rejects_invalid_override` with a top-level and a nested
(recursive) unknown key, each asserting the offending key by name so a
regression that reports the wrong key is caught.

## Test plan
- [x] `uv run pytest tests/unit/_config_test.py` (38 passed)
- [x] `uv run ruff check` / `ruff format --check` / `ty check` on the file

### DIFF
--- a/tests/unit/_config_test.py
+++ b/tests/unit/_config_test.py
@@ -165,8 +165,19 @@ def test_migrate_keep_prev_brightness_contrast(
         ),
         ({"shape_color": "random"}, "Unexpected value for config key 'shape_color'"),
         ({"labels": ["cat", "cat"]}, "Duplicates are detected for config key 'labels'"),
+        ({"not_a_real_key": True}, "Unexpected key in config: not_a_real_key"),
+        (
+            {"shortcuts": {"not_a_real_shortcut": "Ctrl+Z"}},
+            "Unexpected key in config: not_a_real_shortcut",
+        ),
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR extending `test_load_config_rejects_invalid_override` with a top-level and a recursive unknown-key case. Reviewed the diff against `labelme/_config/__init__.py`:

- Top-level: `{"not_a_real_key": True}` hits the `key not in target_dict` guard directly → `Unexpected key in config: not_a_real_key`.
- Nested: `{"shortcuts": {"not_a_real_shortcut": ...}}` recurses (default `shortcuts` is a dict, override value is a dict), then raises `Unexpected key in config: not_a_real_shortcut`. This is the recursive path that had no prior coverage.

Both `match=` patterns are regex-safe. No changelog entry needed (test-only, no user-facing change). CI is green across all 9 test-matrix jobs plus lint. No rebase needed: branch is level with `main` (0 commits behind, no file overlap), merge state `CLEAN`.

`/code-review` (medium) surfaced no meaningful findings; skipped `/review-fix`. Single clean conventional commit, so no recommit reshaping.

PR #2305, opened 2026-07-08
fix(label-file): validate image-level flags on read

wkentaro:
`read_label_file` validated every field of an annotation file strictly except the image-level `flags`, which went through `raw.get("flags") or {}` with no type check. A non-dict `flags` (e.g. `"urgent"`, `[]`, `false`) was silently accepted into `Annotation.flags` — typed `dict[str, bool]` — and only surfaced later as a confusing crash on `.items()` or on re-save.

This extracts the existing per-shape flags validation into a shared `_validate_flags` helper and applies it at both the per-shape and image-level read sites, so a malformed `flags` now raises `LabelFileReadError` at load time, the same clean error every other malformed field already produced. `None` / absent / `{}` still yield `{}`; valid files are unaffected.

## Test plan
- [x] Added `flags_not_dict` / `flags_falsy_not_dict` / `flags_value_not_bool` cases to `tests/unit/_label_file_test.py` (39 passed)
- [x] `uv run ruff check` + `uv run ty check` clean
- [x] Drove `read_label_file` on real on-disk JSON: malformed flags raise `LabelFileReadError`; valid / empty / absent / null still load `{}`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed `read_label_file` silently accepting a non-dict image-level `flags` value in an annotation file; the top-level flags are now validated as a `dict[str, bool]` on load, exactly as per-shape flags already were, so a malformed file surfaces a clear read error instead of loading a bad value that crashes later ([#2305](https://github.com/wkentaro/labelme/pull/2305))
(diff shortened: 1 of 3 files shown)


wkentaro:

Reviewed at medium effort; no meaningful findings. The change extracts the existing per-shape `flags` validation into a shared `_validate_flags` helper and applies it to the top-level image flags, which previously used `raw.get("flags") or {}` and silently accepted a non-dict value. Behavior stays consistent with the per-shape path (`None`/missing/`{}` -> `{}`; non-dict or non-`str->bool` -> read error).

- Branch is up to date with `main` (0 commits behind), so no rebase was needed.
- CI is green across all matrix jobs (ubuntu/macos/windows x 3.12-3.14, lint, cla).
- Verified locally: `pytest tests/unit/_label_file_test.py` -> 39 passed; a runtime smoke confirms a malformed `flags` value now raises `LabelFileReadError` while a valid `dict[str, bool]` loads.

No follow-up or idea issues filed.

PR #2307, opened 2026-07-08
fix(shape): ignore linestrip's unrendered closing edge in hit-testing

wkentaro:
`nearest_edge_index` builds its edge candidates with `np.roll(points, 1)`, which always adds a wrap-around segment from the last point back to the first (at index 0). That segment is the closing edge of a polygon, but a linestrip is an open polyline and the renderer never draws it (`_shape_render.py` special-cases linestrip as open). As a result, hovering or clicking near that invisible diagonal was treated as a hit: edge hover highlighted it, add-point-to-edge inserted a point on it, and body selection grabbed the shape.

The fix masks that one phantom segment to `inf` for linestrip shapes only. Polygons keep their real closing edge, and `line` shapes are unaffected (their two roll-segments are the same physical edge). The guard keys off `shape_type` rather than `Shape.closed` on purpose: shapes loaded from an annotation file are set `closed=True` unconditionally, including linestrips, so `closed` is not a reliable "renders a closing edge" signal.

## Test plan
- [x] Added `test_nearest_edge_index_ignores_phantom_closing_edge_for_linestrip` (fails before the fix: returns edge `0`) and `test_nearest_edge_index_matches_drawn_edge_for_linestrip`
- [x] `uv run pytest tests/unit` (610 passed)
- [x] `uv run ruff check` / `uv run ty check labelme/_shape.py`
- [x] Drove `find_hover_target` end-to-end on a loaded (`closed=True`) linestrip: the phantom point yields no hit while real edge and vertex still register

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed clicking or hovering near a linestrip's unrendered "closing" line (the straight path from its last point back to its first) being treated as a hit on the shape; edge hover, add-point-to-edge, and body selection now ignore that phantom segment, since a linestrip is an open polyline and never draws it ([#2307](https://github.com/wkentaro/labelme/pull/2307))
 - Fixed the `--output` guard rejecting only lowercase `.json` paths; an upper- or mixed-case file path such as `--output notes.JSON` slipped past the "expects a directory" check and was treated as an output directory. The guard now reuses the canonical case-insensitive `is_label_file_path` helper ([#2317](https://github.com/wkentaro/labelme/pull/2317))
 
 ## [7.0.4] - 2026-07-12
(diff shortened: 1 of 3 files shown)


wkentaro:

Reviewed the diff (medium `/code-review` + two focused reviewers tracing callers and rendering).

- **Correct:** `starts = np.roll(points, 1)` makes `segments[0]` the last->first wrap-around edge. For a linestrip (open polyline) that segment is never rendered, so `distances[0] = np.inf` correctly excludes it. Verified against `_shape_render.py` (`_build_shape_points_paths` draws only edges `1..N-1` for linestrip) and against all three consumers of `nearest_edge_index` (edge-hover, `add_point_to_edge`, `is_hit_by_point`).
- **No regressions:** polygon still treats its closing edge as hittable (existing parametrization guards this); the `len(points)==1` path is a safe in-bounds write.
- **Tests:** two new unit tests cover the phantom-edge and drawn-edge cases; full `_shape` suite green locally (52 passed).
- **CI:** all 11 checks green, merge state CLEAN.

No push made this tick (no rebase needed: main gained 0 commits since the merge-base; branch already a single clean atomic commit).

follow-ups: #2308 (e2e canvas coverage for 3+ point linestrip hit-testing; non-blocking)

wkentaro:
lg

PR #2310, opened 2026-07-08
refactor(cli): extract shared list-arg parser for --flags/--labels

wkentaro:
The `--flags` and `--labels` options each ran the same file-or-comma parsing block (read+strip lines if the value is a file, else comma-split and drop empties). Collapse both into a `_parse_list_arg` helper so the logic lives in one place. Pure refactor, no behavior change: the `--label-flags` block is left inline since it parses YAML, not a plain list.

The parsing previously only executed inside `main()` (which needs a full Qt app), so it had no direct coverage; the helper is now unit-tested for both the comma-split and file-read-and-strip branches.

## Test plan
- [x] `uv run pytest tests/unit/__main___test.py` (17 passed) and full suite green
- [x] `uv run ruff format --check`, `ruff check`, `ty check` all clean

### DIFF
--- a/labelme/__main__.py
+++ b/labelme/__main__.py
@@ -169,6 +169,13 @@ class _DeprecatedAlias(argparse.Action):
         setattr(namespace, self.dest, self.const if self.nargs == 0 else values)
 
 
+def _parse_list_arg(value: str) -> list[str]:
+    if Path(value).is_file():
+        with open(value, encoding="utf-8") as f:
+            return [line.strip() for line in f if line.strip()]
+    return [line for line in value.split(",") if line]
+
+
 def main() -> None:
     parser = argparse.ArgumentParser()
     parser.add_argument("--version", "-V", action="store_true", help="show version")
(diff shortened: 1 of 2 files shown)


wkentaro:

Clean, behavior-preserving refactor. `_parse_list_arg` is byte-for-byte the same logic as the two inlined `--flags`/`--labels` blocks it replaces:
- file branch: `.strip()` each line, drop blanks
- comma branch: `split(",")`, drop empties

`--label_flags` is correctly left inline since it uses `_yaml.safe_load` (genuinely different logic), not list-splitting.

Edge cases are covered by the new parametrized tests, including `"" -> []` (since `Path("").is_file()` is `False`, falling through to the split branch).

- `/code-review` (medium): no meaningful findings.
- No rebase needed: main gained 0 commits since the merge-base, merge state `CLEAN`.
- CI: green across all 9 platform/version combos.
- Not user-facing, so no CHANGELOG entry needed.

PR #2312, opened 2026-07-09
test(shape): cover rotate()'s source_points snapshot path

wkentaro:
`_shape.rotate()` takes an optional `source_points` snapshot so an interactive rotation drag composes each step from the original points instead of accumulating drift on the already-rotated `shape.points`. Its only production caller (`canvas.py` `_drag_hovered_rotation_point`) always passes `source_points`, yet every existing test in `_shape_test.py` used the `source_points=None` default, leaving the real code path uncovered.

This adds two unit tests:
- `test_rotate_uses_source_points_snapshot_not_current_points` — seeds `shape.points` with a decoy and asserts the result is computed from `source_points`, pinning the single-call precedence.
- `test_rotate_reports_source_points_length_in_error` — asserts a wrong-length `source_points` raises `ValueError` and the message attributes the count to `source_points`.

## Test plan
- [x] `uv run pytest tests/unit/_shape_test.py` (52 passed)
- [x] `uv run pytest tests/unit` (602 passed)
- [x] `uv run ruff check` / `ruff format --check` / `ty check` clean on the touched file

### DIFF
--- a/tests/unit/_shape_test.py
+++ b/tests/unit/_shape_test.py
@@ -96,6 +96,41 @@ def test_rotate_non_oriented_rectangle_raises() -> None:
         )
 
 
+def test_rotate_uses_source_points_snapshot_not_current_points() -> None:
+    # Given source_points, rotate() computes from it and ignores the shape's
+    # current points. This single-call precedence is what lets an interactive
+    # drag replay from a fixed snapshot; it is the only path production uses
+    # (canvas.py always passes source_points).
+    source_points = np.array(
+        [(0.0, 0.0), (10.0, 0.0), (10.0, 4.0), (0.0, 4.0)], dtype=np.float64
+    )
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding two tests for `_shape.rotate()`:

- `test_rotate_uses_source_points_snapshot_not_current_points` — verifies the `source_points`-takes-precedence path (a π/2 rotation of `[(0,0),(10,0),(10,4),(0,4)]` yields `[(0,0),(0,10),(-4,10),(-4,0)]`, ignoring the shape's current points).
- `test_rotate_reports_source_points_length_in_error` — verifies the length-mismatch `ValueError` reports `len(source_points)=3`.

Both assertions verified against the source; all 52 tests in the file pass. Merge state CLEAN, up to date with `main` (0 commits behind base, no rebase needed). All CI green (lint + test across macOS/Ubuntu/Windows × Python 3.12–3.14). `/code-review` at medium effort found no meaningful issues; no push this tick, so the existing green is authoritative. No CHANGELOG entry needed (not user-facing).

PR #2313, opened 2026-07-09
test(_utils/qt): cover zero-length-line guard in point projections

wkentaro:
`project_point_on_line` and `project_point_on_perpendicular_line` guard the degenerate zero-length line (`length_sq == 0.0`) by returning the point unchanged instead of dividing by the zero segment length, but that branch was unexercised. Removing either guard raises `ZeroDivisionError`, so this locks the guarded behavior down with one test per helper.

## Test plan
- [x] `uv run pytest tests/unit/utils/qt_test.py -q` (47 passed)
- [x] Mutation check: deleting either guard makes the new tests fail with `ZeroDivisionError`; restoring it passes
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run ty check`

### DIFF
--- a/tests/unit/utils/qt_test.py
+++ b/tests/unit/utils/qt_test.py
@@ -50,6 +50,16 @@ def test_project_point_on_perpendicular_line(
     assert (projected.x(), projected.y()) == pytest.approx(expected)
 
 
+def test_project_point_on_perpendicular_line_zero_length_returns_point() -> None:
+    # A zero-length line has no direction, so the perpendicular is undefined and
+    # the point is returned unchanged instead of dividing by the zero length.
+    point = QPointF(4.0, 7.0)
+    projected = project_point_on_perpendicular_line(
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding two unit tests for the zero-length-line guard in `project_point_on_line` and `project_point_on_perpendicular_line`.

- **Correct:** both tests pass `line_start == line_end == (2, 2)`, so `length_sq == 0.0` and each helper hits its `return QtCore.QPointF(point)` guard (`labelme/_utils/qt.py:222`, `:241`), returning the input point `(4.0, 7.0)` unchanged. Assertions match the guarded behavior; deleting either guard raises `ZeroDivisionError`.
- **Conventions:** naming, structure, keyword-only call sites, and placement all match the existing `test_project_point_on_line` / `test_project_point_on_perpendicular_line` tests.
- **CI:** all 11 checks green, merge state CLEAN.

No push made this tick (no rebase needed: main gained 0 commits since the merge-base; branch is already a single clean atomic commit).

PR #2315, opened 2026-07-09
refactor(app): extract shared supported-image-extensions helper

wkentaro:
`dragEnterEvent`, `import_dropped_image_files`, and `_scan_image_files` each
built the same lowercased image-extension tuple with an identical
`QImageReader.supportedImageFormats()` comprehension. Collapse the three copies
into one `_list_supported_image_extensions()` helper returning a tuple, so the
two list-based call sites also drop their redundant `tuple(...)` wrapping at the
`endswith()` check. Behavior-preserving; the two nearby `*.ext` display-string
builders are a different shape and left untouched.

## Test plan
- [x] `uv run pytest` (754 passed), `ruff`, `ty` all green
- [x] Directory-scan smoke: `.png` + case-insensitive `.JPG` matched, `.txt` excluded

### DIFF
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -2266,13 +2266,10 @@ class MainWindow(QtWidgets.QMainWindow):
         self._window_state.setValue("window/state", self.saveState())
 
     def dragEnterEvent(self, a0: QtGui.QDragEnterEvent) -> None:
-        extensions = [
-            f".{fmt.toStdString().lower()}"
-            for fmt in QtGui.QImageReader.supportedImageFormats()
-        ]
+        extensions = _list_supported_image_extensions()
         if a0.mimeData().hasUrls():
             items = [i.toLocalFile() for i in a0.mimeData().urls()]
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: recommend-merge**

Pure internal refactor: extracts `_list_supported_image_extensions() -> tuple[str, ...]` and replaces three duplicated inline comprehensions (`dragEnterEvent`, `import_dropped_image_files`, `_scan_image_files`) in `labelme/_app.py`.

- Behavior preserved at all three call sites: each previously produced a tuple (directly or via `tuple(...)` at the `.endswith` call); all now receive the same tuple from the helper. No semantic change.
- Single clean commit; no reshape needed.
- Pure refactor, no runtime behavior change, so no smoke verification.
- `ruff` and `ty` clean; all 11 CI checks green (lint + test across macOS/Ubuntu/Windows × Python 3.12/3.13/3.14).

No changelog entry (not user-facing).

PR #2316, opened 2026-07-09
test(shapes-to-label): cover mask-shape compositing and instance grouping

wkentaro:
Adds focused coverage for `labelme._utils.shape.shapes_to_label`, whose
`shape_type == "mask"` branch and instance-grouping logic were previously
unexercised. This function is a load-bearing conversion path shipped in the
example/export scripts (`examples/instance_segmentation/labelme2voc.py`,
`examples/tutorial/*`), so a silent regression there corrupts exported label
maps.

The three new tests cover:
- mask-shape bbox compositing: the local boolean patch is placed into the
  full-image `cls`/`ins` maps at the given bbox, with inclusive upper bounds
  (the assertions probe patch-content fidelity, a row/col swap, and one-past
  each bound).
- the `ValueError` guard when a mask shape carries no ndarray.
- instance-id grouping by `(label, group_id)` — distinct group_ids get
  distinct instance ids; a repeated `(label, group_id)` reuses its id.

Test-only change; no production code touched.

## Test plan
- [x] `uv run pytest tests/unit/utils/shape_test.py` (17 passed)
- [x] `uv run pytest tests/` (757 passed)
- [x] `uv run ruff format --check` / `ruff check` / `ty check` clean

### DIFF
--- a/tests/unit/utils/shape_test.py
+++ b/tests/unit/utils/shape_test.py
@@ -37,6 +37,73 @@ def test_shapes_to_label_raises_clear_error_for_unknown_label() -> None:
         shape_module.shapes_to_label((20, 20), [shape], {"road": 1})
 
 
+def test_shapes_to_label_places_mask_shape_at_its_bbox() -> None:
+    # A "mask" shape carries a local boolean patch plus a bbox (2 points, xy).
+    # The patch is composited into the full-image label maps at that bbox, with
+    # the bbox upper bound inclusive. mask is indexed [row=y, col=x].
+    patch = np.ones((3, 7), dtype=bool)
+    patch[0, 0] = False  # cleared cell proves the patch content drives the fill
+    shape = ShapeDict(
+        label="car",
+        points=[[2.0, 3.0], [8.0, 5.0]],  # x in [2, 8], y in [3, 5]
+        shape_type="mask",
+        flags={},
+        description="",
+        group_id=None,
+        mask=patch,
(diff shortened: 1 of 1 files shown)


wkentaro:

Reviewed the diff (test-only; medium `/code-review` gate + a source cross-check).

- **Correct:** all three tests match `shapes_to_label`’s real behavior in `labelme/_utils/shape.py`:
  - mask compositing: `mask[y1:y2+1, x1:x2+1] = shape["mask"]` with bbox `(x1,y1),(x2,y2)` cast to int, inclusive upper bound, `[row=y, col=x]` indexing. The `patch[0,0]=False` / bottom-right-corner / one-past-bound assertions pin down content-driven fill, an inclusive boundary, and row/col orientation (a swap would fail `cls[5,8]`).
  - ndarray guard: `raises(ValueError, "shape\[.mask.\] must be numpy.ndarray")` matches the `isinstance` check.
  - instance grouping: `(label, group_id)` tuple → `ins_id`; the 1/2/1 group_ids correctly yield instances 1/2/1.
- **Non-tautological:** assertions derive from the source contract, not from re-running it.
- **CI:** all 11 checks green (9-job OS×Python matrix + lint + cla), merge state CLEAN.

No push this tick (branch is a single clean atomic commit; main gained 0 commits since the merge-base, so no rebase). Test-only, so no `/verify` smoke.

No follow-ups.

PR #2318, opened 2026-07-10
fix: crash adding a shape after context-menu deletion in label list

wkentaro:
Fixes a reproducible crash reported on the macOS 7.0.2 build: register a rectangle, delete it via the annotation-list context menu, then register another shape with the same label. The app dies in `_on_label_item_changed` with `RuntimeError: libshiboken: Internal C++ object (LabelListWidgetItem) already deleted`.

`LabelListWidget` snapshots the selected items on mouse press (to propagate checkbox visibility toggles across a multi-selection) and clears the snapshot on mouse release. A context menu or drag consumes the release (on macOS the context menu opens on press), so the snapshot outlives a deletion performed from that menu; the next `itemChanged` then touches a dead item wrapper. The snapshot now stores `QPersistentModelIndex`, Qt's handle for keeping a model reference across structural changes, and resolves to live items at use time, dropping indexes invalidated by row removal.

The snapshot code predates v7 but did not crash under PyQt5, whose comparison against a deleted wrapper returns `False` instead of raising; the PySide6 migration turned the stale reference into a crash, matching the report that 6.x was unaffected.

## Test plan
- [x] New regression tests reproduce the reported `RuntimeError` on main and pass here (`tests/unit/widgets/label_list_widget_test.py`)
- [x] The multi-selection checkbox-toggle restore behavior the snapshot exists for is pinned by a new test (verified it fails if the restore branch is disabled)
- [x] `uv run pytest` (758 passed)


wkentaro:

Reviewed the fix for the label-list stale-snapshot crash (`RuntimeError: Internal C++ object (LabelListWidgetItem) already deleted`).

- **Root cause & fix are sound.** The press-time snapshot stored raw `LabelListWidgetItem` wrappers, which the model deletes on row removal, so a context menu / drag consuming the release left dangling wrappers. Storing `QPersistentModelIndex` and resolving lazily through `_resolve_item` (dropping invalidated rows) is the right depth for the fix, not a bandaid.
- **`QPersistentModelIndex` semantics verified.** Confirmed a persistent index goes permanently invalid on row removal and is not reassigned to a later item added at the same row, so `test_selection_at_press_drops_items_removed_before_release` holds by Qt guarantee, not coincidence.
- **No regression in the multi-selection-restore path.** The rewritten `mouseReleaseEvent` loop generalizes the old `any(...)`/tuple comprehensions; skipping deleted entries is the intended behavior. Sole caller of `selection_at_press` (`_app.py`) is unaffected by `None` filtering.
- **Tests drive the real widget** via `qtbot` across the deletion-before-release, live-multi-selection, and checkbox-toggle-restore paths.
- **CI green** across ubuntu/macos/windows × py3.12-3.14; branch sits directly on current `main` (no rebase needed).

No follow-ups or spin-off issues.

PR #2319, opened 2026-07-10
test(app): cover _app.py's pure module-level helpers

wkentaro:
`labelme/_app.py` is the largest module in the package but had no unit test file.
This adds `tests/unit/_app_test.py` pinning the behavior of four pure module-level
helpers and their branches, so a future refactor that narrows a guard or swaps a
branch gets caught by a cheap test rather than in the running app.

## Summary
- `_resolve_text_annotation_shape_type`: AI mode passthrough, text-to-annotation mode, and unrelated-mode fall-through to `None`.
- `_rgb_from_label_colors`: `None`/empty/missing-label to `None`, valid RGB (including the `0` boundary), and the `ValueError` on malformed RGB (wrong length or out-of-range channel), asserted on the message so the explicit length check is what fires.
- `_is_valid_label`: `None` policy, `exact` match/no-match, and unknown-policy fall-through.
- `_format_window_title`: full decision table over image path / index / count / dirty, including `file_index=0` (the `is not None` vs truthy distinction) and the `file_count`-set / index-unset combinations.

## Test plan
- [x] `uv run pytest tests/unit/_app_test.py -q` (27 cases pass)
- [x] full suite green: `uv run pytest tests/` (777 passed)
- [x] `uv run ruff format --check`, `uv run ruff check`, `uv run ty check` clean on the new file

### DIFF
--- a/tests/unit/_app_test.py
+++ b/tests/unit/_app_test.py
@@ -1,0 +1,143 @@
+from __future__ import annotations
+
+import pytest
+
+from labelme import __appname__
+from labelme import _app
+from labelme import _automation
+
+
+@pytest.mark.parametrize(
+    "create_mode, ai_output_format, expected",
+    [
+        ("ai_points_to_shape", "mask", "mask"),
+        ("ai_box_to_shape", "polygon", "polygon"),
+        ("polygon", "mask", "polygon"),
+        ("rectangle", "mask", "rectangle"),
+        ("edit", "polygon", None),
+    ],
+    ids=[
+        "ai-points-passthrough",
+        "ai-box-passthrough",
+        "text-polygon",
+        "text-rectangle",
+        "unrelated-mode",
+    ],
+)
+def test_resolve_text_annotation_shape_type(
+    create_mode: str,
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding `tests/unit/_app_test.py`, which pins the behavior of four pure module-level helpers in `_app.py` (`_resolve_text_annotation_shape_type`, `_rgb_from_label_colors`, `_is_valid_label`, `_format_window_title`).

- Reviewed the parametrized cases against the current implementations (`_app.py:2848-2899`); every assertion matches actual behavior, including the boundary cases the description calls out: the `0` RGB channel, the `file_index=0` `is not None`-vs-truthy distinction, and the `file_count`-set / index-unset combination.
- Branch is up to date with `main` (merge-base is `main` HEAD); no rebase needed.
- All 11 CI checks green (lint + test across ubuntu/macos/windows on 3.12–3.14).
- Single clean commit; no reshaping needed.

No changes pushed this tick. Recommending merge; the final call is yours.

PR #2321, opened 2026-07-10
refactor(_utils): drop unused img_pil_to_data re-export

wkentaro:
`img_pil_to_data` has no consumer via `labelme._utils`: it is only called
internally by `img_arr_to_data` in `labelme/_utils/image.py`. This drops the
orphaned package re-export left behind after the public-API-surface removal,
keeping the function itself since it is still used internally.

Same cleanup class as the earlier drop of the `new_button` / `distance` /
`distance_to_line` / `img_data_to_png_data` re-exports (20604696); this sibling
was missed because the function still has an internal caller, so a naive
"no caller" search over the function name did not flag it.

## Test plan
- [x] `pytest tests/` (754 passed)
- [x] `ruff check labelme/_utils/__init__.py`, `ty check` (clean)

### DIFF
--- a/labelme/_utils/__init__.py
+++ b/labelme/_utils/__init__.py
@@ -4,7 +4,6 @@ from .image import img_arr_to_data
 from .image import img_b64_to_arr
 from .image import img_data_to_arr
 from .image import img_data_to_pil
-from .image import img_pil_to_data
 from .image import img_qt_to_arr
 from .qt import add_actions
 from .qt import apply_color_theme


wkentaro:

One-line removal of the orphaned `img_pil_to_data` re-export from `labelme/_utils/__init__.py`. Verified:

- `img_pil_to_data` is defined in `_utils/image.py:33` and called only internally by `img_arr_to_data` (`image.py:48`); the removed line was its sole re-export and it has no consumer via `labelme._utils` (grep confirms zero external references).
- No `__all__` in the package, so dropping the import breaks nothing.
- The function itself is correctly retained (still used internally).

No rebase needed (branch is even with `main`; merge state CLEAN, 0 commits diverged). CI is fully green across all 11 matrix jobs plus lint/cla. `/code-review` found no meaningful issues, so `/review-fix` was skipped; the single commit is already clean.

Audited the remaining 19 re-exports in the same file: every one has at least one external consumer, so no sibling orphans remain to clean up.

PR #2323, opened 2026-07-10
fix: reject boolean group_id and point coordinates on read

wkentaro:
`_load_shape_json_obj` validates a shape's `group_id` as `int` and its point coordinates as `int | float`, but `bool` is a subclass of `int` in Python, so `isinstance(True, int)` is `True`. A hand-edited or third-party annotation file with `"group_id": true` (or a `true`/`false` inside `points`) slipped past both checks and loaded `True`/`False` as a group identifier or coordinate instead of raising the intended read error. This mirrors the recently fixed `flags` validation gap (#2305).

Both sites now exclude `bool` explicitly, keeping the same inline `isinstance` style as the sibling `label` / `shape_type` / `description` validations. labelme's own write path never produces a boolean here (`group_id` comes from `int(...)`, coordinates are coerced to `float`), so this only tightens rejection of malformed input, no valid file is affected.

Not in scope: the top-level `imageHeight` / `imageWidth` fields have no type check at all (a different gap from this bool-as-int loophole) and are left for a possible follow-up.

## Test plan
- [x] Added `points_bool_coord` and `group_id_bool` cases to the parametrized malformed-shape test in `tests/unit/_label_file_test.py`
- [x] `uv run pytest tests/unit/` (602 passed)
- [x] `ruff format --check`, `ruff check`, `ty check` clean


wkentaro:

**Verdict: recommend-merge**

Small, correctly-scoped fix. Since `bool` is a subclass of `int`, a hand-edited or third-party `"group_id": true` (or a `true`/`false` inside `points`) previously slipped past the `int`/`int | float` checks and loaded as a group identifier or coordinate. Both validation sites now reject `bool` explicitly, surfacing the same clear read error as any other bad value.

- Review gate (medium `/code-review`, report-only): no meaningful findings. `group_id` is the only `int`-typed field and points the only numeric field, so nothing analogous is left unguarded; image/shape `flags` are validated separately as `dict[str, bool]`.
- Tests: two parametrized cases added (`points_bool_coord`, `group_id_bool`), both asserting the read error.
- CI: all checks green (lint, cla, test matrix across ubuntu/macos/windows × 3.12/3.13/3.14).
- No rebase needed (branch is even with `main`); no push this tick, so the existing green is authoritative.

Endorsed for merge.

wkentaro:

wkentaro:

Rebased onto current `main` and force-pushed. The branch was conflicting (`DIRTY`) against `main` on `CHANGELOG.md`, so the existing `recommend-merge` endorsement was not actionable; the rebase resolves that. The change itself is unmodified (the commit is content-identical to the previous head, just replayed onto `bd3ca1d`).

Re-verified after the rebase:

- `uv run pytest tests/unit/` — 738 passed
- `ruff check` / `ruff format --check` — clean
- `/code-review` (Standards + Spec axes) — no findings on either axis. The only note was a style nit: the two bool-exclusion checks are De Morgan duals phrased differently, but each follows its own site's pre-existing idiom (positive predicate inside `all()` vs. negative guard clause matching the sibling `label` / `shape_type` / `description` checks), and factoring out a shared two-call-site helper would cut against the repo's preference for inline `isinstance` validation here. Left as-is.

The spec claim that no valid file is affected was verified against the write path: `group_id` comes from `int(gid_text)` (`labelme/_widgets/label_dialog.py:302`) and coordinates from `shape.points.tolist()` on a `float64` array (`labelme/_app.py:2919`), neither of which can emit a boolean. No other bool-as-int loophole remains in the shape-loading path.

Verdict `recommend-merge` stands, pending the re-triggered CI.

PR #2326, opened 2026-07-11
test(_yaml): cover safe_load parsing, streams, and unsafe-tag rejection

wkentaro:
`labelme/_yaml.py`'s `safe_load` had no unit test of its own — it was only exercised incidentally through config-loading paths, so its contract was unpinned. It's the sole YAML entry point behind `load_config` and the `--config` CLI value, and it uses `typ="safe"` specifically to reject dangerous `!!python/...` tags in user-editable config files.

This adds a test file pinning that contract: nested-structure parsing, `IO[str]` stream input, empty-string returning `None`, and rejection of unsafe python tags across four attack vectors. The rejection assertion binds to ruamel's public `YAMLError` base class rather than an internal exception subtype.

## Test plan
- [x] `uv run pytest tests/unit/_yaml_test.py` (7 cases pass)
- [x] `uv run ruff check` / `uv run ty check` clean

### DIFF
--- a/tests/unit/_yaml_test.py
+++ b/tests/unit/_yaml_test.py
@@ -1,0 +1,34 @@
+from __future__ import annotations
+
+import io
+
+import pytest
+from ruamel.yaml import YAMLError
+
+from labelme import _yaml
+
+
+def test_safe_load_parses_string_into_nested_structure() -> None:
+    assert _yaml.safe_load("a: 1\nb:\n  - 2\n  - 3\n") == {"a": 1, "b": [2, 3]}
+
+
+def test_safe_load_accepts_file_like_stream() -> None:
+    assert _yaml.safe_load(io.StringIO("x: hello\n")) == {"x": "hello"}
+
+
+def test_safe_load_returns_none_for_empty_string() -> None:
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: recommend-merge**

Test-only PR adding `tests/unit/_yaml_test.py` (+34/-0) for `labelme._yaml.safe_load`. Coverage is well-targeted and matches the module contract:

- string → nested structure, file-like stream input, empty string → `None`
- parametrized rejection of unsafe `!!python/*` tags (`apply`/`new`/`name`/`module`), asserting `YAMLError` — locking in the safe-loader guarantee the module comment exists to protect.

Follows repo test conventions (plain `test_` functions, `-> None` annotations, source-mirrored layout). No rebase needed: main advanced only in files disjoint from this PR (`CHANGELOG.md`, `tests/unit/_config_test.py`) and by ≤20 commits, so the existing green reflects the merged result. `/code-review` at medium surfaced no meaningful findings. CI green across ubuntu/macos/windows × py3.12–3.14 plus lint. Mergeable, clean.

PR #2327, opened 2026-07-11
fix(shapes-to-label): clip out-of-bounds mask bbox to the image canvas

wkentaro:
`shapes_to_label` rasterizes a `mask` shape by writing its patch into a raw numpy
slice of the bbox: `mask[y1:y2+1, x1:x2+1] = shape["mask"]`. When the bbox extends
past the image edge — reachable now that `canvas.allow_out_of_bounds_points` lets
points sit outside the Image — this misbehaves two ways:

- a negative corner is interpreted by numpy as an index from the end, so an
  off-left/top bbox silently wraps its pixels onto the opposite edge;
- a right/bottom overflow raises an opaque `ValueError` broadcast error.

ADR 0004 already states rasterization must clip out-of-bounds coordinates to the
image canvas (as the polygon path does via PIL). This clips the destination
rectangle and the matching sub-slice of the mask to `[0, height) x [0, width)`
for the `mask` branch too. For in-bounds bboxes the slice is unchanged, so
well-formed data is unaffected.

The identical fix is applied to the vendored `examples/utils.py` copy, which ADR
0004 names explicitly (labelme2voc / labelme2coco).

## Test plan
- [x] `tests/unit/utils/shape_test.py`: added in-bounds paint, off-left clip, and
      right-overflow clip cases (fail pre-fix, pass post-fix)
- [x] `tests/unit/examples/utils_test.py`: same three cases against the
      `examples/utils.py` copy (that copy had no test before)
- [x] `uv run pytest tests/unit` — 612 passed
- [x] `uv run ruff check` / `uv run ty check` on touched files — clean


wkentaro:

Finalize pass: CI's `lint` was red because `typos` flagged `mis-rasterizing` in the changelog (`mis` -> `miss`/`mist`). Reworded to `mishandling`.

While reviewing, a `/code-review` pass found the source clip logic correct, but the test suite never exercised the fix's headline path: a bbox that *partially* straddles a negative corner (so the source sub-slice start-offset `y_start - y1` / `x_start - x1` is nonzero). The existing "off left edge" case actually short-circuits the fully-out-of-bounds branch, and every patch was a uniform `np.ones`, so a wrong offset couldn't be detected. Added `test_shapes_to_label_mask_clips_bbox_off_top_left_corner` (asymmetric marks) to both `shape_test.py` and `examples/utils_test.py`.

All tests + lint green locally. Force-pushed; awaiting CI before a verdict.

wkentaro:

Finalize pass:
- **Rebase:** skipped — merge state `CLEAN`, `main` gained 0 commits since the merge-base, so the existing green reflects the merged result.
- **Review:** report-only `/code-review` at high effort. The clipping arithmetic was checked symbolically (painted rows/cols always equal the sliced patch rows/cols, so no broadcast mismatch on negative, right/bottom, or corner-straddle overflow) and the load-bearing invariant `shape["mask"].shape == (y2-y1+1, x2-x1+1)` was traced to its source (osam pydantic validator at detection; mask shapes only translate, never per-vertex drag, preserving bbox size). No findings → `/review-fix` skipped.
- **Commits:** already a single well-formed commit; `/recommit` a no-op.
- **Tests/CI:** 22 pass locally; all 11 CI checks green.

Sound, focused, ADR-0004-aligned fix. Recommending merge.

wkentaro:

Rebased onto current \`main\` (was \`DIRTY\`, ~60 commits behind). Conflicts were confined to `CHANGELOG.md` and `tests/unit/utils/shape_test.py`, both keep-both merges of independent additions; the source fix in \`labelme/_utils/shape.py\` auto-merged unchanged. `make lint` (ruff/ty/mdformat/typos) is clean and the affected unit tests (`tests/unit/utils/shape_test.py`, `tests/unit/examples/utils_test.py`) pass locally (21 passed).

Removed the stale `recommend-merge` and left this awaiting the fresh CI run against the new base; a later tick re-verdicts once checks are green.

wkentaro:

Finalize pass (re-verdict after this morning's clean rebase):

- **Rebase:** skipped. The merge-base already equals current `main` HEAD (`main` gained 0 commits since base), branch is `CLEAN`/`MERGEABLE`, so no push this tick and the existing green CI is authoritative.
- **Review:** cheap `/code-review` gate at medium effort found no meaningful findings. The clip is index-consistent — the destination rectangle `[y_start:y_stop, x_start:x_stop]` and the mask sub-slice `shape["mask"][y_start-y1:y_stop-y1, x_start-x1:x_stop-x1]` share identical extents, and clamping keeps source indices within the mask. So `/review-fix` was skipped.
- **Recommit:** single well-formed commit; no reshaping needed.
- **CI:** green across all 9 platform/version combos (ubuntu/macos/windows × 3.12/3.13/3.14) plus lint.

Sound, targeted out-of-bounds-mask-bbox clip per ADR 0004, well covered by the added tests. Ready for maintainer merge.

wkentaro:

Rebased onto current `main` — the branch had gone conflicting (`DIRTY`), so the previous `recommend-merge` was no longer actionable.

**Conflict resolved:** `CHANGELOG.md` only, and purely additive — `main` had appended several `### Fixed` entries under `[Unreleased]` while this branch appended its own. Kept both, with this PR's entry last. No source conflict; the two `shapes_to_label` hunks applied unchanged.

**Re-checks after the rebase:**
- `uv run pytest tests/unit` — 744 passed
- `uv run ruff check .` / `ruff format --check .` — clean
- Two-axis review (repo standards + fidelity to ADR 0004) — no findings on either axis. The only notes were judgement-call duplication smells (the fix and its tests appear twice), and both are correct here: `shapes_to_label` was already duplicated between `labelme/_utils/shape.py` and the vendored `examples/utils.py` before this change, and ADR 0004 names that copy explicitly, so applying the fix to both is required rather than avoidable. Deduplicating them is a separate refactor, out of scope.
- Index arithmetic re-verified against five cases: fully in-bounds (slice reduces byte-for-byte to the pre-fix expression, so well-formed data is untouched), off-left/top, right/bottom overflow, entirely-outside, and a patch whose dimensions disagree with the bbox (still raises `ValueError` — fails loudly rather than corrupting silently).

Verdict label cleared while CI runs against the rebased head; the next processing tick reads the result and re-verdicts.

wkentaro:

CI is green on the rebased head (`ce6776d3`) — all 9 test jobs across ubuntu/macos/windows on Python 3.12/3.13/3.14, plus lint, passed on the first attempt. Merge state is `CLEAN`.

Restoring `recommend-merge`.

follow-ups: #2419 (the two `shapes_to_label` / `shape_to_mask` copies are hand-synced; does not block this PR)

PR #2328, opened 2026-07-11
test(_utils/image): cover img_qt_to_arr pixel layout and scanline padding

wkentaro:
`img_qt_to_arr` (`labelme/_utils/image.py`) converts a `QImage` to a numpy array on the hot path (`_app.py`, `canvas.py`) but had zero test coverage. Its two subtle behaviors are easy to regress: the byte layout of `Format_RGB32` (BGRA on little-endian) and Qt's 4-byte scanline padding, which must be stripped so a padded row never leaks garbage columns into the array.

Three characterization tests pin these down:
- `test_img_qt_to_arr_returns_bgra_pixels` — asserts the BGRA-with-filled-alpha layout at two corners.
- `test_img_qt_to_arr_strips_scanline_padding` — uses a width where `bytesPerLine()` (12) exceeds `width * channels` (9) to exercise the padding-strip branch.
- `test_img_qt_to_arr_copies_source_buffer` — mutates the source `QImage` after conversion and asserts the array is unchanged, locking in that the result owns its data (guards a future zero-copy refactor from reintroducing dangling-buffer UB).

## Test plan
- [x] `QT_QPA_PLATFORM=offscreen uv run pytest tests/unit/utils/image_test.py -q` (15 passed)
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run ty check` on the file

### DIFF
--- a/tests/unit/utils/image_test.py
+++ b/tests/unit/utils/image_test.py
@@ -6,6 +6,7 @@ from typing import Final
 import numpy as np
 import PIL.Image
 import pytest
+from PySide6 import QtGui
 
 from labelme._utils import image as image_module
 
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: recommend-merge**

Test-only PR adding three targeted unit tests for `img_qt_to_arr`. Reviewed the diff against the implementation and all three assertions are correct:

- **BGRA layout** — `Format_RGB32` is stored BGRA on little-endian, so `QColor(10,20,30)` reads back as `[30,20,10,255]`. ✓
- **Scanline padding** — `Format_RGB888` at width 3 pads each row from 9 to 12 bytes; the test confirms the padding is stripped. ✓
- **Buffer copy** — `bytes(img_qt.bits())` copies, so `arr` survives the source being refilled. ✓

Branch is CLEAN and fully current with `main` (0 divergence), so no rebase. `/code-review` at medium effort found nothing meaningful; single clean commit needs no recommit; test-only so no smoke verify. All 11 CI checks green.

PR #2333, opened 2026-07-12
test(_suppression): cover circle rasterization as inscribed disk

wkentaro:
`_rasterize_shape`'s `circle` branch (`labelme/_automation/_suppression.py`) drew the inscribed disk, but no test exercised it — every existing suppression test used rectangle, polygon, or mask shapes, leaving lines 193-208 uncovered. This adds one test that pins the disk-vs-bounding-square distinction: an existing circle suppresses a detection at its center but keeps one in a bbox corner outside the disk, where a rectangle interpretation of the same bbox would wrongly drop it.

Coverage of `_suppression.py` goes from 90% to 97%.

## Test plan
- [x] `uv run pytest tests/unit/_automation/_suppression_test.py` (26 passed)
- [x] Full unit suite green (609 passed); `ruff check`, `ruff format --check`, `ty check` clean

### DIFF
--- a/tests/unit/_automation/_suppression_test.py
+++ b/tests/unit/_automation/_suppression_test.py
@@ -319,6 +319,26 @@ def test_overlapping_uses_existing_mask_shape_geometry() -> None:
     assert kept == [upper_right]
 
 
+def test_overlapping_rasterizes_existing_circle_as_a_disk() -> None:
+    # A circle shape is rasterized as its inscribed disk, not its bounding
+    # square: a detection at the disk center is suppressed, but one in a bbox
+    # corner (outside the disk) survives, where a rectangle bbox would drop it.
+    existing = Shape(
+        shape_type="circle",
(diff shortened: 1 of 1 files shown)


wkentaro:

Verdict: **recommend-merge**.

Test-only PR adding one case that pins the circle-as-inscribed-disk behavior in `_rasterize_shape`. Verified locally:
- The new test passes (`test_overlapping_rasterizes_existing_circle_as_a_disk`, 1 passed).
- It is a genuine guard: the circle branch draws an `ellipse` (disk), and the corner detection at bbox `(0,0,2,2)` sits outside the radius-10 disk centered at `(10,10)` (distance ≈ 11.3), so it survives — a bounding-square interpretation would wrongly drop it.
- CI is green across all 9 OS/Python matrix legs, plus lint and cla.

Single clean commit, CLEAN merge state, branch already at `main`'s tip (no rebase needed). Nothing to fix; ready for maintainer merge.

PR #2334, opened 2026-07-12
Fix typo in labelme

bglglzd:
I fixed a small set of formatting issues that were easy to clean up without touching behavior.


wkentaro:

Thanks for the contribution. Reviewed this as part of routine PR processing.

The diff only appends a trailing newline to six committed example fixtures (`examples/bbox_detection/data_annotated/*.json`, `examples/classification/data_annotated/0001.json`, `class_names.txt`, `labels.txt`). There is no typo or behavior change here despite the title/description, and these are generated dataset fixtures where a trailing newline carries no value.

There is also a hard blocker: the **CLA check is failing** (`cla` → FAILURE), so this cannot be merged as-is.

Given the change is whitespace-only on example data with no functional benefit, and it is mislabeled as a typo fix, I am recommending this be closed rather than merged. Recording a `recommend-close` verdict for the maintainer to make the final call.

PR #2338, opened 2026-07-13
fix(app): load the real icon-256.png for the window icon

wkentaro:
The application window/taskbar icon was silently blank. `main()` requested it with `new_icon("icon")`, which resolves to `labelme/icons/icon.png` — a symlink that was deleted in b660f95a ("remove symlink as it causes issues on PyInstaller"). `QtGui.QIcon` constructs a null icon for a missing path without raising, so the missing file went unnoticed: that same commit updated the README's PyInstaller build to `icon-256.png` but left this in-app call site pointing at the now-gone `icon.png`.

Point the call at the real `icons/icon-256.png` (the file the README already switched to). The `new_icon` unit tests only asserted `isinstance(..., QIcon)`, never `not isNull()`, which is why the blank icon shipped green — they now assert the icon is non-null so a missing asset fails loudly.

Verified that all 29 icon names referenced across `labelme/` resolve to existing files after this change; `icon` was the only broken one.

## Test plan
- [x] `QT_QPA_PLATFORM=offscreen uv run pytest -q` (762 passed)
- [x] `uv run ruff check` / `uv run ty check` on the changed files
- [x] Behavioral repro: `new_icon("icon").isNull()` is `True` (before) vs `new_icon("icon-256.png").isNull()` is `False`, and `app.setWindowIcon(...)` then exposes a 256×256 pixmap


wkentaro:

**Verdict: recommend-merge**

Reviewed at high effort; no meaningful findings.

- **Correct root cause.** `icons/icon.png` no longer exists (the tree has `icon-256.png`, `icon-128.png`, … and `icon.ico` only), so the previous `new_icon("icon")` resolved to a missing `icon.png` and produced a null `QIcon` — a silent blank window/taskbar icon. Switching to `new_icon("icon-256.png")` loads a real file.
- **Only call site.** `setWindowIcon(new_icon("icon"))` was the sole reference to the removed asset; grep finds no other `icon.png` / `new_icon("icon")` usage.
- **Guards against regression.** The `new_icon` tests now assert `not icon.isNull()`, so a future missing icon file fails loudly instead of silently.
- CHANGELOG updated; single clean commit; branch is current with `main` (0 behind, `CLEAN`); all 11 checks green.

No rebase needed (main disjoint and 0 commits ahead of merge-base), so the existing green CI reflects the merged result.

wkentaro:

PR #2339, opened 2026-07-13
refactor(cli): drop _LoggerIO overrides that restate io.StringIO defaults

wkentaro:
`_LoggerIO` subclasses `io.StringIO` (only to satisfy the `redirect_stderr(new_target=...)` type) and overrode six methods. Three of them — `flush()`, `writable()`, and `closed` — returned exactly what `io.StringIO` already provides by default, so they added nothing. This drops those three and keeps `readable()`/`seekable()`, which genuinely deviate (both default to `True`; the sink is write-only and non-seekable).

The one case where the removed overrides would differ from the base — after `.close()` is called on the instance — is unreachable: `contextlib.redirect_stderr` never closes its target and nothing else references the sink.

Internal-only refactor with no user-facing behavior change, so no CHANGELOG entry. A characterization test locks the resulting stream contract (write forwards stripped content to loguru; `readable`/`seekable` are `False`).

## Test plan
- [x] `pytest tests/unit/__main___test.py` (19 passed, incl. 2 new `_LoggerIO` tests)
- [x] full suite `pytest tests/` (764 passed)
- [x] `ruff format --check`, `ruff check`, `ty check` clean
- [x] smoke-drove the production path: `redirect_stderr(new_target=_LoggerIO())` still forwards stderr writes to loguru with the stream contract intact

### DIFF
--- a/labelme/__main__.py
+++ b/labelme/__main__.py
@@ -35,20 +35,10 @@ class _LoggerIO(io.StringIO):
             logger.debug(stripped_s)
         return len(s)
 
-    def flush(self) -> None:
-        pass
-
-    def writable(self) -> bool:
-        return True
-
     def readable(self) -> bool:
         return False
 
     def seekable(self) -> bool:
-        return False
-
-    @property
-    def closed(self) -> bool:
         return False
 
 
--- a/tests/unit/__main___test.py
+++ b/tests/unit/__main___test.py
@@ -11,6 +11,7 @@ import pytest
 from loguru import logger
 from PySide6 import QtCore
 
+from labelme.__main__ import _LoggerIO
 from labelme.__main__ import _parse_list_arg
 from labelme.__main__ import _route_qt_logging_to_loguru
 from labelme.__main__ import main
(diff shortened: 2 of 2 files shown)


wkentaro:

**Verdict: recommend-merge**

Clean, correct refactor. `flush()`, `writable()`, and `closed` on `_LoggerIO` returned exactly what `io.StringIO` already provides, so dropping them is behavior-preserving; verified the base-class defaults match (`writable=True`, `closed=False`, `flush()` no-op). `readable()`/`seekable()` are correctly kept since `io.StringIO` defaults both to `True` while this write-only stderr sink needs `False`.

The only behavioral nuance, that inherited `closed` reports `True` after `.close()` where the old override forced `False`, is unreachable: the sole instance is a `contextlib.redirect_stderr` target, which never closes its target, and no caller inspects `.closed`.

Single well-described commit, characterization test added locking the stream contract, CI green across all 9 platform/version combos, mergeable and up to date with main. Ready for maintainer merge.

PR #2340, opened 2026-07-13
fix(label-file): re-encode palette images to a Pillow-writable format

wkentaro:
`read_image_file` re-encodes any image whose bytes need transforming (a non-jpg/png extension, or an EXIF-oriented file), and it chose the target format from the mode string alone. Palette images then hit a format Pillow cannot write them in and failed to open (the app catches the `OSError` and shows a "failed to open image" dialog):

- a `P` image (GIF, palette BMP/PNG) went to JPEG → `cannot write mode P as JPEG`
- a `PA` image (palette + alpha, e.g. a palette TIFF) went to PNG → `cannot write mode PA as PNG`

The fix converts a non-transparent `P` image to `RGB` before the JPEG save, routes a `P` image carrying index-based transparency to PNG so its transparency is preserved, and converts a `PA` image to `RGBA` before the PNG save.

## Test plan
- [x] New regression tests in `tests/unit/read_image_file_test.py`: palette GIF/BMP → JPEG, transparent palette GIF → PNG (transparency preserved), palette+alpha TIFF → PNG, bilevel `1` → JPEG without RGB widening. All fail on `main`, pass here.
- [x] `uv run pytest tests/unit` (613 passed), `ruff check`, `ruff format --check`, `ty check` all green.
- [x] Smoke: ran `read_image_file` on real palette files (the exact call `_app.py` makes when opening an image). Before: all four palette cases raise `OSError`. After: all load with the correct format/mode and dimensions preserved; a plain RGB control is unchanged.

Not in scope: non-palette `L`/`RGB` images carrying a color-key `transparency` info still route to JPEG (transparency dropped). This is pre-existing and does not crash, so it is left for a separate change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed clicking or hovering near a linestrip's unrendered "closing" line (the straight path from its last point back to its first) being treated as a hit on the shape; edge hover, add-point-to-edge, and body selection now ignore that phantom segment, since a linestrip is an open polyline and never draws it ([#2307](https://github.com/wkentaro/labelme/pull/2307))
 - Fixed the `--output` guard rejecting only lowercase `.json` paths; an upper- or mixed-case file path such as `--output notes.JSON` slipped past the "expects a directory" check and was treated as an output directory. The guard now reuses the canonical case-insensitive `is_label_file_path` helper ([#2317](https://github.com/wkentaro/labelme/pull/2317))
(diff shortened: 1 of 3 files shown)


wkentaro:

Finalized this tick:

- **No rebase needed** — the branch sits at the current `main` tip (0 commits gained since merge-base, merge state `CLEAN`), so the green CI already reflects the merged result.
- **`/code-review` (medium, report-only):** no meaningful findings. The fix is well-scoped and correct. The `_DISPLAYABLE_MODES` gate in `_imread` bounds reachable modes to `{1, L, P, RGB, RGBA, LA, PA}`, so the two crash paths this addresses (`P`→JPEG, `PA`→PNG) are fixed without altering behavior for any other mode; `I`/`F`/`CMYK`/`I;16` never reach this hunk. Skipped `/review-fix`.
- **Commits:** single clean conventional commit; no recommit needed.
- **Verify:** `tests/unit/read_image_file_test.py` — 13 passed locally; CI green across ubuntu/macos/windows × Python 3.12/3.13/3.14.

wkentaro:

PR #2341, opened 2026-07-13
fix(shape-render): align mask outline with its fill

wkentaro:
The outline of a Mask shape was drawn one pixel down-and-right of its filled pixels. The contour is traced with `skimage.measure.find_contours` on a 1-pixel-padded copy of the mask (so a region touching the mask border still closes into a contour), but the resulting pad offset was never subtracted before positioning the stroke, unlike `compute_polygon_from_mask`, which shares the same pad-then-unpad idiom. Subtracting the pad realigns the outline with the fill it encloses.

## Test plan
- [x] Added `test_mask_outline_aligns_with_fill`: renders a mask through the real paint pipeline and asserts the outline centers on the fill. Verified it fails on the pre-fix code (1px offset) and on an x/y axis swap, and passes with the fix.
- [x] `uv run pytest` (763 passed), `ruff check`, `ruff format --check`, `ty check` all green.


wkentaro:

Finalized this tick. Summary:

- **No rebase needed** — the PR's merge-base is current `main`'s tip (main gained 0 commits since branch point), merge state is `CLEAN`/`MERGEABLE`, so the existing green CI already reflects the merged result.
- **Cheap `/code-review` (medium, report-only)** over the diff: no meaningful findings. The `PAD` offset is correctly subtracted on both axes before the `origin` shift; axis mapping is unchanged and matches how the fill is positioned; the new `@pytest.mark.gui` test derives its expected outline center independently and its scan window safely excludes the bbox stroke. Did not re-propose extracting a shared `find_mask_contours` helper (declined here as cross-module scope creep).
- **`/recommit`**: already a single, well-scoped commit — nothing to reshape.
- **CI**: all 11 checks green (lint + test matrix across ubuntu/macos/windows × py3.12–3.14). No push this tick, so these checks are authoritative.

Sound bug fix (mask outline was drawn one pixel down-and-right of its fill) with a targeted regression test. Recommending merge; the merge itself stays your call.

wkentaro:

wkentaro:

Rebased onto current `main` (`bd3ca1d7`). The branch had gone `DIRTY`: `main` picked up several `CHANGELOG.md` entries under `## [Unreleased]` after this PR was branched, conflicting with this PR's own entry. Resolved by keeping both, with this PR's entry appended to the end of the `### Fixed` list.

The rebase touched **only** `CHANGELOG.md`. I verified the code diff is byte-identical to the previously reviewed revision:

```
$ diff <(git diff 9d29dc97~1..9d29dc97 -- labelme tests) \
       <(git diff origin/main...HEAD -- labelme tests)
IDENTICAL: code diff unchanged by rebase
```

So the earlier review still stands and I did not re-run the review loop.

Re-verified locally against the new base:

- `make lint` — clean (ruff format, ruff check, ty, taplo, mdformat, yamlfix, typos)
- `make test` — **892 passed** (`QT_QPA_PLATFORM=offscreen`)

Keeping `recommend-merge`. CI is re-running after the force-push; if it comes back red, a later pass will pick that up and revise the verdict.

PR #2342, opened 2026-07-14
test(canvas): cover right-press reselect and pending-move-shape helpers

wkentaro:
`_should_reselect_on_right_press` and `_pick_pending_moved_shape` in
`labelme/_widgets/canvas.py` are pure, branch-rich helpers that were only
reachable through full mouse-event simulation, so their individual branches
were never pinned. This adds direct per-branch unit tests, following the
file's existing pure-helper test pattern (identity-based `Shape` fixtures,
since `Shape` is `@dataclass(eq=False)`).

Each of the two helpers' four branches gets one test. The empty-selection case
uses `hovered_shape=None`, the input that actually distinguishes the first
guard (`if not selected_shapes: return True`) from the fallthrough — verified
by mutation: deleting the guard fails that test.

## Test plan
- [x] `uv run pytest tests/unit/widgets/canvas_test.py` (85 passed)
- [x] full unit suite green (616 passed)
- [x] `uv run ruff format --check` / `ruff check` / `ty check` clean

### DIFF
--- a/tests/unit/widgets/canvas_test.py
+++ b/tests/unit/widgets/canvas_test.py
@@ -24,9 +24,11 @@ from labelme._widgets.canvas import _DraftShape
 from labelme._widgets.canvas import _is_degenerate_draft
 from labelme._widgets.canvas import _normalize_bbox_points
 from labelme._widgets.canvas import _opposite_corner_in_parallelogram
+from labelme._widgets.canvas import _pick_pending_moved_shape
 from labelme._widgets.canvas import _project_oriented_rectangle_corners
 from labelme._widgets.canvas import _reproject_oriented_rectangle_corners
 from labelme._widgets.canvas import _shape_to_draft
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: recommend-merge**

Test-only PR adding coverage for `_should_reselect_on_right_press` and `_pick_pending_moved_shape` in `canvas.py`. Reviewed at medium effort:

- Every assertion matches the helpers' actual control flow exactly (empty selection → reselect; hovering nothing → keep; hovering outside/inside selection → reselect/keep; and the four `_pick_pending_moved_shape` guards).
- Branch is fully current with `main` (0 commits behind the merge-base, no file overlap), so no rebase was needed.
- CLEAN/MERGEABLE, all 11 checks green across the ubuntu/macos/windows × 3.12–3.14 matrix.
- Single clean conventional commit; nothing to reshape.

No meaningful findings; endorsed for merge.

PR #2345, opened 2026-07-14
fix(cli): guard --labels/--flags/--label-flags/--config against over-length values

wkentaro:
Passing a `--labels`, `--flags`, `--label-flags`, or `--config` value longer than the filesystem's filename limit (~255 bytes on Linux, reachable with a few dozen inline comma-separated labels) crashed startup with an unhandled `OSError: [Errno 36] File name too long`. These options probe whether the value names an existing file via `Path(value).is_file()`, and pathlib re-raises `ENAMETOOLONG` instead of returning `False`, so the value never reached the inline-parsing fallback.

Switched the three probes to `os.path.isfile`, which catches `OSError`/`ValueError` and returns `False`, so an over-length value falls through to comma/YAML parsing (or the clean "config file does not exist" error) as intended.

## Test plan
- [x] `tests/unit/__main___test.py::test_parse_list_arg_splits_value_too_long_to_be_a_path` (fails on `main` with the exact `OSError`, passes here)
- [x] `make lint` (ruff format + ruff check + ty)
- [x] full suite: `uv run pytest tests/` (763 passed)

## Related (not addressed here)
`labelme/_app.py:2703` has the same bug class: `Path(file_or_dir).is_dir()` on the positional `path` CLI argument raises `OSError` on an over-length value (e.g. `labelme "$(python3 -c 'print("a"*5000)')"`). Left out to keep this PR surgical; worth a fast-follow.


wkentaro:

Reviewed the diff (report-only `/code-review`, medium effort): the fix is correct and complete.

- **Root cause confirmed.** `Path(value).is_file()` re-raises `OSError` for `ENAMETOOLONG` (errno 36 is not in pathlib's ignored-errno set), so a `--labels`/`--flags`/`--label-flags`/`--config` value longer than `NAME_MAX` crashed at startup. `os.path.isfile()` swallows `OSError`/`ValueError` and returns `False`, so the over-length value is correctly treated as inline data.
- **Complete.** All four user-controlled `Path(...).is_file()` call sites in `labelme/__main__.py` were switched consistently; no sibling call site was left with the same bug. `config_file = Path(config_str)` is retained intentionally for the `config_file.absolute()` error message.
- **Test is meaningful.** 60 comma-joined labels are well over the 255-byte `NAME_MAX`, so `test_parse_list_arg_splits_value_too_long_to_be_a_path` reliably exercises the `ENAMETOOLONG` path.
- **CI green** across all 9 OS/Python combinations; branch is up to date with `main` (`CLEAN`, 0 divergent commits), so no rebase was needed and the existing green reflects the merged result.

No follow-up issues warranted. Ready for maintainer review and merge.

wkentaro:

PR #2348, opened 2026-07-14
fix(config): reject a non-list value for the labels key

wkentaro:
A scalar `labels` override such as `labels: cat` in `~/.labelmerc` slipped past
config validation: a `str` is `Sized` and had no duplicate characters, so the
`labels` duplicate check accepted it and left `config["labels"]` as the string
`"cat"`. Downstream code then iterated it character by character (`c`, `a`, `t`),
silently producing three bogus labels instead of failing.

`_validate_config_item` now rejects any non-list, non-`None` `labels` value with
a clear configuration error (surfaced by the existing config-error dialog in
`_app.py`), mirroring the sibling "section must be a mapping" check. `None`
(default) and valid lists are unaffected; the existing duplicate check is
preserved.

## Test plan
- [x] Added `labels_not_a_list` case to `test_load_config_rejects_invalid_override`
- [x] `uv run pytest tests/unit/_config_test.py` (41 passed) and full suite (763 passed)
- [x] `uv run ruff check` + `uv run ty check` clean
- [x] Smoke: `load_config(config_overrides={"labels": "cat"})` now raises `ValueError: Config key 'labels' must be a list, but got str: 'cat'`


wkentaro:

**Verdict: recommend-merge**

Reviewed the diff (report-only code-review, medium effort) plus a full trace of every producer of the `labels` config value. The change is correct and well-scoped:

- Tightens `labels` validation from any `Sized` to `list`, fixing the reported bug where a scalar string (`labels: cat`) was truthy, passed the duplicate check, and got iterated character-by-character.
- `Sized` import removal is safe (no other uses).
- `value is not None` guard preserves the prior "None skips validation" behavior.
- No real input path produces a tuple/set/dict for `labels` (`_yaml.safe_load` sequences and `_parse_list_arg` both yield `list[str]`), so only the previously-buggy scalar-string case is newly rejected.

CI is green across all 9 platform/version combos; branch is up to date with `main` (0 commits behind), so no rebase was needed. Single clean commit. Test and CHANGELOG entry included.

wkentaro:

wkentaro:

Rebased onto `main` and re-verified. Removing `recommend-merge` for now: the branch has a new head, so the previous endorsement no longer describes what CI has actually tested. A follow-up tick re-applies it once checks land green.

## Why the rebase

The PR had gone `DIRTY` (conflicting). `main` advanced 20 commits past the merge-base and touched all three files this PR changes — `CHANGELOG.md`, `labelme/_config/__init__.py`, and `tests/unit/_config_test.py` — so the prior green check no longer reflected the merged result.

Only `CHANGELOG.md` actually conflicted, and it was a pure additive collision in the `### Fixed` list (this entry versus the entries added by #2340, #2345, #2383, #2338, #2384, #2391). Both sides kept. The two code files auto-merged.

Old head `60488539`, new head `7f4d30ec`. One commit, unchanged message.

## Verification after the rebase

- `uv run pytest tests/unit` — 737 passed
- `uv run ruff check`, `ruff format --check`, `uv run ty check` — all clean
- Smoke on the merged tree, covering the scenario in the description:
  - `labels: cat` in a config **file** → `ValueError: Config key 'labels' must be a list, but got str: 'cat'`
  - same via `config_overrides` → same error
  - `labels: ["cat", "cat"]` → duplicate check still fires
  - `labels: ["cat", "dog"]` → passes
  - default (unset) → stays `None`

The file path matters here: `_validate_config_item` is wired into `_update_dict` for both the config-file merge and the overrides merge, so the headline `~/.labelmerc` case in the description is genuinely covered, not just the overrides path.

The removed `Sized`/`cast` usage left no orphan import — `cast` is still used by `_update_dict`.

## Review

Ran a two-axis review (repo standards + fidelity to the PR description). No meaningful findings on either axis, so no fix pass was needed.

Two low-severity notes, neither worth acting on:

1. The new error message duplicates the shape of the existing "section must be a mapping" error in `_update_dict`. Collapsing them into a helper would touch code outside this PR's purpose, and two call sites do not earn the indirection.
2. The added test drives `config_overrides` rather than writing a temp YAML file, so it is one step removed from the literal `~/.labelmerc` scenario. The implementation is confirmed correct for both paths by the smoke above, and `test_load_config_malformed_crosshair_raises` is the precedent if a file-path case is ever wanted.

## Relationship to #2376

This is an ad-hoc per-field guard, the same category as #2358, which the `wayfinder:config` map explicitly says may land now as an interim crash-fix and be superseded by the typed-schema overhaul later. Track, don't block.

No spin-off issue filed. The sibling list-typed keys `flags` and `file_search` have no type validation either and carry the same latent "string iterated character by character" shape, but that is already inside the scope of #2376 and its per-key inventory in #2378.

wkentaro:

CI is green on the rebased head `7f4d30ec` — all 11 checks pass (lint, plus test across ubuntu/macos/windows on Python 3.12/3.13/3.14), and the branch is no longer conflicting.

Restoring `recommend-merge`. Verdict: **merge**.

PR #2349, opened 2026-07-15
fix(app): render zoom shortcut tooltip readably instead of a raw list

wkentaro:
The zoom control's "What's This" help (Shift+F1 on the Zoom widget) rendered a raw Python list instead of readable shortcuts. `shortcuts["zoom_in"]` is configured as a list of key aliases (`[Ctrl++, Ctrl+=]`), and the tooltip interpolated it straight into an f-string, so the list repr leaked through.

```
before:  The shortcuts <b>['Ctrl</b>+<b>+', 'Ctrl+='],Ctrl+-</b> and <b>Ctrl</b>+<b>Wheel</b> also work on the canvas.
after:   The shortcuts <b>Ctrl</b>+<b>+</b>, <b>Ctrl</b>+<b>-</b> and <b>Ctrl</b>+<b>Wheel</b> also work on the canvas.
```

Each zoom shortcut is now formatted independently through its primary alias and joined, so both `zoom_in` and `zoom_out` render as bold key sequences. This also hardens two config-driven startup crashes the naive per-shortcut formatting would otherwise introduce: a disabled shortcut (`null` or `[]`) is skipped instead of crashing, and `format_shortcut` renders a modifier-less key such as `Delete` as bold text instead of raising.

## Test plan
- [x] `tests/unit/utils/qt_test.py` covers `primary_shortcut` (list alias / disabled / passthrough) and single-key `format_shortcut`
- [x] `tests/e2e/zoom_test.py` drives the real `MainWindow` and asserts `zoom_widget.whatsThis()` for default / list-`zoom_out` / null-`zoom_in` / bare-key configs
- [x] full suite green (`uv run pytest tests/`), `ruff check`, `ty check`
- [x] drove the app under offscreen Qt: default renders cleanly; the null and bare-key configs no longer crash at startup


wkentaro:

Reviewed the diff at medium effort (report-only) as the finalize gate — no meaningful findings.

- **Fix is correct and minimal.** `zoom_in` is configured as a list of key aliases (`[Ctrl++, Ctrl+=]`) while `zoom_out` is a plain string, so the old `f"{shortcuts['zoom_in']},{shortcuts['zoom_out']}"` produced the raw-list artifact `['Ctrl++', 'Ctrl+='],Ctrl+-`. The new `primary_shortcut` helper picks the first alias, each shortcut is formatted independently, and disabled/empty shortcuts are skipped instead of crashing.
- **`format_shortcut` behavior change is safe.** Returning `<b>{text}</b>` for a plus-less key (instead of raising) affects only its two callers, both in `_app.py`; the removed `ValueError` had no production consumer beyond the replaced test.
- **Well covered.** The new e2e test exercises 4 config variants (list, string, `None`, single-key) and asserts the rendered `whatsThis` contains no `[` or `'`; unit tests cover `primary_shortcut` and single-key `format_shortcut`.
- **CI green** (11/11), branch sits on current `main` (no rebase needed), and the CHANGELOG entry matches the file's existing house style.

No `/review-fix`, `/recommit`, or push needed this tick. Merge remains the maintainer's call.

wkentaro:

wkentaro:

Withdrew `recommend-merge` and rebased this branch; it is now awaiting CI, and a later tick will re-verdict it.

**Why the verdict was withdrawn.** `main` had advanced far enough that this branch went `DIRTY` (textually conflicting), so the earlier `recommend-merge` was no longer true: the PR could not have been merged as it stood. Rebasing onto current `main` was required, and that force-push re-triggers CI.

**What changed in the push.** The source fix is byte-for-byte the same as what was reviewed. The rebase also carried in test hardening that a prior processing tick had produced locally but never pushed: the e2e parametrization now asserts the exact expected shortcut strings (with `pytest.param` ids) instead of only asserting the absence of `[` and `'`, and adds explicit cases for one-disabled and both-disabled shortcuts.

**Verified locally on the rebased tree:** `901 passed`, and `make lint` fully clean (`ruff format --check`, `ruff check`, `ty check`, `taplo`, `mdformat`, `yamlfix`, `typos`).

**Review.** I ran a two-axis review (repo standards + fidelity to this PR's description) over the rebased diff.

- Standards: no hard violations. The walrus in the join matches existing repo idiom, annotations and naming follow the conventions, and the re-export ordering is preserved. Widening `format_shortcut` to render a modifier-less key as bold instead of raising is the right call rather than a workaround: `default_config.yaml` already treats bare-key shortcuts as first-class (`delete_shape: Delete`, `toggle_all_shapes: T`), so the old `ValueError` encoded an invariant the shortcut schema never actually had.
- Spec: the default rendering claim checks out exactly. I ran the real formatting code against `default_config.yaml` and got `The shortcuts <b>Ctrl</b>+<b>+</b>, <b>Ctrl</b>+<b>-</b> and <b>Ctrl</b>+<b>Wheel</b> also work on the canvas.`, matching the PR description verbatim. `shortcuts["zoom_in"]` also cannot `KeyError`, since `get_config` always merges over `default_config.yaml`.

One real edge case found, filed as a follow-up rather than fixed here: with **both** zoom shortcuts disabled, the sentence renders `The shortcuts  and <b>Ctrl</b>+<b>Wheel</b> also work on the canvas.` — a double space and a dangling "and". The natural fix is a second source string for the singular case, but all 20 locales in `labelme/translate/` currently have zero `type="unfinished"` entries and `make check_translate` is a hard release gate, so adding a source string would unfinish all 20 and block the gate for a cosmetic edge case. Better bundled with the next batch of translatable-string changes.

follow-ups: #2420

wkentaro:

`recommend-merge`. The rebase from the previous tick has landed and CI came back fully green: all 11 checks succeeded against the current head `fab8fa01` (lint, plus test across ubuntu/macos/windows on Python 3.12/3.13/3.14, plus CLA). Merge state is `CLEAN`.

No new push was made this tick, so these checks are authoritative for the merged result and the PR is verdicted directly rather than waiting another round. The review recorded in the prior comment stands unchanged: the source fix is byte-for-byte what was reviewed, standards and spec both check out, and the one cosmetic edge case (both zoom shortcuts disabled) is deliberately deferred to #2420 rather than blocking this merge.

wkentaro:
> *This was generated by AI.*

Closing in favor of #2443, which removes the string this PR was fixing.

The zoom control's `whatsThis` text is unreachable from the UI. Qt has no global "enter What's This mode" binding: `QWidget::event` handles `Shift+F1` only on the widget that currently holds keyboard focus, and only when that widget's own `whatsThis` is non-empty. labelme has no Help menu entry, no `QWhatsThis` action, and no titlebar `?` button, so displaying the text required clicking the toolbar zoom box to focus it and then pressing `Shift+F1`. Before an image is loaded the widget is disabled and cannot take focus at all.

`format_shortcut` had exactly one caller in the codebase: that string. #2443 deletes both, and moves the one hint not documented elsewhere (`Ctrl+Wheel`) into a hover tooltip. The zoom key bindings this PR was formatting are already shown next to Zoom In and Zoom Out in the View menu.

PR #2350, opened 2026-07-15
fix: skip an invalid label_flags regex instead of crashing

wkentaro:
An invalid regular expression in a `label_flags` config entry (e.g. a typo like `person-(`) raised an uncaught `re.error` and crashed the app — either the first time a labeled shape was loaded (`_shapes_from_dicts`), or on the first keystroke in the label dialog (`_update_flags`, wired to `textChanged`). `label_flags` patterns arrive straight from `~/.labelmerc` / `--label-flags` with no validation, so a single typo took the app down.

Both call sites now validate their `label_flags` patterns once up front instead of on every `re.match` call: `_shapes_from_dicts` before the shape loop, and `LabelDialog` in `__init__` before the first keystroke. An uncompilable pattern is dropped with a single warning naming it, and matching proceeds using only the patterns known to be valid, mirroring the existing `file_search` `re.error` guard.

## Test plan
- [x] `tests/unit/_app_test.py`: invalid pattern is skipped; a valid pattern still applies when another entry is invalid
- [x] `tests/unit/widgets/label_dialog_test.py`: an invalid pattern is skipped; a valid sibling still applies
- [x] `tests/e2e/label_dialog_flags_test.py` (happy path) unchanged and green
- [x] Reproduced the original crash on `main`, confirmed the branch loads without crashing and warns once across 100 shapes
- [x] `uv run ruff check`, `ruff format --check`, `ty check`, full unit suite green


wkentaro:

**Verdict: recommend-merge**

- CI green across all 9 platform/Python combinations; merge state CLEAN, sitting at main's tip (no rebase needed).
- `/code-review` (medium) found nothing meaningful: the fix pre-validates each `label_flags` pattern via `re.compile` in `_shapes_from_dicts` (skipping invalid ones with a warning) and guards `re.match` with `try/except re.error` in `LabelDialog._update_flags`. The `isinstance(shape.label, str)` guard is preserved; no behavior is dropped.
- Single clean commit; `/recommit` was a no-op.
- Verify smoke: the 11 unit tests in `tests/unit/_app_test.py` and `tests/unit/widgets/label_dialog_test.py` drive both real code paths (including qtbot for the dialog) and pass locally in 0.5s.

Ready for maintainer review and merge.

wkentaro:

wkentaro:

Rebased onto current `main` (`bd3ca1d7`) and force-pushed with lease. `mergeStateStatus` went `DIRTY` → `UNSTABLE`, so the conflict that caused the earlier withdrawal is resolved; CI is re-running and the verdict follows once it lands.

The rebase was content-preserving: the patch against `main` is still the same five files, and no conflict resolution changed the fix. Re-checked on the rebased tree:

- `make lint` green (ruff format/check, `ty`, taplo, mdformat, yamlfix, typos).
- Full suite green: 895 passed.
- Report-only `/code-review` (Standards + Spec axes) found no hard violations and no missing/extra behavior versus the PR description. Only two judgement-call nits, both left alone deliberately: the dialog path skips a bad pattern silently while the load path warns (per-keystroke logging would spam), and the `try/except re.error` shape appears in both files but with different mechanics, so extracting a helper would not pay.
- Smoke through a real `~/.labelmerc` carrying `label_flags: {"person-(": [occluded], "cat": [sitting]}`: the load path logs `Invalid label_flags pattern: 'person-('` once and still applies `{"sitting": False}`; the dialog path renders the `sitting` checkbox. The same config raises `re.error: missing ), unterminated subpattern at position 7` on `main`, so the crash and the fix both reproduce.

No re-review was implied by the withdrawal, and none of the above changed the branch beyond the rebase.

wkentaro:

CI landed green on the rebased head (`lint`, `cla`, and all nine `test` jobs across ubuntu/macos/windows x Python 3.12/3.13/3.14). `mergeStateStatus` is `CLEAN`.

Nothing changed on the branch since the previous tick, which already covered the review, the lint/test run, and the smoke through a real `~/.labelmerc`. This tick only resolved the pending checks.

PR #2351, opened 2026-07-15
fix(shape): guard insert_point against non-polyline shapes

wkentaro:
`Shape.remove_point` guards on `can_remove_point()` and no-ops with a warning
when the shape type does not support point removal, but its sibling
`Shape.insert_point` had no analogous check: it would happily insert a vertex
into a fixed-cardinality shape (rectangle, circle, line, point,
oriented_rectangle, mask), corrupting its point count. This adds the symmetric
`can_add_point()` guard so the model layer is self-consistent.

Defense-in-depth only: the sole production caller (`add_point_to_edge`) is
already gated upstream by `find_hover_target`'s edge pass, which filters on
`can_add_point()`, so this is not reachable through the current UI. No
CHANGELOG entry for that reason, consistent with how internal refactor/test
PRs in this repo skip the changelog.

## Test plan
- [x] Added `test_insert_point_is_noop_for_non_polyline` and `test_insert_point_noop_logs_shape_state`, mirroring the existing `remove_point` no-op tests
- [x] `uv run pytest tests/unit/_shape_test.py` (54 passed)
- [x] `uv run ruff check` / `ruff format --check` / `ty check` on the changed files
- [x] `uv run pytest tests/e2e/canvas_interaction_test.py tests/e2e/oriented_rectangle_test.py tests/e2e/shape_editing_test.py` (add-point-to-edge flow still green)

### DIFF
--- a/labelme/_shape.py
+++ b/labelme/_shape.py
@@ -57,6 +57,13 @@ class Shape:
         return self.shape_type in POLYLINE_SHAPE_TYPES
 
     def insert_point(self, i: int, point: npt.ArrayLike, label: int = 1) -> None:
+        if not self.can_add_point():
+            logger.warning(
+                "Cannot add point to: shape_type={!r}, len(points)={:d}",
+                self.shape_type,
+                len(self.points),
+            )
+            return
         point = np.asarray(point, dtype=np.float64).reshape(2)
         self.points = np.insert(self.points, i, point, axis=0)
         self.point_labels = np.insert(self.point_labels, i, label)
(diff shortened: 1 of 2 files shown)


wkentaro:

Reviewed at medium effort (report-only); no meaningful findings.

- The new guard in \`insert_point\` mirrors the adjacent \`remove_point\` exactly (same `can_add_point()` condition, warning format, and early return) — consistent and idiomatic.
- No caller breaks: the sole production caller \`add_point_to_edge\` (\`canvas.py\`) only ever fires for a \`HitKind.EDGE\` target, which \`_canvas_interaction.py\` produces exclusively for shapes where \`can_add_point()\` is already true. The guard is defense-in-depth.
- Both new tests are correct, including the loguru format assertion.

No rebase needed (branch is on the current \`main\` tip, 0 divergence); single clean conventional commit, so no history reshaping. CI is green across all 9 test matrices + lint. No push this tick, so the existing green is authoritative.

PR #2357, opened 2026-07-16
refactor(app): flatten save_labels duplicate-file check into guard clauses

wkentaro:
The nested `len(items) > 0` / `len(items) != 1` check in `save_labels` reads more directly as two guard clauses: raise on more than one match, then check the single match. Behavior is identical for 0, 1, and >1 matches, and it drops a redundant `len()` call.

No CHANGELOG entry: pure internal refactor with no user-facing behavior change.

## Test plan
- [x] `make lint` (ruff format/check, ty) clean
- [x] `uv run pytest tests/` — 807 passed, including the e2e save-path suite that exercises `save_labels`

### DIFF
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -1805,9 +1805,9 @@ class MainWindow(QtWidgets.QMainWindow):
             items = self._docks.file_list.findItems(
                 self._image_path, Qt.MatchFlag.MatchExactly
             )
-            if len(items) > 0:
-                if len(items) != 1:
-                    raise RuntimeError("There are duplicate files.")
+            if len(items) > 1:
+                raise RuntimeError("There are duplicate files.")
+            if items:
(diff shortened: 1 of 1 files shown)


wkentaro:

Reviewed the diff (`labelme/_app.py`, `save_labels`): the change inverts the nested `len(items) > 0` / `len(items) != 1` check into two guard clauses (`if len(items) > 1: raise`, then `if items: setCheckState`). Behavior is identical across the 0 / 1 / >1 cases, and the `RuntimeError` still propagates uncaught by the `LabelFileError`-only `except`, exactly as before. `/code-review` at medium effort found nothing; single clean commit; no CHANGELOG needed (internal refactor, no user-facing change).

CI: all 11 checks green. The `ubuntu-latest, 3.12` job had been hung `in_progress` for ~24h (siblings passed in ~90s) — a stuck runner, not a real failure. I re-ran it and it passed.

PR #2358, opened 2026-07-17
fix(config): don't crash at startup on a non-string ai.default

wkentaro:
A config whose `ai.default` is not a string (e.g. `ai:\n  default: true`) made the
whole config unloadable: `_migrate_config_from_file` passed the value straight to
`re.match`, raising `TypeError: expected string or bytes-like object`. The
config-error backstop in `MainWindow._load_config` catches it, but only by
discarding the entire file and falling back to defaults, so one bad `ai.default`
silently dropped every other setting the user had configured.

The old-model-name migration now only runs when `ai.default` is a string, mirroring
the `isinstance(shortcuts, dict)` guard directly above it in the same function. The
rest of the config then loads normally and an unusable `ai.default` degrades to the
AI model widget's existing logged warning.

Observed with a `~/.labelmerc` of `labels: [cat, dog]` + `ai: {default: true}`:

| | `labels` | `ai.default` | startup |
| --- | --- | --- | --- |
| before | `None` (user's list discarded) | `'Sam2 (balanced)'` (default) | "Configuration Errors" dialog |
| after | `['cat', 'dog']` | `True` | `Default AI model is not found: True` warning |

## Test plan
- [x] `test_migrate_tolerates_non_string_ai_default` (bool/int/list) and `test_load_config_keeps_other_keys_when_ai_default_is_not_a_string`; both fail with the guard reverted
- [x] `uv run pytest tests/` (895 passed)
- [x] `make lint` (ruff format/check, `ty`, taplo, mdformat, yamlfix, typos)
- [x] Offscreen `MainWindow` launch against the config above, on this branch and on `main`, producing the table


wkentaro:

Reviewed at high effort (correctness + conventions/cleanup finders); no findings.

- The fix replaces the truthy check `(model_name := ai.get("default"))` with `isinstance(model_name := ai.get("default"), str)`, so `re.match` is only called on strings. The walrus binds `model_name` before `isinstance` evaluates, so behavior is correct. Edge cases (`""`, `None`, `False`, `0`) skip the migration identically to before, so no regression for previously-working configs.
- Mirrors the neighboring `shortcuts` guard style, as the PR describes.
- Two new tests genuinely exercise the previously-crashing path: `_migrate_config_from_file` at the unit level (bool/int/list) and `load_config` end-to-end via a `tmp_path` YAML with `ai.default: true`. Full `_config_test.py` suite passes (44).
- Changelog entry correctly placed under `[Unreleased] > Fixed` with `#2358` linked.

No rebase needed: `main` gained 0 commits since the merge-base and there is no file overlap; merge state is `CLEAN` and all 11 CI checks are green. Leaving the merge to the maintainer.

wkentaro:

wkentaro:

wkentaro:

Rebased onto current `main` (`bd3ca1d7`) and force-pushed with lease; `mergeStateStatus` `DIRTY` → resolved. The only conflict was an additive one in `CHANGELOG.md` (both sides appending under `### Fixed`), and the code diff is unchanged by the rebase: still the single `isinstance(model_name := ai.get("default"), str)` guard.

The report-only `/code-review` gate found one real issue, so the branch was corrected rather than verdicted as-is.

**Standards axis: clean.** No documented-standard violations; changelog placement, annotations, and the local walrus idiom all match the repo.

**Spec axis: the PR premise had gone stale.** The description and CHANGELOG both claimed a *startup crash*, but #2384 landed on `main` in the meantime and widened `MainWindow._load_config` to `except Exception`, so the `TypeError` is now caught. The fix is still worth having, but for a different reason: the backstop recovers by discarding the **whole** config file and falling back to defaults, so one bad `ai.default` silently drops every other setting the user configured.

Measured with `~/.labelmerc` = `labels: [cat, dog]` + `ai: {default: true}`, launching `MainWindow` offscreen on each side:

| | `labels` | `ai.default` | startup |
| --- | --- | --- | --- |
| `main` | `None` (user list discarded) | `'Sam2 (balanced)'` (default) | "Configuration Errors" dialog |
| this branch | `['cat', 'dog']` | `True` | `Default AI model is not found: True` warning |

Changes made this tick:
- Reworded the CHANGELOG entry and PR description to the accurate framing (config silently reset to defaults, not a crash); the commit message body too.
- Renamed `test_load_config_non_string_ai_default_does_not_crash` → `test_load_config_keeps_other_keys_when_ai_default_is_not_a_string` and made it assert a sibling key survives, since "does not crash" no longer describes what the guard buys.

Verified: `make lint` green; `uv run pytest tests/` 895 passed; both new tests fail with the guard reverted (4 failed, 49 passed), so they are genuine regression tests.

CI is re-running on the force-pushed head; the verdict follows once it lands.

wkentaro:

`recommend-merge`.

CI landed green on the force-pushed head `2aa8832` (11/11 checks: `cla`, `lint`, and `test` across ubuntu/macos/windows x Python 3.12-3.14). `mergeStateStatus` is `CLEAN` against current `main` (`bd3ca1d7`), which is the exact base this branch was rebased onto last tick, so nothing has gone stale since.

No push this tick, so the checks above are authoritative and the verdict follows directly. The review, the corrected framing, and the retitled/strengthened test are all in the previous comment; nothing further was needed here.

Merging stays yours.

PR #2360, opened 2026-07-17
fix(app): resolve current label path honoring output dir

wkentaro:
`current_label_file_path` derived an image-adjacent `.json` path via `Path(self._image_path).with_suffix(".json")`, ignoring both the tracked label file and the output directory. Every other call site (autosave, save-as, the file-list checkmarks) already resolves through `_resolve_label_path`, so this one was the outlier. With `--output` / Change Output Dir set, `delete_file` and `has_label_file` (which drives the Delete-File action state) therefore targeted the wrong path: the action could stay wrongly disabled, or delete a stale image-adjacent file while leaving the saved annotation untouched.

It now returns `self._label_file_path` (the loaded/saved label file) when set, falling back to the output-dir-aware `_resolve_label_path` when none exists yet. Preferring the tracked path also keeps a directly-opened label file (opened from outside the output dir) as the delete target, rather than an unrelated file inside the output dir.

## Test plan
- [x] Added `tests/e2e/file_operations_test.py::test_delete_file_respects_output_dir` (covers the pre-save output-dir fallback and the save-then-delete flow) and `::test_current_label_file_path_prefers_opened_file` (direct-open branch). Both fail on `main`.
- [x] Full suite green: `uv run pytest tests/` (809 passed)
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run ty check` clean
- [x] Drove the real `MainWindow` (offscreen Qt): with `--output` set, the Delete-File menu action enables only after a save into the output dir, deletes that file, and leaves a committed image-adjacent label file untouched


wkentaro:

Reviewed the diff finalizing `current_label_file_path()`. The fix correctly returns the loaded/saved `_label_file_path` when present and otherwise falls back to the output-dir-aware `_resolve_label_path(...)`, so both `has_label_file()` (Delete-File action state) and `delete_file()` (deletion target) now honor `--output` / Change Output Dir instead of always assuming an image-adjacent `.json`.

- Traced every `current_label_file_path()` caller and every `_label_file_path` / `_image_path` assignment: they are always updated together, so the fallback cannot leak a stale path.
- Two new e2e tests cover both branches (opened-file preference and output-dir resolution before/after save); both pass locally.
- CHANGELOG entry present. CI green, merge state CLEAN, based on latest `main` (no rebase needed).

A code-review pass at medium effort surfaced no meaningful findings.

wkentaro:

PR #2364, opened 2026-07-18
fix(canvas): slide along the exiting edge when a clamped corner point is dragged

wkentaro:
With out-of-bounds points disabled (the default), dragging a shape vertex out past an image corner correctly snaps it to the corner, but continuing to drag the cursor sideways while staying above or below the image left the vertex frozen at the corner's x instead of sliding along the top/bottom edge under the cursor.

`_compute_intersection_edges_image` clips the drag segment against the image with Liang-Barsky. In the `t_exit == 0` case (the start already sits on a boundary) it picked which edge to slide along from `start_x`'s literal position. At a corner the start sits on both an x and a y boundary, so that check always froze x even when only the y-axis was actually exiting. The fix records which boundary the loop's minimum came from and slides along that axis instead.

## Test plan
- [x] Added regression cases completing the corner x exit-direction matrix in `tests/unit/widgets/canvas_test.py` (all four corners, both slide directions plus diagonal-stays); the four branch-sensitive cases fail on `main`
- [x] Drove `Canvas._bounded_move_vertex` through the real constrain path under offscreen Qt: a vertex at corner `(100,0)` dragged to `(50,-10)` now moves to `(50,0)` instead of freezing at `(100,0)`
- [x] `make lint` (ruff format + ruff check + ty) and full `pytest` green


wkentaro:

Finalize pass on this own PR:

- **Rebase:** not needed. `main` gained 0 commits since the merge-base and the PR is up to date; merge state was `UNSTABLE` only because CI was still running.
- **Review (`/code-review`, medium):** the fix is correct. The new `exit_axis` tracking picks the slide edge from the axis the segment actually exits through, which is per-call *more* geometrically correct than the old start-position check (e.g. a point clipped to the bottom-left corner with the cursor below-right now slides along the bottom edge instead of freezing at the corner). Proven equivalent to the old check for all non-corner boundary points, so no regression there. Tie-breaking at a corner (diagonal-out) correctly keeps the point pinned.
- **Tests:** the 10 new table-driven cases cover all four corners × three exit directions; `tests/unit/widgets/canvas_test.py` green locally (87 passed).
- **Commits:** already a single clean, well-scoped commit — no `/recommit` needed.
- **CI:** green across all 11 checks (lint + test on ubuntu/macos/windows × py3.12–3.14). No push this tick, so existing checks are authoritative.

**follow-ups: #2365** — while fuzzing I found a *pre-existing* best-effort limitation in the oriented-rectangle consumer of this helper (`_reproject_oriented_rectangle_corners`): far-off-canvas corner drags can leave a corner out of bounds even with out-of-bounds points disabled, via a degenerate `edge_a == edge_b` collapse. It is not introduced by this PR (the pre-change code exhibits comparable out-of-bounds rates) and does not block merge; filed separately so it isn't lost.

wkentaro:

wkentaro:

wkentaro:

Rebased onto current `main` (`bd3ca1d7`) and force-pushed with lease: `1599a7b2` → `15f3cba0`. `mergeStateStatus` `DIRTY` is resolved. The only conflict was an additive one in `CHANGELOG.md` (both sides appending under `### Fixed`); the code and test diff is byte-identical to the pre-rebase one.

The report-only `/code-review` gate came back clean on both axes, so `/review-fix` was skipped and nothing in the change was altered.

**Standards: clean.** ADR 0004 governs this exact path; the fix stays inside the gated clamp and doesn't reintroduce a blanket one. Changelog placement, the `pytest.param(..., id=...)` test convention, and the why-not-what comment all match the repo. One judgement call only, not acted on: `exit_axis` is a raw `"x"`/`"y"` string rather than a `Literal["x", "y"]`. It's a local, and the repo's own rule is not to annotate locals, so leaving it is the conventional choice here.

**Spec: clean.** No scope creep, nothing missing. The one thing worth probing was the `t_exit == 0` tie between an x and a y boundary, where strict `t < t_exit` plus the initial `exit_axis = "x"` makes x win arbitrarily. That tie is provably inconsequential: a genuine tie means `p2` is exterior on both axes, so `clip(p2.x())  == start_x` and `clip(p2.y()) == start_y` and both branches compute the identical frozen-corner point. Confirmed over 2039 randomized diagonal-exit trials across all four corners, 0 mismatches.

Verified on the rebased head:
- `make lint` green (ruff format + check, `ty`, taplo, mdformat, yamlfix, typos).
- `uv run pytest tests/` → **901 passed**.
- Drove the real `Canvas._bounded_move_vertex` path under offscreen Qt on a 100x50 image: `(100,0)→(50,-10)` gives `(50,0)`, `(0,0)→(60,-10)` gives `(60,0)`, `(0,50)→(60,60)` gives `(60,50)`, `(100,50)→(110,20)` gives `(100,20)`, and the diagonal `(100,0)→(110,-10)` correctly stays frozen at `(100,0)`.
- Load-bearing check: with `canvas.py` reverted to `main` and the tests kept, the first three of those smokes regress to the frozen corner and 4 of the new cases fail. The description said two branch-sensitive cases fail; the measured number is four, so I corrected that line in the PR body.

History was already a single atomic commit (fix + regression tests + changelog) with an accurate message, so `/recommit` was a no-op.

CI is re-running on `15f3cba0`; the verdict follows once it lands.

wkentaro:

`recommend-merge`.

CI landed green on the rebased head `15f3cba0` (11/11: `cla`, `lint`, and `test` across ubuntu/macos/windows x Python 3.12-3.14), and `mergeStateStatus` is now `CLEAN` against current `main` (`bd3ca1d7`), the exact base this branch was just rebased onto.

Both review axes were clean and the change itself is unmodified from what you wrote; the rebase only re-based it and resolved an additive `CHANGELOG.md` conflict. Details and the verification runs are in the previous comment.

Merging stays yours.

PR #2368, opened 2026-07-19
test(shape-render): cover is_hit_by_point and bounds branches

wkentaro:
`is_hit_by_point` is the shape hit-test that canvas selection and hover depend on, yet it had no direct unit coverage despite five distinct branches (line/linestrip edge distance, `points` never-hit, `point` radius, `mask` pixel lookup with an out-of-bounds guard, and the polygon/rectangle path-containment fallback) — it was exercised only transitively through GUI e2e canvas tests. This adds focused unit tests for each branch plus `bounds()`.

The two tests that reach the `QPainterPath` fallback are left unmarked (no `@pytest.mark.gui` / `qapp`): `QPainterPath` is a Qt value type whose `contains()` / `boundingRect()` run with no `QApplication`, confirmed by running them in isolation.

## Test plan
- [x] `uv run pytest tests/unit/widgets/_shape_render_test.py -q` — 12 passed
- [x] `uv run ruff check` / `ruff format --check` / `uv run ty check` — clean

### DIFF
--- a/tests/unit/widgets/_shape_render_test.py
+++ b/tests/unit/widgets/_shape_render_test.py
@@ -9,6 +9,7 @@ from labelme._shape import ShapeType
 from labelme._widgets._shape_render import Palette
 from labelme._widgets._shape_render import ShapeRenderContext
 from labelme._widgets._shape_render import bounds
+from labelme._widgets._shape_render import is_hit_by_point
 from labelme._widgets._shape_render import render_shape
 
 _SIZE = 200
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding coverage for \`is_hit_by_point\` and \`bounds\` in \`_widgets/_shape_render.py\`. Reviewed the diff against source: the tests exercise every branch of \`is_hit_by_point\` and their assertions match the implementation.

- line/linestrip -> \`nearest_edge_index\` (incl. the phantom closing-edge case, asserting the #2307 fix stays fixed)
- \`points\` -> never a body hit; \`point\` -> radius = \`point_size / 2\` = 4px; empty point guarded to a miss
- \`mask\` -> translated pixel read with an out-of-bounds guard
- polygon -> path containment; \`bounds\` -> bounding rect extent

Branch is rebased onto current \`main\` and CI is green across all 11 configs (lint + 3 OS x Python 3.12/3.13/3.14). No rebase or push needed this tick. Test-only, so no runtime smoke to verify.

PR #2370, opened 2026-07-19
fix(ai-assist): initialize the canvas AI output format at startup

wkentaro:
Reworked on top of main after #2403 landed. The settings-page merge fixed the model half of the original bug (the configured `ai.default` not reaching the canvas when it resolves to the first combo entry) by pulling `current_model_id` right after constructing the panel, which conflicted with and superseded this PR's original approach of pushing the initial selections through the change-signal closures.

This rework completes the remaining half with the same pull pattern. The output-format combo still reaches the canvas only via `currentIndexChanged`, which never fires during construction because a fresh combo already sits at index 0, so the canvas session's output format is correct only by coincidence: its hardcoded default (`polygon`) happens to match the first combo entry, and reordering the combo items or changing the session default would silently desync the UI from the canvas. `MainWindow` now initializes the canvas output format from the panel's resolved selection, exactly as it already does for the model, and both combos in the widget are wired set-index-then-connect so construction fires no change signals. The CHANGELOG entry documents the user-visible `ai.default` bug, which shipped in v7.0.4 and is fully fixed by #2403 plus this PR.

## Test plan
- [x] Added `tests/unit/widgets/_ai_assisted_annotation_widget_test.py`: construction exposes the resolved defaults via `current_model_id` / `output_format` without firing callbacks, an unknown default falls back to the first entry, and post-init combo changes fire the callbacks
- [x] End-to-end offscreen check: constructing `MainWindow` with `ai.default` set to `EfficientSam (speed)`, `Sam2 (balanced)`, and an unknown value yields the matching canvas session model and output format `polygon` in all three cases
- [x] `make lint` and full `pytest` (996 passed)


wkentaro:

Reviewed the diff (report-only `/code-review`, medium effort) plus the enclosing widget and the `_app.py` call site. Findings:

- **Correct fix.** `setCurrentIndex(model_index)` runs while `currentIndexChanged` is still disconnected, then a single explicit `push_model(model_index)` fires, then the signal is connected. That's exactly one push, no double-push and no missed-push for index 0, closing the reported gap where an already-selected first entry (or the unknown-value fallback to index 0) silently left the canvas on the hardcoded `sam2:latest`. The output-format combo gets the same treatment, which also fixes it never being pushed at startup before.
- **Call site is safe.** Both callbacks (`canvas.set_ai_model_name` / `set_ai_output_format`) are trivial setters on `_ai_assist_session`, which the canvas builds in its own `__init__` well before this widget is constructed. Calling them during `__init__` was already the pre-existing behavior for any non-first default model, so nothing new is exercised too early.
- **Tests** cover first-listed default, non-first default, unknown-value fallback, and a later user selection. All green locally and in CI (all 11 checks passing).
- CHANGELOG entry present with the PR link, in the right `### Fixed` subsection.

No meaningful findings; skipped `/review-fix`. History is a single clean commit. No follow-ups filed.

PR #2371, opened 2026-07-19
test(_shape): cover Shape rejecting an unknown shape_type

wkentaro:
`Shape.__post_init__` guards every construction against an invalid `shape_type`
(`raise ValueError(f"Unexpected shape_type: ...")`), but that branch was the sole
uncovered line in `_shape.py`. This adds a test pinning that contract so a
regression in the guard is caught.

## Test plan
- [x] `uv run pytest tests/unit/_shape_test.py -q` (57 passed)
- [x] `uv run ruff check` and `uv run ty check` on the file

### DIFF
--- a/tests/unit/_shape_test.py
+++ b/tests/unit/_shape_test.py
@@ -204,6 +204,11 @@ def test_can_add_point(shape_type: ShapeType, expected: bool) -> None:
     assert Shape(shape_type=shape_type).can_add_point() is expected
 
 
+def test_constructor_rejects_unknown_shape_type() -> None:
+    with pytest.raises(ValueError, match="Unexpected shape_type: bogus"):
+        Shape(shape_type="bogus")  # ty: ignore[invalid-argument-type]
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR: adds `test_constructor_rejects_unknown_shape_type`, pinning the existing `ValueError("Unexpected shape_type: ...")` guard in `Shape.__post_init__` (`labelme/_shape.py:50`).

- Branch is up to date with `main` (0 divergent commits, mergeState CLEAN) — no rebase needed.
- Cheap `/code-review` over the diff surfaced no findings; the `match=` pattern has no stray regex metacharacters and the `# ty: ignore[invalid-argument-type]` suppression is correct for the intentionally-invalid literal.
- Single clean atomic commit; no reshape needed.
- All CI checks green (lint + test across ubuntu/macos/windows × py3.12-3.14).

Recommending merge; the merge itself stays your call.

PR #2374, opened 2026-07-20
fix(label-file): reject a non-int imageHeight/imageWidth

wkentaro:
`read_label_file` compared the top-level `imageHeight` / `imageWidth` against the real image size with `!=`, with no type check on the declared value. Because `bool` is a subclass of `int` (`True == 1`, `False == 0`) and `3.0 == 3`, a hand-edited or third-party annotation file with a boolean or float dimension slipped past the check whenever the value happened to equal the actual dimension, instead of being rejected as malformed. `_check_image_dimensions` now validates both fields as `int` (excluding `bool`) before the mismatch comparison, matching the per-field `raise TypeError(f"{name} must be int: {value}")` idiom used elsewhere in the module.

Sibling of #2323, which closes the same `bool`-as-`int` gap for `group_id` and point coordinates. The two are independent and can merge in either order.

## Test plan
- [x] `tests/unit/_label_file_test.py` gains `imageHeight`/`imageWidth` bool and float cases (each fails before the fix, passes after)
- [x] `uv run pytest tests/unit` (722 passing) + `uv run ruff check` + `uv run ty check`
- [x] Smoke: a file with `"imageHeight": 3.0` on a 3px-tall image loaded silently before, now raises `LabelFileReadError`; a valid file still loads


wkentaro:

Reviewed at medium effort (report-only; no findings survived).

- Single clean commit; `mergeState: CLEAN`; branch is based on current `main` HEAD (0 commits behind), so no rebase was needed and the existing green CI is authoritative (all 11 checks pass).
- The new guard in `_check_image_dimensions` mirrors the established `group_id` idiom exactly (`isinstance(x, bool) or not isinstance(x, int)` → `raise TypeError(f"... must be int: ...")`), correctly rejecting `bool` (an `int` subclass) and `float` dimensions that previously slipped past the `!=` comparison when numerically equal.
- Both call sites (`read_label_file`, `write_label_file`) already catch `TypeError` and re-wrap it as `LabelFile*Error`, so no caller breaks.
- Tests cover all four cases (bool/float × height/width); CHANGELOG entry is in the correct `### Fixed` subsection with the PR linked.

Correctly scoped as the follow-up to #2323, which fixed the same `bool`-as-`int` class for `group_id`.

wkentaro:

wkentaro:

Rebased onto current `main` to clear the conflict that withdrew the earlier verdict (the CHANGELOG `[Unreleased]` section had moved on; the code hunk in `_label_file.py` merged cleanly). No code change from the previous revision.

Re-reviewed the rebased diff at medium effort on both axes (repo standards + PR spec); nothing needing a fix:

- Standards: the guard follows the file's existing idiom (early raise before the `img_data_to_pil` decode, `f"X must be TYPE: {value}"` message shape), `TypeError` was already in `read_label_file`'s except tuple, and the tests slot into the existing parametrize table.
- Spec: read-path fix, bool and float cases for both fields, and the CHANGELOG entry are all present. One note, not a defect: `_check_image_dimensions` is shared with `write_label_file`, so the write path tightens too. Its only caller passes `QImage.height()`/`.width()`, so nothing in-tree is affected, and rejecting a non-int dimension on write is the same correct behavior.

Smoke on the rebased branch: a 3x4 image with `"imageHeight": 3.0`, `true`, or `"imageWidth": 4.0` now raises `LabelFileReadError: imageHeight must be int: 3.0` (etc.), and a valid file still loads. `uv run pytest tests/unit` is green (740 passed).

Awaiting CI on the rebased head; the verdict lands next tick.

wkentaro:

Re-checked after the rebase onto current `main` (head `af92775`):

- Merge state is `CLEAN`; the branch is a single logical commit on top of `main`.
- All 11 checks are green on the current head (`cla`, `lint`, and `test` across ubuntu/macos/windows x 3.12/3.13/3.14).
- Re-read the rebased diff: `_check_image_dimensions` now rejects a non-`int` `imageHeight`/`imageWidth` before comparing, with the `bool`-excluding `isinstance` guard needed because `bool` subclasses `int`. The `TypeError(f"... must be int: {...}")` shape matches the existing validators in `labelme/_label_file.py` (`flags`, `label`, `group_id`). Tests extend the existing parametrized read-error table with bool and float cases for both fields.

No new findings. Ready for your review and merge.

Note for merge ordering: #2323 also touches `labelme/_label_file.py` with bool-rejection validation (for `group_id` and point coordinates). The edits are in distinct functions, but whichever lands second may want a quick rebase.

PR #2383, opened 2026-07-20
fix(cli): strip whitespace around comma-separated --labels/--flags entries

wkentaro:
`_parse_list_arg`'s file-reading branch strips each line, but its inline comma-split branch did not, so a naturally-spaced list such as `--labels "cat, dog, person"` registered `" dog"` and `" person"` with a leading space. Those spaced strings then flow into the predefined-label list, and with `--validate-label exact` a later typed `"dog"` is rejected because the registered label is actually `" dog"`. The inline branch now strips whitespace and drops blank entries, using the exact same comprehension idiom as the file branch.

## Test plan
- [x] New parametrize cases (`"cat, dog, person"`, `"cat, , dog"`) in `tests/unit/__main___test.py` fail on `main`, pass with the fix
- [x] `uv run pytest tests/unit` (727 passed), `uv run ruff check`, `uv run ty check` all green


wkentaro:

A tight, well-scoped fix: the inline comma-split branch of \`_parse_list_arg\` now strips whitespace around each entry and drops blanks (\`[line.strip() for line in value.split(",") if line.strip()]\`), matching the file-reading branch just above it. So `--labels "cat, dog, person"` no longer registers `" dog"`/`" person"` with leading spaces, and a later `--validate-label exact` no longer spuriously rejects `"dog"`.

- Code review (medium): **CLEAN** — logic correct on all edge cases (empty string, all-whitespace segments, trailing comma), correctly scoped (only `--labels`/`--flags` use this helper), matches the existing in-file convention.
- Smoke: `_parse_list_arg("cat, dog, person") -> ["cat", "dog", "person"]`, `("cat, , dog") -> ["cat", "dog"]`, `("") -> []`.
- Tests: two new parametrized cases cover the whitespace and blank-drop behavior.
- CI: all 11 checks green; `CLEAN`/`MERGEABLE`, already atop current `main` (no rebase needed).

Issue #2376, opened 2026-07-20
Config robustness: never crash, validate, degrade gracefully


wkentaro:
## Destination

labelme ships a hardened config subsystem: a malformed `~/.labelmerc` or `--config` file never hard-crashes the app; invalid values produce informative per-field errors instead of tracebacks; config is validated against a typed schema; valid settings survive alongside invalid ones per the chosen degradation policy; and every migration is uniformly type-tolerant. Backward-compatible with existing YAML config files. Reaching the end of this map means this is **designed and shipped**, not merely specced.

## Notes

**Execution is carried in-map** (plan + execute): a ticket whose decision is settled may resolve by shipping a PR (via `/make-pr`, CHANGELOG per repo convention, `/review-fix`), not only by recording a decision.

**Domain surface** (`labelme/_config/__init__.py`):
- `load_config()` — merges `default_config.yaml` <- user YAML <- overrides; runs migration then `_update_dict`.
- `_update_dict()` / `_validate_config_item()` — raise `ValueError` for unknown keys, non-mapping sections, bad enums, duplicate labels.
- `_migrate_config_from_file()` — the recurring crash source (e.g. `re.match` on a non-string `ai.default`).
- Call sites: `labelme/__main__.py:327` already wraps the early read in `except Exception`; `labelme/_app.py:1223 _load_config()` only catches `ValueError`, so any other exception hard-crashes before the window opens. Schema source of truth: `labelme/_config/default_config.yaml`.

**Constraint:** stay backward-compatible with existing user YAML config files.

**Interim:** PR #2358 (per-field guard for a non-string `ai.default`) may land now as an interim crash-fix; this overhaul supersedes ad-hoc per-field guards later. Track, don't block.

**Skills:** `/grilling` + `/domain-modeling` for HITL tickets; `/prototype` where "what shape" dominates; `/make-pr` + `/review-fix` for execution.

**Tracker convention:** GitHub, body-convention wayfinding. Membership = label `wayfinder:config`; type = `wayfinder:<type>`; blocking = `Blocked by:` line in each ticket body; claim = assign to wkentaro. See `docs/agents/issue-tracker.md` -> Wayfinding operations.

## Decisions so far

[...]

Issue #2382, opened 2026-07-20
Uniform tolerance policy for config migrations


wkentaro:
## Question

Define a uniform tolerance policy for `_migrate_config_from_file` so no migration step ever throws on an unexpected type/shape (the #2358 class of bug), and settle migrate-vs-validate ordering: do migrations run pre-validation on raw input (and must therefore be defensive), or post-validation on typed data?

Blocked until the pipeline design is settled.

---
Map: #2376
Blocked by: #2381

Issue #2378, opened 2026-07-20
Inventory of every config key, type, and constraint


wkentaro:
## Question

Produce a complete inventory of the config surface as a linked markdown asset: every key in `labelme/_config/default_config.yaml` (including nested sections such as `shortcuts`, `ai`, `canvas`), its default value, expected type, allowed values / constraints (enums like `validate_label`, `shape_color`; cross-field rules like labels-required-when-`validate_label`), and which keys are user-facing vs internal. Feeds the schema-approach decision.

AFK: read the yaml and how each key is consumed across the codebase; link the summary from this issue.

---
Map: #2376
Blocked by: none

wkentaro:
> *This was generated by AI.*

## Config surface inventory

Source of truth: `labelme/_config/default_config.yaml`. Loader: `labelme/_config/__init__.py`. A GUI-exposed subset schema already exists: `labelme/_config/_schema.py` (`SETTINGS`, driving `labelme/_widgets/settings_dialog.py`) — **8 keys** surface as Settings controls; per `README.md:130-138` the *whole* `~/.labelmerc` is a documented public interface.

### Headline finding (drives #2380 / #2381)

Validation is **scattered across three layers**, which is exactly why bad config leaks past `_load_config`'s `except ValueError`:

1. **Load-time** (`_config/__init__.py`) — only `validate_label`, `shape_color`, `labels` (dup check) get bespoke checks; `_update_dict` structurally rejects unknown keys and non-mapping sections; one cross-field rule (`labels` required when `validate_label`).
2. **Construction-time** (widgets `__init__`) — e.g. `canvas.double_click` is validated in `Canvas.__init__` and raises **after** the config is "successfully loaded" → crash before the window opens. Same risk shape as the `ai.default` `TypeError`.
3. **Consumption-time** — `label_colors` and the `shape.*_color` unpacking into `QColor(*...)`; most keys have **no** check at all.

A typed schema (#2380) only closes the gap if it pulls layers 2 and 3 forward to load-time.

### Per-key inventory

| Key | Type | Constraint | Consumed at | Notes |
|---|---|---|---|---|
| `auto_save` | bool | none | `_app.py:333` | `--no-auto-save` |
| `display_label_popup` | bool | none | `_app.py:1891` | GUI (`_schema.py:58`) |
| `with_image_data` | bool | none | `_app.py:339,1803,2255` | runtime-mutated; `--with-image-data` |
[...]

PR #2384, opened 2026-07-20
fix(config): never hard-crash config loading (backstop)

### LINKED ISSUE #2377: Config loading never hard-crashes (backstop)
## Question

Broaden `MainWindow._load_config` (`labelme/_app.py:1223`) so config loading never hard-crashes: catch not just `ValueError` but any exception raised while parsing/merging config, log it at `warning`, then fall back to the same non-modal "Configuration Errors" dialog + defaults path already used for `ValueError`. Mirror the broad `except Exception` the early read in `labelme/__main__.py:327` already uses.

This is defense-in-depth **beneath** the future validator, not a replacement for graceful per-field degradation. Resolve by shipping a PR with a test that a config raising a non-`ValueError` (e.g. the non-string `ai.default` `TypeError`) no longer crashes.

---
Map: #2376
Blocked by: none

wkentaro:
`MainWindow._load_config` caught only `ValueError`, so any other error raised while parsing or merging config crashed the app before the window opened. A malformed `~/.labelmerc` or `--config` file with ungrammatical YAML raises a `yaml` parser error, which slipped past the handler and crashed on startup. This bug shipped in v7.0.4.

Broaden the catch to any exception, log it at `warning`, and reuse the existing non-modal Configuration Errors dialog + defaults fallback, mirroring the broad `except Exception` the early config read in `__main__.py` already uses.

This is the backstop layer from the config-robustness map ([#2376](https://github.com/wkentaro/labelme/issues/2376)): defense-in-depth *beneath* the per-field guards, not a replacement for graceful per-field degradation. It composes with — does not supersede — the interim per-field fix in #2358; the test deliberately triggers the backstop with malformed YAML (which no per-field guard neutralizes) rather than a specific field so it stays valid once #2358 lands.

## Test plan
- [x] Added `tests/e2e/config_test.py::test_MainWindow_config_load_error_falls_back` (malformed YAML raises a non-`ValueError`); it asserts the window opens on defaults instead of crashing
- [x] Full suite: 880 passed; `ruff` + `ty` clean

Closes #2377


wkentaro:

Minimal, sound backstop: broadens `MainWindow._load_config`'s `except ValueError` to `except Exception`, logs at `warning`, and reuses the existing non-modal Configuration Errors dialog + defaults fallback. Mirrors the broad `except Exception` already used by the early config read in `__main__.py`.

- Branch is current with `main` (0 commits behind, merge-base = `main` HEAD); no rebase needed.
- CI green across all 9 platform/Python combos (lint + tests on ubuntu/macos/windows × 3.12/3.13/3.14).
- `logger.warning("Failed to load config: {}", e)` is correct loguru brace-formatting, matching the existing precedent at `_app.py:2829`.
- `except Exception` does not swallow `KeyboardInterrupt`/`SystemExit` (BaseException), so no signal/exit path is masked.
- Dedicated e2e test `test_MainWindow_config_invalid_ai_default_falls_back` reproduces the `TypeError` (red-before/green-after) and verifies the defaults fallback.
- CHANGELOG entry present, `Closes #2377`.

No meaningful review findings; `/review-fix` skipped. Follow-up work (per-field validation / graceful degradation) is already tracked by #2376, #2378–#2382, so no new spin-off issue was filed.

PR #2386, opened 2026-07-21
test(label-list): cover _ItemModel drag-drop reorder semantics

wkentaro:
`_ItemModel` (the model behind the label list) overrides `dropMimeData` to
re-target an internal-move drop that lands on top of a row — inserting after
the row instead of letting Qt nest it as a hidden child — and overrides
`removeRows` to emit `item_dropped`. Neither had any unit or e2e coverage, so a
regression in that row/parent arithmetic (which corrupts label reorder on
drag-drop) would go uncaught.

These tests drive the model directly via `mimeData()` / `dropMimeData()` rather
than simulating a mouse drag, so they are fast and non-flaky:
- drop onto an item inserts it after that row as a sibling (not overwriting or
  nesting it)
- drop in empty space appends to the end
- `removeRows` emits `item_dropped`

## Test plan
- [x] `pytest tests/unit/widgets/label_list_widget_test.py` (3 new tests pass)
- [x] full suite green (885 passed), `ruff check`, `ty check` clean

### DIFF
--- a/tests/unit/widgets/label_list_widget_test.py
+++ b/tests/unit/widgets/label_list_widget_test.py
@@ -12,6 +12,7 @@ from pytestqt.qtbot import QtBot
 from labelme._shape import Shape
 from labelme._widgets.label_list_widget import LabelListWidget
 from labelme._widgets.label_list_widget import LabelListWidgetItem
+from labelme._widgets.label_list_widget import _ItemModel
 from labelme._widgets.label_list_widget import format_shape_label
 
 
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding three tests for `_ItemModel` drag-drop reorder semantics (`dropMimeData` on-item insert-after, empty-space append, and `removeRows` emitting `item_dropped`).

- No rebase needed: `main` gained 0 commits since the merge-base and the changed file is disjoint, so the existing green CI reflects the merged result.
- Cheap `/code-review` gate at medium effort: **clean, no meaningful findings.** Assertions were verified against the real `_ItemModel`/`QStandardItemModel` behavior (empirically confirmed the without-fix case nests the dragged item as a hidden child), the tests drive real Qt machinery end-to-end with no mock/pinning smell, and each expected value traces to the source's row/parent adjustment logic.
- Single clean commit; no `/recommit` reshaping needed. Test-only, so `/verify` skipped.
- CI green across all 11 jobs (macOS/Ubuntu/Windows x 3.12-3.14, plus lint + cla). No push this tick, so the existing green is authoritative.

Ready for maintainer review and merge.

PR #2390, opened 2026-07-21
test(canvas): cover set_last_label trailing-run labeling

wkentaro:
`Canvas.set_last_label` had no direct test. It labels the trailing run of
just-drawn, still-unlabeled shapes after the label dialog is accepted
(`_app.py`'s `_on_new_shape`), stopping the backward scan at the nearest
already-labeled shape. This adds coverage for that boundary and the empty-text
guard.

The trailing-run fixture deliberately places an unlabeled shape *before* the
labeled one so the test discriminates the real `break` from a `continue`
regression: mutating `break` to `continue` makes the test fail (the leading
unlabeled shape gets wrongly relabeled).

## Test plan
- [x] `uv run pytest tests/unit/widgets/canvas_test.py -q` (109 passed)
- [x] mutation check: `break` -> `continue` in `set_last_label` fails the new test

### DIFF
--- a/tests/unit/widgets/canvas_test.py
+++ b/tests/unit/widgets/canvas_test.py
@@ -271,6 +271,46 @@ def test_shape_visibility_survives_backup_and_restore(canvas: Canvas) -> None:
     assert canvas.shapes[0].visible is False
 
 
+def _make_rectangle(label: str | None) -> Shape:
+    return Shape(
+        label=label,
+        shape_type="rectangle",
+        points=np.array([(0, 0), (10, 10)], dtype=np.float64),
+        closed=True,
+    )
+
+
+@pytest.mark.gui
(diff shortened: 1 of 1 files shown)


wkentaro:

**Verdict: recommend-merge**

Own, test-only PR adding coverage for `Canvas.set_last_label`. Reviewed at medium effort: the tests accurately mirror the implementation — the backward scan stops at the nearest labeled shape (so a stale unlabeled shape *before* a labeled one is left untouched, only the trailing unlabeled run is relabeled), flags are applied, and empty text raises `ValueError`. No correctness, simplification, or convention issues found.

- Merge state: `CLEAN`, no rebase needed (main is at the merge-base, 0 commits ahead, no file overlap).
- Single clean commit; no reshaping needed.
- CI: all 11 checks green.

PR #2391, opened 2026-07-21
fix(config): migrate legacy ai_polygon/ai_mask crosshair keys

### LINKED ISSUE #2387: Migrate legacy canvas.crosshair.ai_polygon/ai_mask config keys to ai_points_to_shape
## What to build

A user on v7.0.4 reported (via support email) that Labelme shows a "Configuration Errors" dialog at startup: `Unexpected key in config: ai_polygon`. Their `~/.labelmerc` predates v6.1.0, where the two AI crosshair keys `canvas.crosshair.ai_polygon` and `canvas.crosshair.ai_mask` were consolidated into `canvas.crosshair.ai_points_to_shape`. The config loader rejects unknown keys, so the old keys now trigger the error dialog instead of being migrated like other renamed keys.

Starting Labelme with such a pre-6.1.0 config file must work without the warning dialog: the legacy keys are removed during config loading and folded into `ai_points_to_shape` (logical OR of the two old values), following the existing migration pattern for renamed config keys. If the user's config already sets `ai_points_to_shape` explicitly, that value wins and the legacy keys are just dropped.

A key-name diff of the default config between v6.0.0 and HEAD confirms these two are the only removed keys without a migration, so this closes the gap completely.

## Acceptance criteria

- [ ] A config file with `canvas.crosshair.ai_polygon` and/or `ai_mask` loads without raising or showing the "Unexpected key in config" dialog
- [ ] `ai_points_to_shape` ends up true if either legacy key was true, false if both were false
- [ ] An explicit `ai_points_to_shape` in the same config file is not overridden by the legacy keys
- [ ] A migration log line is emitted, consistent with the existing migrations
- [ ] Unit tests cover the migration following the existing per-migration test pattern

## Blocked by

None — can start immediately.

wkentaro:
A pre-6.1.0 `~/.labelmerc` with `canvas.crosshair.ai_polygon` or `canvas.crosshair.ai_mask` (reported via support on v7.0.4) triggered the "Configuration Errors" dialog at startup with `Unexpected key in config: ai_polygon`. Those two keys were consolidated into `canvas.crosshair.ai_points_to_shape` in v6.1.0, but unlike the other renamed keys they had no migration, so the loader rejected them as unknown.

`_migrate_config_from_file` now folds them into `ai_points_to_shape` (logical OR of the two old values) during config loading, dropping the legacy keys and logging the migration like the existing ones. An explicit `ai_points_to_shape` in the same file takes precedence. A key-name diff of the default config between v6.0.0 and HEAD confirms these were the only removed keys without a migration.

## Test plan
- [x] `uv run pytest tests/unit/_config_test.py -q` (48 passed) — new tests cover the OR truth table, explicit-key-wins precedence, an end-to-end `load_config` from a real YAML file, and the malformed-`crosshair` safety net
- [x] Smoke: `load_config` on a real legacy `.labelmerc` loads without raising, emits the migration log line, and yields `ai_points_to_shape=True`
- [x] `uv run ruff check` + `uv run ty check` clean; full suite `888 passed`

Closes #2387


wkentaro:

Reviewed the diff (`_config/__init__.py` migration block + tests). All 11 CI checks are green against current `main` (the PR is at the merge-base, no rebase needed).

- Correctness pass: clean. The migration pops both legacy keys unconditionally (correct — a short-circuit would leave one un-popped), gates the write on an explicit `ai_points_to_shape` taking precedence, and leaves a malformed `crosshair` section untouched for `_update_dict` to report.
- Test coverage: parametrized OR-semantics, explicit-override precedence, end-to-end `load_config` with a real legacy YAML, and the malformed-section error path.
- One minor idiom note (not blocking): the `logger.info` fires on legacy-key presence rather than being coupled to the write like the sibling migration blocks. Left as-is deliberately — the legacy keys are always removed, so logging their removal is informative, and coupling it to the write would silently drop the keys in the explicit-override case.

PR #2392, opened 2026-07-21
test(brightness-contrast): cover alpha preservation in apply()

wkentaro:
`BrightnessContrastDialog.apply()` re-attaches an image's alpha channel after
brightness/contrast enhancement (it enhances an RGB copy, then restores the
captured alpha via `putalpha`). That RGBA path had no unit coverage — the only
existing exercise is the e2e test, which feeds it an RGB JPEG and never touches
alpha. This locks in the invariant that enhancement never alters alpha, plus the
no-alpha (`Format_RGB888`) path and the slider `valueChanged` -> `apply()` wiring.

## Test plan
- [x] `uv run pytest tests/unit/widgets/brightness_contrast_dialog_test.py` (3 passed)
- [x] full suite: `QT_QPA_PLATFORM=offscreen uv run pytest tests/` (885 passed)
- [x] `uv run ruff format --check`, `uv run ruff check`, `uv run ty check`


wkentaro:

Finalized. Test-only PR adding unit coverage for `BrightnessContrastDialog.apply()`:
- Verified the three tests match the source's actual behavior: the identity-factor RGBA round-trip (`convert("RGB")` → `putalpha` restore), the slider `valueChanged` → `apply()` wiring that brightens RGB while preserving alpha, and the no-alpha `Format_RGB888` path (stride/format correct for both band counts).
- Branch is current with `main` (0 commits behind base), so no rebase.
- `/code-review` (medium): no findings.
- Tests pass locally (`3 passed`); CI green across all 11 matrix jobs.

Verdict: **recommend-merge**.

PR #2394, opened 2026-07-22
test(shape-builders): cover dropping a detection with neither bbox nor mask

wkentaro:
`shapes_from_detections` builds a shape per detection, dropping any whose
geometry cannot be derived. Every output format needs a bbox, a mask, or both:
`rectangle` reads the bbox, `polygon` traces the mask, `mask` requires both, and
`circle`/`oriented_rectangle` derive from the mask with a bbox fallback. A
detection carrying neither is therefore dropped rather than emitted as a
degenerate shape.

Only the `rectangle` slice of that contract was covered
(`test_shapes_from_detections_rectangle_without_bbox_is_dropped`, which still
supplies a mask); the other formats were tested only for the "bbox present, mask
absent" fallback. This test parametrizes over `typing.get_args(AiOutputFormat)`
so the previously-untested "neither present" branch is covered for every format
and a newly added format is automatically held to the same contract, mirroring
how `MASK_REQUIRED_SHAPE_TYPES` is derived in the module itself.

## Test plan
- [x] `pytest tests/unit/_automation/_shape_builders_test.py` (21 passed)
- [x] `ruff check` / `ruff format --check` / `ty check` clean

### DIFF
--- a/tests/unit/_automation/_shape_builders_test.py
+++ b/tests/unit/_automation/_shape_builders_test.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 import math
+import typing
 
 import numpy as np
 import pytest
 from numpy.typing import NDArray
 
 from labelme._automation import MASK_REQUIRED_SHAPE_TYPES
+from labelme._automation import AiOutputFormat
 from labelme._automation import Detection
 from labelme._automation import shapes_from_detections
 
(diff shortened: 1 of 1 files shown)


wkentaro:

Finalized: **recommend-merge**.

- Test-only PR adding one parametrized case: a `Detection()` with neither bbox nor mask is dropped for the mask-deriving `circle`/`oriented_rectangle` shape types (asserts `shapes == []`).
- Matches surrounding test style; `shape_type: AiOutputFormat` annotation is correct (`shapes_from_detections` takes exactly that Literal, and both values are members).
- No rebase needed (branch base is current `main` HEAD, 0 commits behind, merge state CLEAN).
- CI green across all 9 test matrix configs plus lint. No runtime behavior, so no smoke needed.
- Single clean commit; history needs no reshaping.

PR #2397, opened 2026-07-23
refactor(label-list): drop no-op HTMLDelegate.__init__ override

wkentaro:
`HTMLDelegate.__init__` only forwarded `parent` to `super().__init__(parent)`, which `QStyledItemDelegate` already does by inheritance. Removing the override changes nothing: both call sites (`HTMLDelegate()` and `HTMLDelegate(parent=self)`) construct identically against the inherited constructor. Same cleanup as #2339 (dropping `_LoggerIO` overrides that restated `io.StringIO` defaults).

## Test plan
- [x] `uv run pytest tests/` (882 passed)
- [x] `uv run ruff check .` / `uv run ty check` clean on the touched file

### DIFF
--- a/labelme/_widgets/label_list_widget.py
+++ b/labelme/_widgets/label_list_widget.py
@@ -36,9 +36,6 @@ def format_shape_label(shape: Shape, fill_rgb: tuple[int, int, int]) -> str:
 class HTMLDelegate(QtWidgets.QStyledItemDelegate):
     _VERT_FUDGE: Final = 4
 
-    def __init__(self, parent: QtWidgets.QWidget | None = None) -> None:
-        super().__init__(parent)
-
     def paint(
         self,
         painter: QtGui.QPainter,


wkentaro:

**Verdict: recommend-merge**

Reviewed at medium effort. This drops a no-op `HTMLDelegate.__init__` override that only forwarded `parent` to `super().__init__(parent)` — behavior `QStyledItemDelegate` already provides by inheritance. Both call sites (`HTMLDelegate()` in `label_list_widget.py:200` and `HTMLDelegate(parent=self)` in `unique_label_qlist_widget.py:21`) bind identically against the inherited constructor, so the removal is provably behavior-preserving.

- No removed invariant, no caller breakage, nothing to simplify.
- Branch is a single clean commit, current with `main` (0 commits behind base), merge state CLEAN.
- All 11 CI checks green.

No changes pushed this tick; existing green CI is authoritative.

PR #2399, opened 2026-07-23
fix: don't crash on --logger-level fatal (no such loguru level)

wkentaro:
`labelme --logger-level fatal` crashed on startup with an unhandled `ValueError: Level 'FATAL' does not exist`. `fatal` was an advertised `--logger-level` choice, but loguru has no such level, so `_setup_loguru` raised before the window opened for anyone who passed it. Replaced `fatal` with loguru's real `critical` level (matching the existing `QtFatalMsg -> "CRITICAL"` mapping used for Qt logging) and lifted the choices into a single-source `_LOGGER_LEVELS` constant the regression test asserts against.

## Test plan
- [x] `test_logger_level_choice_is_a_valid_loguru_level` — every advertised choice resolves to a real loguru level
- [x] `test_logger_level_rejects_fatal` — `--logger-level fatal` now exits 2 via argparse instead of crashing
- [x] Manual: `labelme --logger-level fatal` -> clean `invalid choice` error (exit 2); `labelme --logger-level critical --version` -> starts, no `ValueError`
- [x] `uv run pytest -q` (887 passed), `ruff check`, `ruff format --check`, `ty check`


wkentaro:

Full-finalize pass on this own PR:

- **No rebase needed** — `main` gained 0 commits since the merge-base; branch is `CLEAN` and current.
- **Review gate** — cheap `/code-review` found no meaningful findings. The fix is correct: loguru has no `FATAL` level, so the advertised `--logger-level fatal` choice always crashed `_setup_loguru` with `ValueError: Level 'FATAL' does not exist`. Replacing it with the real `critical` level (and single-sourcing the choices as `_LOGGER_LEVELS`) is idiomatic (`Final` + `UPPER_CASE`, shared by `main()` and the regression test).
- **Verified end-to-end** — `labelme --logger-level fatal` now exits 2 with a clean argparse error listing `debug, info, warning, error, critical` (no traceback); confirmed the old `FATAL` path raised `ValueError`.
- **CI** — all 11 checks green (lint + test across ubuntu/macos/windows × py3.12/3.13/3.14).

Commit history is a single well-formed commit; no recommit needed. No follow-up issues warranted.

PR #2400, opened 2026-07-23
test(shape-render): cover bounds() across shape types and incomplete shapes

wkentaro:
`bounds()` (aliased `_shape_bounds` in `canvas.py`) computes the bounding box that backs the multi-shape drag-move anchor — the union of every selected shape's box. It was exercised only indirectly, through e2e tests that call `.center()` on its result as a click target; `.center()` is insensitive to width/height errors, so the actual per-shape-type extent math went unpinned.

These tests assert the exact `getRect()` for each `_build_image_path` branch (rectangle, circle radius, oriented_rectangle, polygon, single point) and the point-count guards that must return a null rect mid-draw rather than crash. No production code changes.

## Test plan
- [x] `uv run pytest tests/unit/widgets/_shape_render_test.py -q` (12 passed)
- [x] `uv run pytest tests/unit -q` (735 passed)

### DIFF
--- a/tests/unit/widgets/_shape_render_test.py
+++ b/tests/unit/widgets/_shape_render_test.py
@@ -5,8 +5,10 @@ import pytest
 from PySide6 import QtGui
 
 from labelme._shape import Shape
+from labelme._shape import ShapeType
 from labelme._widgets._shape_render import Palette
 from labelme._widgets._shape_render import ShapeRenderContext
+from labelme._widgets._shape_render import bounds
 from labelme._widgets._shape_render import render_shape
 
 _SIZE = 200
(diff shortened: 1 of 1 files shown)


wkentaro:

Reviewed at medium effort. Pure test addition (57 lines, one file); no production code changes.

- The `bounds()` assertions pin genuine per-shape-type extent math that the existing e2e `.center()` callers left unpinned: rectangle box, circle radius via `‖p0−p1‖` inflated into the ellipse box, oriented_rectangle/polygon path extents, and the single-point zero-size rect. All match `_build_image_path` in the source.
- The incomplete-shape cases exercise each type's own point-count guard (`len==2` for rectangle/circle, `len==4` for oriented_rectangle), asserting a null rect mid-draw rather than a crash — a real regression guard, not a tautology.
- CI green across all 9 ubuntu/macos/windows × 3.12–3.14 combos; lint/format/ty clean; branch up to date with `main` (no rebase needed).

No findings.

PR #2402, opened 2026-07-24
refactor(config): remove dead logger_level config key

wkentaro:
The `logger_level` key in `default_config.yaml` is never read from the loaded config: `__main__.py` pops it out of the CLI overrides, and only the `--logger-level` flag configures loguru. This removes the dead key and adds a migration that drops it from existing `~/.labelmerc` files so they keep loading without an unknown-key error.

## Test plan
- [x] `test_migrate_removes_logger_level` covers a config file that still contains the key
- [x] `uv run pytest tests -x -q` (883 passed), ruff + ty clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Removed the `logger_level` config key, which had no effect: log verbosity has always been controlled solely by the `--logger-level` CLI flag. A `logger_level` entry remaining in an existing `~/.labelmerc` is dropped on load, so old config files keep working unchanged ([#2402](https://github.com/wkentaro/labelme/pull/2402))
+
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


wkentaro:

**Verdict: recommend-merge**

Removes the dead `logger_level` config key and adds a migration that drops it from existing `~/.labelmerc` files so they keep loading without an unknown-key error.

- Verified `logger_level` is never read from the loaded config: `__main__.py:294` pops it from the argparse namespace (populated by the `--logger-level` flag at `:179`), not from the config-file defaults, so removing the YAML key does not affect that pop. `_setup_loguru` is driven solely by `args.logger_level`.
- Migration mirrors the existing `store_data -> with_image_data` pattern and runs before validation, so old configs are cleaned rather than rejected.
- Smoke: default config now has no `logger_level`; an old config containing it loads cleanly (migration drops it, no crash).
- Test mirrors `test_migrate_store_data_to_with_image_data`; CHANGELOG entry present.
- CI green across all platforms; single clean commit.

No rebase needed (branch is at the merge-base with `main`, zero divergence). Code review surfaced no meaningful findings.

PR #2404, opened 2026-07-24
test(label-file): cover the shape JSON codec directly

wkentaro:
The shape JSON codec (`_load_shape_json_obj` / `_dump_shape_to_json_obj` in
`labelme/_label_file.py`) was only exercised indirectly, through file-I/O-heavy
`read_label_file` / `write_label_file` round-trips. The malformed-field error
branches are already covered directly by
`test_read_label_file_raises_on_malformed_shape_field`, but the happy-path parse
contract was not: nothing asserted the optional-field defaults (`description` ->
`""`, `group_id` -> `None`, `flags` -> `{}`), unknown-key bucketing into
`other_data`, the falsy-but-valid `group_id: 0` case, or the mask base64
encode/decode, so a regression in any default would pass the whole suite.

These direct unit tests pin that contract with no filesystem access, building
mask input via the production `img_arr_to_b64` encoder rather than hand-rolled
base64.

## Test plan
- [x] `uv run pytest tests/unit/_label_file_test.py -q` (56 passed)
- [x] `uv run ruff check` / `ruff format --check` / `ty check` on the file

### DIFF
--- a/tests/unit/_label_file_test.py
+++ b/tests/unit/_label_file_test.py
@@ -8,16 +8,20 @@ from typing import Any
 
 import numpy as np
 import pytest
+from numpy.typing import NDArray
 
 from labelme._label_file import Annotation
 from labelme._label_file import LabelFileReadError
 from labelme._label_file import LabelFileWriteError
 from labelme._label_file import ShapeDict
 from labelme._label_file import _check_image_dimensions
+from labelme._label_file import _dump_shape_to_json_obj
+from labelme._label_file import _load_shape_json_obj
 from labelme._label_file import _normalize_to_uint8
 from labelme._label_file import is_label_file_path
 from labelme._label_file import read_label_file
 from labelme._label_file import write_label_file
+from labelme._utils import img_arr_to_b64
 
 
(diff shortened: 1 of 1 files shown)


wkentaro:

Test-only PR adding direct unit coverage of the shape JSON codec (`_load_shape_json_obj` / `_dump_shape_to_json_obj`).

- **Rebase:** skipped. `main` advanced 4 commits since the merge-base, all in files disjoint from this PR (`_config`, its test, `CHANGELOG.md`), and merge state is CLEAN, so the existing green still reflects the merged result.
- **Review (`/code-review`, both axes):** clean. Standards found no hard violations (plain `test_` functions, full `-> None` annotations, fixture-backed, kwargs at call sites). Spec review confirmed every assertion faithfully matches the source behavior, including the non-obvious `group_id=0` falsy-but-not-None branch and the lossless PNG mask round-trip. No passing-but-wrong assertions.
- **CI:** all 11 checks green.

Minor, non-blocking: the codec's error branches (missing/mistyped fields) are exercised only indirectly via the pre-existing `test_read_label_file_raises_on_malformed_shape_field`, not directly against `_load_shape_json_obj`. Not worth blocking on.

PR #2405, opened 2026-07-25
fix: translate the Brightness/Contrast slider labels

wkentaro:
## Summary

- The Brightness/Contrast dialog built its two slider rows in a loop and called `self.tr(title)` on the loop variable. `pyside6-lupdate` only extracts string *literals*, so `"Brightness:"` and `"Contrast:"` were never written into any catalog: both labels rendered in English in all 20 bundled locales, while the dialog's own title (a literal) translated fine.
- The labels are now passed as literals (`self.tr("Brightness:")` / `self.tr("Contrast:")`) via a `_add_slider_row` helper, and are translated for all 20 locales. Each locale's wording is taken from the half of its existing `Brightness/Contrast` title translation, so terminology stays consistent.
- The two rows moved from per-row `QHBoxLayout`s to a single `QGridLayout`, dropping a hardcoded `setFixedWidth(75)` on the label. That width was an English-metrics assumption that only became load-bearing once the labels actually translated: the label column needs 89px for `vi_VN` and 76px for `ja_JP`, so both would have been clipped.
- No other non-literal `tr()` call is affected. `settings_dialog.py` also passes variables to `tr()`, but its sources are marked with `QT_TRANSLATE_NOOP` in `_config/_schema.py`, which is the supported way to do it.

## Test plan

- [x] `test_slider_labels_are_extractable_by_lupdate` runs `pyside6-lupdate` over the module and asserts both strings are extracted. Verified it fails against `origin/main`'s version of the file (extracts only `Brightness/Contrast`) and passes here.
- [x] `test_slider_labels_render_the_installed_translation` installs the real `ja_JP.qm` and asserts the rendered `QLabel` text matches the catalog. Verified it fails against the pre-fix code and catalogs.
- [x] `make test` (893 passed), `make lint`, and `make check_translate` all green.
- [x] Rendered the dialog offscreen in `en_US`/`ja_JP`/`vi_VN`/`de_DE`: labels translate, nothing clips, sliders stay aligned, and dragging each slider updates only its own value label (`1.50` / `0.50`).


wkentaro:

Full finalize pass:

- **Rebase:** not needed. `main` gained 0 commits since the merge-base (`bd3ca1d7`), no file overlap, merge state `CLEAN`. The existing green CI already reflects the merged result, so the branch was left untouched and no push happened this tick.
- **Review gate** (`/code-review`, medium, report-only): clean on both axes, so `/review-fix` was skipped.
  - *Standards* — no hard violations. `_add_slider_row` / `_format_factor` are verb-prefixed, fully annotated, and called with kwargs; the helper has two callers, so the extraction clears the "don't extract trivial single-use helpers" bar. Test file uses plain `test_` functions and mirrors the source layout. CHANGELOG entry is filed under `[Unreleased]` → `### Fixed` with the PR link. Ruff clean. Only judgement call: `_add_slider_row` both mutates the grid and returns the slider, a mild command/query mix — but the return is the single value the caller needs, which is the "interesting return values only" case, so it stands as written.
  - *Spec* — 0 findings. Claims were verified rather than trusted: all 20 `.ts` catalogs carry both `Brightness:` and `Contrast:`, each derived from that locale's existing `Brightness/Contrast` title, none left `unfinished`; `make check_translate` regenerates to a zero diff, so no `.qm` is stale; the only other non-literal `tr()` calls (`settings_dialog.py`) are correctly backed by `QT_TRANSLATE_NOOP` in `_config/_schema.py`; measured label widths (ja_JP 76px, vi_VN 89px) confirm the deleted `setFixedWidth(75)` really would have clipped once the labels translated.
- **Commits:** left as one atomic commit. Splitting the source fix from the regenerated catalogs would produce an intermediate commit where `make check_translate` fails, so the current shape is the correct one.
- **Verify:** rendered the dialog offscreen in `en_US` / `ja_JP` / `vi_VN` / `de_DE`. Labels translate (`明るさ:` / `Độ sáng:` / `Helligkeit:`), each slider updates only its own value label (`1.50` / `0.50`), and `apply()` fires once per change — the per-call closure over `value_label` preserves the isolation the old default-arg capture provided.
- **CI:** 11/11 green against `86a5815`, mergeable `CLEAN`.

ideas: #2409 (repo-wide guard so a future non-literal `tr()` call fails CI instead of silently going untranslated; does not block this PR)

PR #2410, opened 2026-07-25
fix(shape): copy the caller's arrays so a Shape owns its points

wkentaro:
`Shape.__post_init__` normalized `points` / `point_labels` with `np.asarray`, which returns the caller's buffer untouched when it already has the target dtype and layout. `move_vertex` writes into `self.points` in place, so editing a `Shape` silently mutated the array its constructor was handed:

```python
points = np.array([[0.0, 0.0], [10.0, 10.0]], dtype=np.float64)
shape = Shape(shape_type="line", points=points)
shape.move_vertex(i=0, pos=(999.0, 999.0))
points[0]  # -> [999., 999.], never touched by the caller
```

`np.array` copies unconditionally, so a `Shape` now owns its arrays.

No `CHANGELOG.md` entry: the bug is latent, not user-reachable. All three construction sites already hand over a fresh array — `np.array(...)` in `_app.py:2822` and `canvas.py:86`, and a dtype-mismatched `float32`/list in `_shape_builders.py:40`, which forces `asarray` to copy — so no shipped flow could observe it. The extra copy costs microseconds on shapes of 2-4 points (a few hundred at most for a polygon), and `Shape` construction is not a hot loop.

## Test plan
- [x] two regression tests in `tests/unit/_shape_test.py`, both failing before the fix: a shape edit no longer leaks into the caller's `points`, and a caller's edit no longer leaks into `shape.point_labels`
- [x] `make test` (893 passed), including the offscreen e2e suite that drives the real window through shape editing, annotation, and oriented-rectangle drag
- [x] `make lint`

### DIFF
--- a/labelme/_shape.py
+++ b/labelme/_shape.py
@@ -48,8 +48,8 @@ class Shape:
     def __post_init__(self) -> None:
         if self.shape_type not in typing.get_args(ShapeType):
             raise ValueError(f"Unexpected shape_type: {self.shape_type}")
-        self.points = np.asarray(self.points, dtype=np.float64).reshape(-1, 2)
-        self.point_labels = np.asarray(self.point_labels, dtype=np.int_).reshape(-1)
+        self.points = np.array(self.points, dtype=np.float64).reshape(-1, 2)
+        self.point_labels = np.array(self.point_labels, dtype=np.int_).reshape(-1)
         if len(self.point_labels) == 0 and len(self.points) > 0:
             self.point_labels = np.ones(len(self.points), dtype=np.int_)
 
--- a/tests/unit/_shape_test.py
+++ b/tests/unit/_shape_test.py
@@ -348,6 +348,28 @@ def test_move_vertex_replaces_only_target_point() -> None:
     assert shape.points[3] == pytest.approx((0.0, 10.0))
 
 
+def test_move_vertex_does_not_mutate_the_array_given_to_the_constructor() -> None:
+    points = np.array([(0.0, 0.0), (10.0, 0.0)], dtype=np.float64)
+    shape = Shape(shape_type="line", points=points)
+
+    shape.move_vertex(i=0, pos=(999.0, 999.0))
+
+    assert points[0] == pytest.approx((0.0, 0.0))
+
+
+def test_constructor_copies_the_given_point_labels() -> None:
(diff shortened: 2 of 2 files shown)


wkentaro:

No rebase and no push this tick, so the existing checks are authoritative: all 11 green (lint, cla, and the 9-way `test` matrix across ubuntu/macos/windows x 3.12/3.13/3.14, including the offscreen e2e suite). `main` has gained 0 commits since the merge-base, so the green reflects the merged result exactly.

Ran a report-only `/code-review` over the diff on both axes. Both came back clean; `/review-fix` was skipped as unwarranted.

**Standards** — the `np.asarray` -> `np.array` swap is surgical (2 source lines, no adjacent code touched, no new abstraction). Both new tests match the file's conventions: plain `test_` functions, `-> None` annotated, kwargs at the `Shape(...)` call sites. Two nits, neither worth a round-trip: no `# why` comment guarding against a future "optimization" back to `asarray` (the rationale lives in the commit message), and the two tests each build their own line `Shape` rather than sharing a helper.

**Spec** — verified against the PR description rather than an issue (none linked):
- `move_vertex` is confirmed the *only* in-place element write in `_shape.py`. `insert_point`, `remove_point`, `translate`, and module-level `rotate()` all rebind to a fresh array, so the constructor copy closes the one real hole.
- The three cited construction sites are the only production `Shape(...)` sites in `labelme/`; every other is in `tests/`. No missed production site passes a float64 `(N,2)` array, so the "latent, not user-reachable" claim holds and the omitted CHANGELOG entry is justified.
- No remaining aliasing hole on `Shape`'s own surface: `copy()` deep-copies, `_shape_to_dict` goes through `.tolist()`, and the rotation-handle / oriented-rectangle accessors all return freshly computed arrays.
- No scope creep.

One factual slip in the PR description, in the prose only, not the code: the third construction site is `labelme/_automation/_shape_builders.py:40`, not `labelme/_shape_builders.py`. The line number and the float32-forces-a-copy reasoning are both correct.

follow-ups: #2413 — `Canvas._apply_in_place_move` rebinds `original.points = clone.points`, aliasing the two Shapes' buffers and bypassing the very invariant this PR establishes. Harmless today only because the caller discards the clone on the next line. Does not block this merge.

PR #2414, opened 2026-07-25
test(_app): cover _shapes_from_dicts label_flags merge on load

wkentaro:
`_shapes_from_dicts` is the annotation-load path that rebuilds Shapes from saved
ShapeDicts and seeds each Shape's Flags from the `label_flags` config. It had no test,
so a regression in that merge would silently reset every Shape Flag on reopened
Annotation Files — data loss that surfaces long after the change that caused it.

These tests pin the merge contract: keys of a matching pattern default to `False`, saved
Flags win over those defaults, saved Flags outside the config are kept, patterns are
anchored at the label start (`re.match`, matching `LabelDialog`'s behavior), and each
Shape gets its own Flags dict rather than sharing one. A separate test covers the plain
field carry-over (label, shape_type, group_id, description, mask, other_data, and the
float64 points conversion).

No production code changes.

## Test plan
- [x] tests added: `tests/unit/_app_test.py` (10 cases)
- [x] mutation-checked — each of these makes the new tests fail: reversing the merge
      order so defaults overwrite saved Flags, `re.match` -> `re.search`, and seeding
      defaults as `True`
- [x] `make test` (901 passed), `make lint`

### DIFF
--- a/tests/unit/_app_test.py
+++ b/tests/unit/_app_test.py
@@ -4,6 +4,7 @@ import struct
 import zlib
 from collections.abc import Callable
 
+import numpy as np
 import pytest
 from PySide6 import QtCore
 from PySide6 import QtGui
@@ -12,6 +13,7 @@ from PySide6 import QtWidgets
 from labelme import __appname__
 from labelme import _app
 from labelme import _automation
+from labelme._label_file import ShapeDict
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


wkentaro:

- **Rebase**: not needed. The branch sits on the current `main` tip (0 commits gained since the merge-base, no file overlap), merge state `CLEAN`, so the existing green still reflects the merged result. Branch left untouched.
- **CI**: all 11 checks green (lint, cla, and test across ubuntu/macos/windows on 3.12/3.13/3.14).
- **Standards review**: no violations. Imports, annotations (`-> None` on every test), kwargs at call sites, and the `parametrize` + kebab-case `ids` shape all match the neighbouring tests; the module-level `_make_shape_dict` helper matches the existing `_make_*` practice in `tests/unit/_shape_test.py`. No baseline smells worth flagging.
- **Spec review**: the PR description's claims hold. The 10-case count checks out (8 parametrized + 2 standalone), and each named contract has a test that actually discriminates it. Notably `pattern-anchored-at-the-label-start` is a real discriminator: `re.search("cat$", "bigcat")` matches while `re.match` does not, so the case genuinely pins the anchoring. All three claimed mutations were reproduced and each breaks at least one new test.
- **`/recommit`**: no-op. One atomic commit, one file, message already explains the why.
- **`/verify`**: skipped, test-only change with no runtime behavior.

ideas: #2415 (loop-invariant non-`str` label guard in `_shapes_from_dicts` re-warns once per configured pattern; cleanup only, does not block this PR)

PR #2417, opened 2026-07-25
fix(app): report the decode allocation limit instead of "Allowed formats"

### LINKED ISSUE #2388: Misleading "Allowed formats" error when the decoded image exceeds Qt's allocation limit
> *This was generated by AI.*

Labelme shows "The selected image file could not be opened ... Allowed formats: ..." for a supported TIFF file when the decoded image is larger than the Qt allocation limit, and the dialog appears only after minutes of load time. A user reported this via support email for labelme 7.0.4 with a 3.4 GB GeoTIFF (37296 x 49319 pixels, 4 bands, LZW).

To reproduce, create a 37296 x 49319 RGBA TIFF and open it in labelme:

```python
import numpy as np
import tifffile

arr = np.zeros((49319, 37296, 4), dtype=np.uint8)
arr[:, :, 3] = 255
tifffile.imwrite("big.tif", arr, photometric="rgb", compression="zlib", rowsperstrip=1)
```

After a long delay, the "Error opening file / Allowed formats" dialog appears:

```
read_image_file: 36.3s, returned 7.3 MB PNG, peak RSS 7.3 GB
QImageIOHandler: Rejecting image as it exceeds the current allocation limit of 256 megabytes
QImage.fromData: isNull=True
```

The chain has three parts. `read_image_file()` decodes the full image with PIL and re-encodes it to PNG in memory (`labelme/_label_file.py:171-190`), which uses ~7.4 GB and causes the delay. `QtGui.QImage.fromData()` then returns a null image (`labelme/_app.py:2150`), because Qt 6 rejects decodes above `QImageReader.allocationLimit()` (256 MB default). The null image triggers the generic "Allowed formats" dialog (`labelme/_app.py:2153-2163`).

The error message should state that the image is too large, show the image dimensions and the limit, and suggest tiles (`gdal_retile.py`) or a smaller copy. A possible fix is to compare the decoded image size against `QImageReader.allocationLimit()` before the `QImage.fromData()` call and show a specific message. Raising the limit is not a fix: a 7.4 GB QImage is not usable, and the Qt raster engine has a ~32767 px per-side limit while this image is 37296 px wide. Tiled or pyramid rendering for very large images is a separate feature request.

Sanjays2402:
Closes #2388

A supported image whose decode exceeds `QImageReader.allocationLimit()` (256 MB by default) makes `QImage.fromData()` return a null image, which fell through to the generic "Allowed formats" dialog and wrongly implied the file format was unsupported.

`_image_too_large_message()` re-reads the header with `QImageReader` (no decode) to get the pixel dimensions, compares the required allocation against the limit, and returns a message naming both. The generic message is still used when the data is unreadable or fits within the limit.


wkentaro:

Thanks for taking this on. I ran a two-axis review (repo standards + fidelity to #2388) plus the test suite and lint locally. The core approach is right, and I verified the mechanism actually works for the reported case. A few things to address before this can be recommended for merge.

## Verified working

I traced the concern that would have sunk this: `read_image_file()` re-encodes to PNG/JPEG in memory before `_app` ever sees the bytes, so I checked whether `QImageReader(buffer).size()` on the re-encoded blob still reports the true dimensions. It does. Reproduced with a 20000x20000 RGBA PNG under a lowered allocation limit: `QImage.fromData()` returns null while `QImageReader(...).size()` reports `(20000, 20000)`, because PNG/JPEG headers carry the dimensions and Qt's size probe does not allocate the decode buffer. For the reported 37296x49319 GeoTIFF the new function fires and reports accurate numbers (~7016 MB against a 256 MB limit).

Locally: `739 passed` on `tests/unit`, `ruff check` and `ruff format --check` both clean.

## Blocking

**1. The CLA is unsigned.** The `cla` check is the only red one, and it is the reason the test workflow has not run. Please post the comment the bot asked for:

```
```

**2. No CHANGELOG entry.** Per `CLAUDE.md`, user-facing changes go in `CHANGELOG.md` under `## [Unreleased]` in the right subsection with the PR number linked. This changes a user-facing error dialog, so it needs a `### Fixed` entry ending with `([#2417](https://github.com/wkentaro/labelme/pull/2417))`.

## Spec gap worth closing

**3. The ~32767 px per-side raster limit is unhandled.** #2388 explicitly calls this out:

> Raising the limit is not a fix: a 7.4 GB QImage is not usable, and the Qt raster engine has a ~32767 px per-side limit while this image is 37296 px wide.

`_image_too_large_message()` only compares `width * height * 4` against `allocationLimit()`. An image with one side over ~32767 px whose allocation math still fits under the limit produces a null `QImage`, `required_mb <= limit_mb` returns `None`, and the user lands back on the misleading "Allowed formats" message. That is the exact failure this PR exists to fix, just via the other door. It also means the current message can quote an MB limit that, if raised, would still not let the image open, which is the false hope #2388 warns against.

Suggest adding a per-side check alongside the allocation check, with a message that names whichever ceiling was actually hit.

**4. The message drops `gdal_retile.py`.** #2388 asked for it by name:

> The error message should state that the image is too large, show the image dimensions and the limit, and suggest tiles (`gdal_retile.py`) or a smaller copy.

The current text says "split the image into tiles" without naming a tool. Naming it is the difference between advice and an actionable next step.

## Minor

**5. Redundant local annotations.** `limit_mb: int = ...allocationLimit()` and `required_mb: float = ... / 1024 / 1024` are both fully inferable. The repo convention is to annotate signatures, not locals.

**6. The docstring's first line restates the function name.** Sibling module-level functions in `_app.py` (`_format_window_title`, `_resolve_label_path`, `_shape_to_dict`) carry none. The second sentence explaining *why* `None` means "fall back to the generic message" is genuinely useful; the first is not.

Not an issue: `QtCore.QCoreApplication.translate("MainWindow", ...)` is the correct call here. `self.tr()` is unavailable in a module-level function, and `MainWindow.tr()` resolves to the same `"MainWindow"` context, so the translation context is preserved exactly.

Follow-up filed, not blocking this PR: #2418 covers the ~36s / ~7.4 GB eager decode-and-re-encode in `read_image_file()` that happens before this error path is ever reached.

Sanjays2402:
Pushed 1dc918a covering points 2-6.

The per-side gap was real and you're right that it's the same failure through the other door. `_image_too_large_message` now checks `max(width, height) > 32767` before the allocation math and returns a message that says the raster engine's per-side limit was hit and that raising the decode limit won't help, so it can't quote an MB number that would be false hope. The allocation branch is unchanged apart from wording. Both messages now name `gdal_retile.py`. Added a test that stubs `QImageReader` to report 37296x49319 (the dimensions from #2388) and asserts the per-side message, since materializing an image that size isn't practical in the suite.

Also: CHANGELOG entry under `## [Unreleased]` / `### Fixed` with the PR link, dropped the `limit_mb: int` / `required_mb: float` locals, and cut the first docstring line.

`tests/unit` is 740 passed, `ruff check` and `ruff format --check` clean on both touched files.

On the CLA: that's tied to my identity as a legal agreement, so I'd rather sign it deliberately than have it happen as part of an automated pass. I'll post the signature comment myself.

wkentaro:

Re-reviewed after 1dc918a. Good work — points 2 through 6 from the earlier review are all resolved:

- CHANGELOG entry added, correctly formatted and linked.
- The per-side ~32767px raster-engine limit is now checked and reported, with a test pinned to the exact 37296x49319 dimensions from #2388.
- Both messages now name `gdal_retile.py`.
- The redundant local annotations (`limit_mb: int`, `required_mb: float`) are gone.
- The docstring's first line no longer restates the function name.

I ran the suite locally (CI hasn't executed yet, see below): `tests/unit/_app_test.py` is 31 passed including the 4 new tests, and `ruff check` / `ruff format --check` are clean, matching what you reported.

Two things left before this can merge:

**1. CLA still unsigned.** The `cla` check is still red, and because of that the `Test` and `Lint` GitHub Actions workflows on 1dc918a are stuck at `action_required` — they haven't run at all. Please post:

```
```

**2. `ty check` fails on the new test code** (part of `make lint`, so this will fail Lint as soon as it actually runs). Locally:

```
tests/unit/_app_test.py:154  image.save(buffer, "PNG")          -> no-matching-overload (stub wants bytes, not str, for `format`)
tests/unit/_app_test.py:155  bytes(buffer.data())                -> invalid-argument-type (QByteArray not accepted by bytes())
tests/unit/_app_test.py:198  _app.QtGui.QImageReader = _Reader   -> invalid-assignment
```

`ty check` is clean on `main` for this file, so these are new. The third one is an easy fix: this repo suppresses stub gaps with `# ty: ignore[...]`, not `# type: ignore[...]` — `ty` doesn't recognize the latter, which is why it's still flagged despite the comment. See `tests/unit/widgets/label_dialog_characterize_test.py:65` for the same monkeypatch-a-Qt-attribute pattern done with `# ty: ignore[invalid-assignment]`. The other two need either a `# ty: ignore[...]` or a small tweak (`image.save(buffer, b"PNG")` for the format arg).

Non-blocking nitpick: `_RASTER_MAX_SIDE` is a module-level constant used by only one function; sibling single-use constants in this file (`SCORE_FOR_EXISTING_SHAPE`, `FIT_WINDOW_SCROLLBAR_MARGIN`, etc.) are declared locally with `Final`. Not blocking, just for consistency if you're touching this again.

Once the CLA is signed and the `ty` errors are fixed, this looks ready.

Sanjays2402:
Pushed 496df1e for the `ty` errors.

Line 198 was the `# type: ignore[misc]` / `# ty: ignore[...]` mixup you pointed at — switched to `# ty: ignore[invalid-assignment]` per `label_dialog_characterize_test.py:65`. The restore in the `finally` doesn't need one (`ty` flagged it as an unused suppression), so only the assignment carries it.

For `image.save(buffer, "PNG")` I kept the `str` and suppressed with `# ty: ignore[no-matching-overload]` rather than taking `b"PNG"` — PySide6 rejects the bytes form at runtime with `ValueError: 'PySide6.QtGui.QImage.save' called with wrong argument values`, so the stub is wrong, not the call. `bytes(buffer.data())` got `# ty: ignore[invalid-argument-type]`.

Also took the nitpick: `_RASTER_MAX_SIDE` is now a local `RASTER_MAX_SIDE: Final` inside `_image_too_large_message`, matching `SCORE_FOR_EXISTING_SHAPE` and friends. The comment moved with it.

`uv run ty check` clean across the repo, `ruff check` / `ruff format --check` clean, `tests/unit` 740 passed.

CLA is still mine to sign and I'll post it myself.

wkentaro:

Pushed three finalizing commits on top of your branch (thanks for the solid base — the review loop found only refinements, no flaws in the approach):

- **fdf0f1f** `fix(app)`: the allocation estimate now derives bytes/pixel from the reader's decode format instead of assuming 4, so a 16-bit-per-channel image that Qt rejects at 8 bytes/pixel no longer slips past the estimate and falls back to the generic "Allowed formats" message. Also: `math.ceil` for the reported need (a fractional overage could previously render the self-contradictory "needs about 2 MB, but the decode limit is 2 MB"), an explicit unknown-format guard, a verb-led rename to `_make_image_too_large_message`, and the per-side test now drives the real `QImageReader` with a 57-byte header-only PNG claiming 37296x49319 (the #2388 dimensions) instead of stubbing the reader class — which also removed the `ty: ignore`.
- **52be2f1** `test(e2e)`: an integration test that loads a real PNG through `MainWindow._load_file` with a lowered allocation limit and asserts the dialog text, closing the "driving the real image-load path" acceptance criterion from #2388.
- **bd00676** `i18n`: regenerated the translation catalogs and filled the two new strings in all 20 languages (placeholder integrity verified programmatically across every locale).

Locally green: full suite 898 passed, `ruff` and `ty` clean, `make check_translate` would pass (0 unfinished entries).

wkentaro:
> *This was generated by AI.*

I pushed three follow-up commits on top of the original submission; summarizing here since the PR body describes the initial version:

- **25bb211 `fix(app): estimate the decode need from the image's bit depth`** - The allocation estimate now derives bytes/pixel from the reader's decode format instead of assuming 4, so a 16-bit-per-channel image Qt rejects at 8 bytes/pixel no longer slips past the estimate and falls back to the misleading generic "Allowed formats" message. It also bails out when the decode format is unknown (bitsPerPixel 0), and rounds the reported need up with `math.ceil` so a fractional overage can never render the contradictory "needs about N MB, but the decode limit is N MB". The per-side-limit test now drives a real header-only PNG claiming 37296x49319 through `QImageReader` instead of stubbing the reader class, and the allocation-limit save/restore moved into a shared fixture.
- **8c8289b `test(e2e): cover the too-large dialog through the real load path`** - Adds an e2e test that loads a real image file through `_load_file` under a 1 MB allocation limit and asserts the dialog reports the pixel dimensions and the limit rather than "Allowed formats".
- **1bb3ff6 `i18n: translate the image-too-large messages`** - Translates the two new messages for all 19 locales (`.ts` + `.qm`).

PR #2421, opened 2026-07-25
fix: start with the default settings when the default config file is missing

wkentaro:
`get_user_config_file` deliberately tolerates a config file it could not create: it logs a warning and returns the path anyway, so labelme can run on the built-in defaults. `MainWindow._load_config` degrades the same way when reading the file fails. `main` contradicted both — the returned path was the argparse default for `--config`, so a default path that does not exist was treated exactly like an explicit `--config /typo.yaml` and exited with status 1.

The result: with an unwritable `HOME` (a read-only home in a container, a locked-down profile), labelme refused to start at all.

A missing file is now fatal only when the user asked for that file; a missing default path falls back to the built-in defaults with a warning. The branch table moves into `_resolve_config_source`, which keeps `main` thin and makes it testable without a window. The explicit-path branch keeps probing with `os.path.isfile`, since `Path.is_file` raises instead of returning `False` on an over-length value (#2345).

## Test plan
- [x] 6 unit tests over `_resolve_config_source`: missing default → fallback (asserting the warning is emitted), existing default, explicit missing → exit 1, over-length value → exit 1, explicit file, inline YAML string
- [x] Manual repro against a genuinely unwritable `HOME` (`chmod 555`), running the real CLI on `examples/tutorial/apc2016_obj3.jpg`: `main` exits 1 and never starts; this branch opens the window with the image and its annotations loaded
- [x] `make lint` and `make test` (897 passed)


wkentaro:

### Rebase
Skipped. The branch sits on the current `main` tip (0 commits behind the merge-base, merge state `CLEAN`), so the existing green already reflects the merged result. No push this tick, so CI was read as-is rather than re-triggered.

### Review
A report-only `/code-review` pass over the diff on both axes (standards, spec) returned no hard violations and no correctness findings. The extraction of `_resolve_config_source` is a faithful, surgical move of the existing branch table; naming, annotations, kwargs at the call site, test layout, and the CHANGELOG entry all match repo convention. Not escalated to `/review-fix`.

Two nits, neither worth a push:

- On a home where the create attempt failed, the user now sees two warnings for one underlying event: `Failed to save config: ...` from `get_user_config_file`, then `Config file does not exist: ...; using the default settings` here. They say different things (could not create vs. falling back), so the pair is arguably informative.
- The `--config` help text still reads `(default: <path>)` while the argparse default is now `None`. The text remains accurate about behaviour, since an existing `~/.labelmerc` is still read when the flag is omitted; only `parser.get_default("config")` changes, which nothing introspects.

### History
Single commit, conventional message, code + tests + CHANGELOG together. Already the clean sequence; `/recommit` was a no-op.

### Verification
Ran the real CLI on `examples/tutorial/apc2016_obj3.jpg` under `QT_QPA_PLATFORM=offscreen`, against a home directory `chmod 555` with `~/.cache/labelme` pre-created, which is the scenario in the description:

| | `main` @ 8e8e7b95 | this branch |
|---|---|---|
| log line | `ERROR ... Config file does not exist` | `WARNING ... using the default settings` |
| exit | `1`, window never opens | `124` (still running), image loaded |

CI is green across all 11 checks.

### Follow-ups
- #2422 — while reproducing the above from a *fresh* empty read-only home, startup crashed earlier with `PermissionError` at `_setup_loguru`'s unguarded `cache_dir.mkdir(...)`, before any config handling. Pre-existing on `main` and unrelated to this diff; it did not show up in the manual repro because `chmod 555` on an existing home leaves an already-created `~/.cache/labelme` writable. Filed separately, does not block this merge.

PR #2423, opened 2026-07-25
fix(config): drop a superseded polygon shortcut key

wkentaro:
A `~/.labelmerc` carrying both an old polygon-named shortcut and its renamed
replacement (e.g. `edit_polygon` alongside `edit_shape`) is rejected at startup
with `Unexpected key in config: edit_polygon`. The polygon -> shape migration
only renamed an old key when the new key was absent, so in the both-present case
the stale key survived the migration and the merge in `_update_dict` then refused
it. The cost is not limited to that one key: the whole load fails, so the
configuration-error dialog appears and *every* setting in the file is discarded
in favor of the defaults.

The old key is now always removed. An existing new key still wins (unchanged
behavior), and the drop is logged with both values so a user can see which
binding was discarded. This matches the `add_point_to_edge` removal just above
it, which already drops a key the current config no longer knows.

## Test plan
- [x] `tests/unit/_config_test.py`: the both-keys-present migration test is parametrized over all 8 renamed pairs and now asserts the old key is dropped; a new `load_config` test drives the full path that produced the rejection, with a colliding pair and a non-colliding one in the same file
- [x] `make test` (899 passed) and `make lint`
- [x] Manual, `labelme --config` with a file holding `edit_polygon`, `edit_shape` and `delete_polygon`. Before: `Failed to load config: Unexpected key in config: edit_polygon`, config dropped for defaults. After: both migrations log and the config loads.


wkentaro:

**Rebase:** not needed. `main` gained 0 commits since the merge-base (`bd3ca1d7`), so the existing green CI already reflects the merged result. Branch left untouched.

**Review:** `/code-review` (Standards + Spec axes) came back clean, so the full `/review-fix` loop was skipped.

- *Standards* — CHANGELOG entry correctly placed under `[Unreleased]` → `Fixed` with the PR link; the rewritten loop uses guard clauses / early `continue`, matching the neighbouring `add_point_to_edge` and `ai.default` migrations; tests are plain annotated `test_` functions. No violations.
- *Spec* — traced `_migrate_config_from_file` → `_update_dict`: the old key is unconditionally popped, so it never reaches the `Unexpected key in config` raise; `shortcuts[new_key]` is never overwritten on the collision path, so "an existing new key still wins" holds; the parametrization genuinely covers all 8 pairs; the CHANGELOG matches the code.

**Commits:** one logical commit (fix + tests + changelog) with a clear message. `/recommit` was a no-op.

**Verify:** smoke-tested the real startup path with a config carrying `edit_polygon` + `edit_shape` (colliding) and `delete_polygon` (non-colliding).

- On `main`: `ValueError: Unexpected key in config: edit_polygon` — bug reproduced.
- On this branch: both migrations log and the app boots with no configuration-error dialog, `auto_save: true` preserved rather than discarded.

**CI:** green, 11/11 (`cla`, `lint`, and `test` across ubuntu/macos/windows × 3.12/3.13/3.14).

No follow-up or idea issues filed. The one thing worth noting is not actionable: `tests/unit/_config_test.py` keeps its own copy of the `_POLYGON_TO_SHAPE_RENAMES` table, duplicating the one inside `_migrate_config_from_file`. That predates this PR, and since the source table is function-local by design (a constant used in one function), importing it instead would mean widening its scope against convention. The duplication looks like the right tradeoff.

PR #2424, opened 2026-07-26
fix(label-dialog): show one checkbox per flag when label_flags patterns overlap

wkentaro:
Two `label_flags` patterns can legitimately match the same label and both name the same
flag — a catch-all alongside a per-label override, say `{".*": [occluded], "^cat$": [occluded, urgent]}`.
The label dialog built one checkbox per *pattern match* rather than per *flag*, so labeling
a "cat" rendered two checkboxes both reading "occluded". `_collect_flags()` keys its result
by checkbox text, so on OK the last duplicate overwrote the first — ticking the visible
"occluded" box saved `occluded: false`, with no error or other indication the tick was dropped.

`_update_flags` now accumulates into a dict keyed by flag name, so each flag gets exactly one
checkbox. This is the same dedup shape the annotation-loading path (`_shapes_from_dicts`)
already applies to these patterns; only the dialog was missing it.

## Test plan
- [x] Two regression tests in `tests/unit/widgets/label_dialog_characterize_test.py` (one checkbox per flag; the tick survives the popup round-trip). Both fail on `main`, pass here.
- [x] End-to-end smoke through a real `MainWindow` with overlapping patterns: `main` renders `['occluded', 'occluded', 'urgent']` and saves `occluded: False` after ticking it; this branch renders `['occluded', 'urgent']` and saves `occluded: True`.
- [x] `make lint` and the full suite (893 passed)


wkentaro:

**Rebase:** not needed. The branch head sits directly on current `main` (merge-base `bd3ca1d7`, `main` gained 0 commits since), merge state `CLEAN`, so the existing green already reflects the merged result.

**Review:** a report-only `/code-review` pass over `origin/main...HEAD` came back clean on both axes, so the full `/review-fix` loop was skipped.
- *Standards*: no documented-standard breaches. CHANGELOG entry is in `## [Unreleased]` → `### Fixed` with the PR linked, per `CLAUDE.md`; tests match the file's existing no-test-class `qtbot` style. Two judgement calls only, neither actionable: the dict-keyed dedup loop rhymes with the one in `_shapes_from_dicts`, but the two sites build different value shapes across different layers (Qt UI vs annotation loading), and the added 3-line comment explains genuine *why* (the silent overwrite via duplicate checkbox text) rather than restating the code.
- *Spec*: faithful. No missing requirements, no scope creep. The ordering question was checked explicitly — reassigning an existing dict key does not move it, so first-insertion order is preserved and `["occluded", "urgent"]` is the correct render; and `current_states.get(key, False)` makes repeated writes idempotent, so there is no last-wins hazard. The claim that this mirrors the `_shapes_from_dicts` dedup holds.

**History:** already one clean, well-scoped commit — `/recommit` was a no-op. Nothing was pushed this tick.

**Verify:** smoked the real `LabelDialog` offscreen with the overlapping patterns from the description, on both sides.

| | checkboxes rendered | saved after ticking `occluded` |
|---|---|---|
| `main` | `["occluded", "occluded", "urgent"]` | `{"occluded": False, "urgent": False}` |
| this branch | `["occluded", "urgent"]` | `{"occluded": True, "urgent": False}` |

The user's tick is dropped on `main` and survives here, so the reported bug reproduces and this fixes it.

**CI:** all 11 checks green (lint, CLA, and test across ubuntu/macos/windows × Python 3.12/3.13/3.14).

No follow-up issues filed. Merging remains yours.

PR #2425, opened 2026-07-26
fix(main): degrade to stderr-only logging when the log file fails

### LINKED ISSUE #2422: Startup crashes with PermissionError when ~/.cache/labelme cannot be created

Noticed while verifying #2421 (which makes a missing default `~/.labelmerc` non-fatal).

## Problem

`_setup_loguru` creates its log directory unguarded:

```python
# labelme/__main__.py:51-56
if os.name == "nt":
    cache_dir = Path(os.environ["LOCALAPPDATA"]) / "labelme"
else:
    cache_dir = Path("~/.cache/labelme").expanduser()

cache_dir.mkdir(parents=True, exist_ok=True)
```

If that path cannot be created, `mkdir` raises and labelme dies with a traceback before the window opens. This runs very early in `main()`, ahead of all config handling.

## Repro

```console
$ TMPH=$(mktemp -d) && chmod 555 "$TMPH"
$ HOME="$TMPH" labelme examples/tutorial/apc2016_obj3.jpg
...
  File "labelme/__main__.py", line 56, in _setup_loguru
    cache_dir.mkdir(parents=True, exist_ok=True)
PermissionError: [Errno 13] Permission denied: '/tmp/tmp.xxxx/.cache'
```

Reproduced on `main` @ 8e8e7b95. A read-only home directory in a container, or a locked-down profile, hits this.

Note the repro needs a home where `~/.cache` itself is absent or unwritable. `chmod 555` on an *existing* home leaves an already-created `~/.cache/labelme` writable, so the crash does not appear there; that is why #2421's manual repro did not surface it.

## Why this is a bug and not a policy

The surrounding code deliberately tolerates exactly this condition instead of exiting:

- `_config.get_user_config_file` wraps its `touch()` in `try/except`, logs `Failed to save config`, and returns the path anyway.
- `MainWindow._load_config` degrades to defaults when reading the config file fails.
- #2421 extends the same tolerance to a missing default config path.

Logging setup is strictly less essential than config, so hard-crashing there is inconsistent with the established policy. A missing log file should degrade to stderr-only logging, not prevent annotation.

## Suggested direction

Guard the `mkdir` (and the subsequent `logger.add(log_file, ...)`), warn on failure, and continue with the stderr sink that is already installed just above. The Windows branch has a second failure mode worth covering at the same time: `os.environ["LOCALAPPDATA"]` raises `KeyError` when that variable is unset.

Not a blocker for #2421, which is correct on its own terms.

Sanjays2402:
Closes #2422

`_setup_loguru` created the cache directory and registered the rotating file sink unguarded, so a read-only or restricted home aborted startup with a traceback before any window opened — reachable via `OSError` from the `mkdir`, `OSError` from registering the sink when the directory exists but the log file cannot be opened, and `KeyError` on Windows when `LOCALAPPDATA` is unset.

A stderr sink is registered immediately above, so log-file setup is now best-effort: a failure warns through that sink and startup continues, matching the tolerance the config-loading code already applies to strictly more essential state.

Tests in `tests/unit/__main___test.py` cover all three failure modes (3 failed before the fix, pass after); `tests/unit` is 739 passed.


wkentaro:

Reviewed the diff against `main` (checked out the PR commit in an isolated worktree). The fix itself looks correct and matches #2422's brief closely:

- `ruff check` and `ty check` on the changed files are clean.
- `pytest tests/unit` passes (739 passed), including the three new failure-mode tests.
- Confirmed the three tests genuinely exercise the `except (KeyError, OSError)` branch — loguru's file sink opens synchronously at `logger.add()` time even with `enqueue=True`, so an unwritable path raises immediately rather than being silently swallowed.
- `except (KeyError, OSError)` is appropriately narrow, not a blanket `except Exception`.
- CHANGELOG entry is correctly placed and linked.

Two process items block merge as-is:

1. **CLA**: the `cla` check is failing ("Committers of Pull Request number 2425 have to sign the CLA"). Could you sign it?
2. **Rebase**: the branch currently conflicts with `main` — a trivial adjacent-line conflict in `CHANGELOG.md` against the already-merged #2429 (both append a `Fixed` entry at the end of the list). A rebase should resolve it cleanly; `labelme/__main__.py` and the test file merge without conflict.

Two small non-blocking suggestions, no action required:

- `labelme/__main__.py`: the warning uses an f-string (`logger.warning(f"...: {e}")`), whereas the sibling pattern this PR is modeled on (`labelme/_config/__init__.py`, `get_user_config_file`) uses loguru's native `{}`-placeholder style (`logger.warning("Failed to save config: {!r}", ...)`). Matching that would be more consistent.
- The three new tests assert only that `_setup_loguru` doesn't raise; they don't capture the log to confirm the warning text is actually emitted (an explicit acceptance criterion in #2422). Since I verified this manually already fires correctly, this is just a coverage nice-to-have, e.g. via `caplog`.

Once the CLA is signed and the branch is rebased, this looks ready to merge.

Sanjays2402:
Rebased onto main — the CHANGELOG conflict with #2429 was just the adjacent-line append, both entries are kept. `labelme/__main__.py` and the test file replayed without conflict.

Also took both suggestions:

- the warning now uses loguru's placeholder style, matching `get_user_config_file`: `logger.warning("Failed to set up the log file, logging to stderr only: {!r}", e)`
- the three tests now assert the warning actually reaches stderr via `capsys`, not just that `_setup_loguru` doesn't raise. Checked they're non-vacuous: replacing the `logger.warning` call with `pass` makes all three fail.

`pytest tests/unit` is 775 passed on f8924a4.

The CLA I'll leave to a separate comment.

Sanjays2402:

wkentaro:
> *This was generated by AI.*

Pushed b747498e to this branch (maintainer edit): `test_setup_loguru_degrades_to_stderr_when_the_log_file_cannot_be_opened` steered path resolution by setting `HOME`, but `ntpath.expanduser` ignores `HOME` (since Python 3.8), so on Windows the test would resolve the real `%USERPROFILE%`, register a file sink on the real log file, and fail its assertions. CI only runs pytest on ubuntu, so this never showed up there. The test now patches `Path.expanduser` to return the tmp cache dir directly, the same approach the home-unknown test already uses; assertions are unchanged. All 28 tests in `tests/unit/__main___test.py` pass.

wkentaro:

Reviewed for the milestone finalize pass: the change matches the spec in #2422 (best-effort log-file setup, stderr sink stays authoritative), the broad catch is justified by the non-OSError failure modes it names (`RuntimeError` from `expanduser`, `KeyError` from `LOCALAPPDATA`), and the four degradation tests plus the full unit file pass locally on a merge with the latest main. CI is green across the matrix and the branch is mergeable. Recommending merge.

PR #2429, opened 2026-07-29
fix(canvas): paint the committed shape before the label dialog opens

wkentaro:
An AI-Box annotation opens the label dialog over a canvas that still shows the
drag bbox, not the segmentation it is asking you to name.

`_reset_after_shape_creation` emits `new_shape`, whose handler blocks in
`_label_dialog.popup()` -> `exec()`, before invalidating the canvas, so the
committed shape is only scheduled for repaint after the dialog closes.

Three non-obvious bits:

1. It presents as intermittent rather than always broken. If a paint request
   from the drag is still pending when the second click lands, the dialog's own
   nested event loop flushes it, and because `self.shapes` was already extended
   it happens to render the new shape. When nothing is pending, which is the
   normal case once the drag repaints have drained, the dialog opens over the
   previous frame. Stack traces confirmed both paths, and this is why the
   existing GUI tests never caught it.

2. AI-Box is where this is most visible. Its last painted frame is the drag bbox
   because `_build_preview_shape` returns `None`, so the segmentation result is
   missing entirely. Point mode can also finalize before any matching preview is
   painted. Other create modes leave a preview that already matches the
   committed geometry.

3. `repaint()` is required rather than hoisting the existing `update()` above
   the emit. `update()` only posts an event, so it would still depend on the
   modal loop delivering it, which is the same mechanism that masks the bug
   today. The trailing `update()` stays because the handler mutates shapes
   afterwards by applying labels on accept or calling `undo_last_line()` on
   cancel.

## Test plan

- [x] `tests/unit/widgets/canvas_test.py::test_finalize_paints_new_shape_before_notifying`
      asserts that the shape is painted by the time `new_shape` fires; verified
      for point mode and both AI create modes
- [x] Screenshotted the running app while the dialog was open, before and after:
      bbox only before, segmentation polygon after
- [x] `make lint`
- [x] Full suite: 913 tests passed


wkentaro:
confirmed this only happens on linux (not on macOS)

PR #2434, opened 2026-08-01
fix: keep focus in the File List while navigating with arrow keys

### LINKED ISSUE #2433: File List loses keyboard focus after each load, so arrow keys navigate only one file
## Summary

In the File List dock, pressing Down/Up navigates to the next/previous file exactly **once**. The second arrow keypress does nothing, because loading the image steals keyboard focus away from the file list and gives it to the canvas.

Walking a directory with the arrow keys is the natural interaction for a file list, and today it is impossible. `A`/`D` work, but they are not what a user reaches for when their cursor is in a list.

## Reproduce

1. `labelme <a directory with several images>`
2. Click an item in the File List (or otherwise focus it).
3. Press Down. -> Selection moves to the next file and the image loads. Correct.
4. Press Down again. -> Nothing happens.

## Cause

`MainWindow._load_file` ends with an unconditional `self._canvas_widgets.canvas.setFocus()` (`labelme/_app.py:2189`, introduced in bb021c7d, 2021).

The chain is: arrow key -> `QListWidget` moves the current row -> `itemSelectionChanged` -> `_file_list_item_selection_changed` -> `_load_file` -> `canvas.setFocus()`. So focus lands on the canvas while the user is still holding the keyboard over the list, and the next arrow key is delivered to the canvas instead (where in edit mode it nudges a selected shape, or does nothing).

## Expected

An image load should not move keyboard focus out of the File List when the File List is what is driving the navigation. Focusing the canvas on load is still correct when the load originates elsewhere (opening a file/directory, `A`/`D` shortcuts while the canvas has focus), since the canvas is where the user then works.

wkentaro:
Arrow keys in the File List navigate exactly one file, then stop. `_load_file` ended with an unconditional `canvas.setFocus()`, so the chain `Down -> itemSelectionChanged -> _load_file -> canvas.setFocus()` pulled focus off the list while the user was still driving it from the keyboard, and the next arrow key was delivered to the canvas instead. Walking a directory with the arrow keys is the natural interaction for a file list, and it was impossible; `A`/`D` are not what a user reaches for when their cursor is in a list.

The policy is now: **a load never pulls the keyboard out of the File List, whatever drove the load.** That covers the arrow-key walk, a mouse click on an item, and `A`/`D` pressed while the list holds focus (those are window-context `QAction`s, so they fire either way and keep working). Loading from anywhere else — opening a file or directory, or `A`/`D` while the canvas has focus — still hands the canvas the keyboard exactly as before.

Keeping focus on the list costs nothing: every canvas shortcut is a window-context `QAction`. The only focus-dependent keys are `Canvas.keyPressEvent`'s raw arrow/Escape/Return handling, which all require a canvas click to select or draw something first.

## Test plan
- [x] `test_arrow_keys_walk_file_list_across_loads` drives two consecutive `Down` presses through `QApplication.focusWidget()` rather than at the widget directly — that indirection is what makes it catch the focus theft; it times out at the second press on `main`
- [x] `test_canvas_takes_focus_when_file_list_has_none` starts focus in the search box so the canvas can only end up focused by the load handing it over; verified by mutation (deleting the `setFocus()` call makes it fail)
- [x] `uv run pytest tests/` — 915 passed
- [x] `uv run ruff check` / `ruff format --check` / `uv run ty check`

Closes #2433


wkentaro:

Finalized this tick and endorsing for merge.

- **/review-fix**: five parallel reviewers (code-review on high effort, simplify, brooks-review, review, exemplar-review) over the diff vs `main` — zero kept findings in a clean first round. One reviewer independently mutation-tested the fix (reverted `_app.py`, confirmed `test_arrow_keys_walk_file_list_across_loads` fails on the pre-fix code, restored) and ran the full e2e suite green (146 passed).
- Dropped as not worth churn: an optional suggestion to also drive the literal `A`/`D` shortcut dispatch in the canvas-focus test (the fallback branch is already covered via `_open_next_image()`, and all three `_load_file` call paths were traced to confirm the guard's scope), and an informational note about the `setFocus()`/`waitUntil` focus-seeding pattern being new to `tests/e2e/` (ran clean and fast offscreen).
- **/recommit**: history is already a clean two-commit sequence (fix+tests, then the changelog note — both styles exist on `main`); no reshaping, no push, tree untouched.
- **Checks**: `ruff check`, `ruff format --check`, `ty check`, and the new tests all pass locally; all 12 CI checks green; merge state CLEAN; branch is based on current `main`.

Verdict: `recommend-merge`.

PR #2435, opened 2026-08-01
fix: read annotation list label color from the state's palette group

wkentaro:
Selecting a shape on the canvas left its label in the Annotation List unreadable — white text on a pale gray row ([discussion #2432](https://github.com/wkentaro/labelme/discussions/2432)).

reported symptom

`HTMLDelegate.paint` hardcoded `QPalette.ColorGroup.Active` when picking the text color, but the row background is painted one line above by `drawControl(CE_ItemViewItem)`, which derives its color group from `opt.state`: `Disabled` when not enabled, `Inactive` when `State_Active` is clear, else `Active`. `QAbstractItemView` clears `State_Active` whenever the view does not hold focus — exactly the case when the selection came from a canvas click. So the row was filled from `Inactive/Highlight` (pale gray) while the label kept being drawn in `Active/HighlightedText` (white).

The File List sits in the same window and stays legible throughout, because it uses Qt's stock delegate and follows that rule. This change makes the Annotation List follow it too, using the same derivation Qt uses in `QItemDelegate::drawDisplay`, so the two docks now match in every focus state.

Not platform-specific despite the report: it reproduces on macOS too, and depends on focus rather than OS. Platforms only differ in *how* visible it is, because their palettes disagree on `Inactive/Highlight` (`#f0f0f0` in the reporter's screenshot, `#d4d4d4` on macOS).

The fix also routes the `Disabled` group and the unselected `Text` role through the same derivation, not just the reported selected-and-unfocused case — all four were wrong for the same reason, and 4 of the 6 parametrized cases below fail on `main`.

## Test plan
- [x] `tests/unit/widgets/label_list_widget_test.py::test_html_delegate_takes_text_color_from_the_state_color_group` — 6 cases covering the full `ColorGroup` × `ColorRole` matrix; 4 fail on `main`, all pass here
- [x] Offscreen render harness comparing the two docks: the Annotation List now matches the File List pixel-for-pixel in both focus states (unfocused `bg=#f0f0f0 ink=#000000`, contrast 1.14 → 18.43)
- [x] `make lint`
- [x] `uv run pytest tests/unit tests/e2e` (775 + 144 green)


wkentaro:
Closed via https://github.com/wkentaro/labelme/pull/2436

PR #2448, opened 2026-08-05
fix: guard incompatible AI prompts at runtime

wkentaro:
AI-Points Prompt Compatibility was enforced during mode navigation. This coupled the AI Assist mode to the model Setting and did not protect runtime states introduced by startup, restored state, Settings changes, or direct state changes.

This change lets users select AI-Points without changing the model Setting. When the user clicks the canvas with an incompatible SAM3 plus AI-Points state, the canvas rejects the point prompt before model download or inference and shows a compatibility warning. The Model Session guard remains as defense for non-UI callers. Returning to AI-Box keeps SAM3 selected.

## Test plan

- [x] 158 AI Assist, canvas, and annotation tests
- [x] Exact CI `make lint` command
- [x] Interface anti-pattern detector


wkentaro:

## Agent Brief

**Category:** bug
**Summary:** Reject incompatible AI-Points clicks after the selected model changes during an active point prompt

**Current behavior:**
The PR checks Prompt Compatibility only when the first AI-Points point starts a draft. If a compatible model starts the draft and a direct state change then selects Sam3, the next canvas click extends the draft without emitting the compatibility warning. The Model Session guard can then surface the unsupported prompt as a generic inference failure.

**Desired behavior:**
Every canvas click that would add an AI-Points prompt point must check the current model's Prompt Compatibility before it mutates the draft, downloads a model, or runs inference. An incompatible click must keep the selected model Setting and active mode unchanged, preserve the existing draft, and emit the existing compatibility rejection signal.

**Key interfaces:**

- The canvas pointer-press boundary must guard both new and existing AI-Points drafts.
- The existing point-prompt rejection signal remains the UI notification contract.
- The Model Session Prompt Compatibility guard remains as defense for non-UI callers.

**Acceptance criteria:**

- [ ] An incompatible first AI-Points click is rejected before model download or inference.
- [ ] If an AI-Points draft starts with a compatible model and the canvas model then changes to Sam3, the next click emits the compatibility rejection signal.
- [ ] That rejected click does not add a point, download a model, or run inference.
- [ ] The model Setting, active AI-Points mode, and existing draft stay unchanged after rejection.
- [ ] Compatible AI-Points prompts and AI-Box behavior remain unchanged.

**Out of scope:**

- Changing the selected model Setting during mode navigation or rejection.
- Removing proactive disabling of incompatible model choices in AI-Points Setting Controls.
- Removing the Model Session defense guard.

Issue #2354, opened 2026-07-16
AI Assist: surface predictions that match existing shapes


wkentaro:
> *This was generated by AI.*

## Problem

AI Assist can return a valid prediction that overlaps an existing Shape. If the prediction is suppressed without visual feedback, a successful point or box interaction appears to do nothing.

## Decision

Existing-Shape suppression is optional and disabled by default.

- When suppression is disabled, each selected prediction remains a new Shape, including predictions that overlap existing annotations.
- When suppression is enabled, a matching prediction highlights the existing Shape in amber instead of creating a duplicate.
- The Setting is `ai.suppress_existing_shape_matches` in the Settings dialog.
- The Setting applies only to AI Assist. AI Text Prompt behavior does not change.

## Expected behavior

- Every successful AI Assist interaction produces visible canvas feedback.
- AI Points shows one candidate selected by prompt-point satisfaction, then model score.
- AI Box processes all predictions remaining after greedy NMS.
- Hidden matching Shapes are temporarily visible in the highlight layer when suppression is enabled.
- Existing-only results do not open the label dialog and do not show the no-Shapes warning.
- Mixed results add new Shapes and highlight existing matches.
- Highlights persist until the next canvas interaction, mode change, image change, or Setting change.

## Matching rule

When suppression is enabled, choose the qualifying existing Shape with the highest mask IoU. A match qualifies when mask IoU is at least 0.5 or containment is at least 0.85.

## Scope

The suppression behavior is configurable. The match thresholds and highlight style are fixed.

PR #2463, opened 2026-08-06
fix(ai-assist): distinguish all-matched sweeps

### LINKED ISSUE #2355: AI Assist: distinct status feedback when a Sweep's proposals all match existing annotations
> *This was generated by AI.*

## What to build

Tell the user when an AI Assist Sweep found only already-annotated objects, instead of implying the model failed.

After #2354, the all-suppressed case can only occur for Sweeps (SAM3 box prompts): the user boxes a region, the model proposes Shapes, and every proposal matches an existing annotation. Today that shows the same generic "AI inference produced no new annotation." status message as a model that returned nothing, so a correctly-working sweep reads as a failure.

Make the proposal call return a small frozen result carrying the proposed Shapes plus the count of detections suppressed as duplicates of existing Shapes (replacing the bare list return). The canvas keeps the existing no-arg "inference produced no shapes" signal for the model-returned-nothing case and gains a second no-arg signal for the proposals-existed-but-all-matched case; the main window connects each signal to its own translated status-bar message (same 5-second status style). New message text: "AI proposals all matched existing annotations."

Regenerate translation catalogs for the new source string the usual way; source-language fallback is acceptable initially.

## Acceptance criteria

- [ ] The AI Assist proposal call returns proposed Shapes plus the suppressed-as-duplicate count; the count is zero when the model returns no annotations
- [ ] When a Sweep returns proposals and all are suppressed as duplicates, the new signal fires and the status bar shows "AI proposals all matched existing annotations."
- [ ] When the model returns no annotations, the existing generic signal and message still fire; the new message never appears for point prompts
- [ ] The in-progress drawing state is cancelled the same way in both empty-result cases
- [ ] Unit tests at the AI Assist session seam assert the returned count; canvas-level tests with monkeypatched inference assert which signal fires (following the existing finalize signal-capture tests)
- [ ] Translation catalogs regenerated; CHANGELOG entry under `[Unreleased]` with the PR number linked

## Blocked by

- #2354

wkentaro:
> *This was generated by AI.*

A successful Sweep currently looks like empty inference when Existing Shape Suppression filters every proposal. This gives it distinct status feedback without changing point prompts or non-SAM3 box behavior.

The all-matched fact is computed before Shape conversion, so an unmatched invalid detection cannot be misreported.

Closes #2355

## Test plan

- [x] Session and canvas tests cover empty, all-matched, mixed, point-prompt, non-Sweep, and invalid-detection results.
- [x] `make test` (1,169 passed)
- [x] `make lint`
- [x] `make check_translate`


wkentaro:
> *This was generated by AI.*

Closing as redundant with #2354. Current `main` already highlights every matching Shape, including hidden Shapes, keeps the highlight until the next canvas interaction, and does not show the empty-inference message for an existing-only result. The separate status message would duplicate that feedback.

PR #2465, opened 2026-08-07
test: make widget exposure waits effective

### LINKED ISSUE #2431: tests: bare qtbot.waitExposed(...) calls never wait

## Problem

Sixteen tests call `qtbot.waitExposed(widget)` as a bare statement after `widget.show()`:

- 6 Label dialog screen-fit tests
- 10 Settings dialog tests

In pytest-qt 4.x, `waitExposed` returns a context manager. A bare call constructs and discards that context manager, so the test does not wait and no error is raised. These tests then inspect geometry, scrolling, navigation, or other state that can depend on the window being mapped.

The tests pass today, but they can be flaky on platforms where window mapping is asynchronous.

## Agent Brief

**Category:** bug

**Summary:** Make every Qt widget-exposure wait in the test suite perform a real wait.

**Current behavior:**

Sixteen calls use this ineffective sequence:

```python
widget.show()
qtbot.waitExposed(widget)
```

Other tests already use pytest-qt correctly:

```python
with qtbot.waitExposed(widget):
    widget.show()
```

**Desired behavior:**

Every test that waits for a widget to become exposed must enter the context manager before it shows the widget. Existing correct context-manager calls must stay unchanged.

Add a repo-wide regression check that fails when `qtbot.waitExposed(...)` is used as a bare expression statement. The check must accept calls used as context managers.

If a converted test exposes a synchronization problem, fix that test without weakening its UI assertion or adding an arbitrary sleep.

**Key interfaces:**

- pytest-qt `waitExposed` context manager, with the widget shown inside the context
- The repository test guard for distinguishing bare expression calls from context-managed calls

**Acceptance criteria:**

- [ ] All 16 current bare `qtbot.waitExposed(...)` calls perform a real context-managed wait.
- [ ] A repo-wide test rejects a new bare `qtbot.waitExposed(...)` expression.
- [ ] Existing context-managed calls remain valid.
- [ ] The affected Label dialog and Settings dialog tests pass.
- [ ] The full test suite passes without new arbitrary sleeps.

**Out of scope:**

- Production UI behavior changes
- Replacing pytest-qt or changing how tests wait for conditions other than widget exposure
- Unrelated test timing cleanup

wkentaro:
Closes #2431.

Bare `qtbot.waitExposed(...)` calls only build the context manager and never wait, so 16 exposure waits were no-ops. They now use the context-managed form around `show()`, and a new AST-based repository test rejects any future bare call.

## Test plan
- [x] New repository test scans tests/ for bare `waitExposed` calls
- [x] `make lint` and `make test` (1165 passed)

### DIFF
--- a/tests/unit/widgets/label_dialog/screen_fit_test.py
+++ b/tests/unit/widgets/label_dialog/screen_fit_test.py
@@ -24,8 +24,8 @@ def test_flag_spacing_is_identical_with_and_without_scrollbar(qtbot: QtBot) -> N
     few = _make_dialog(qtbot, flag_count=3)
     many = _make_dialog(qtbot, flag_count=12)
     for dialog in (few, many):
-        dialog.show()
-        qtbot.waitExposed(dialog)
+        with qtbot.waitExposed(dialog):
+            dialog.show()
 
     assert few._flags_scroll.verticalScrollBar().maximum() == 0
     assert many._flags_scroll.verticalScrollBar().maximum() > 0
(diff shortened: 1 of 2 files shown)


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) found one portability defect, fixed and folded in: the AST scanner now decodes test sources as UTF-8 instead of the locale default. Lint and the full suite pass locally; all checks green.

Issue #1284, opened 2023-05-26
segment_anything.py中对图像做归一化的部分是不是顺序反了？


qjr1997:
### Provide environment information

Mac

### What OS are you using?

macOS

### Describe the Bug

    x = (x - np.array([123.675, 116.28, 103.53], dtype=np.float32)) / np.array(
        [58.395, 57.12, 57.375], dtype=np.float32
    )
    x = np.pad(
        x,
        (
            (0, image_size - x.shape[0]),
            (0, image_size - x.shape[1]),
            (0, 0),
        ),
    )
    x = x.transpose(2, 0, 1)[None, :, :, :]

x的顺序起初是bgr，归一化时候那些数字顺序是rgb，错误顺序归一化后再transpose会不会有影响

PR #2466, opened 2026-08-07
fix: pass RGB images to AI models

### LINKED ISSUE #2356: AI inference receives BGR instead of RGB (red/blue channels swapped for SAM / YOLO-World / SAM3)

## Summary

Every AI feature (AI-Points, AI-Box, AI Text-to-Annotation) feeds a **BGR** image to the underlying vision model, not the RGB the models expect. The red and blue channels are silently swapped for the whole inference pipeline (SAM, EfficientSAM, YOLO-World, SAM3).

## Root cause

`img_qt_to_arr` (`labelme/_utils/image.py:52-60`) returns the QImage's raw memory. For the 32-bit formats Qt uses when loading a PNG/JPEG (`Format_RGB32` / `Format_ARGB32`), that raw layout is **BGRA** on little-endian, because Qt packs each pixel as `0xAARRGGBB` and little-endian byte order puts blue first. The two AI call sites then take `[:, :, :3]`, which yields **B, G, R** and hands it to the model as if it were RGB.

Affected call sites:

- `labelme/_widgets/canvas.py:412-414` (`_shapes_from_ai_points`, drives AI-Points and AI-Box)
- `labelme/_app.py:1354` (`_submit_ai_prompt` -> `get_bboxes_from_texts`, drives AI Text-to-Annotation)

The image flows straight into `osam` (`OsamSession._get_or_compute_embedding` -> `model.encode_image`) and the text-detection path with **no compensating channel swap** anywhere in `labelme/_automation/`, so the swap reaches the model uncorrected.

Not affected: `labelme/_widgets/canvas.py:1808` (`load_pixmap`) also calls `img_qt_to_arr`, but only to hash the bytes, so channel order is irrelevant there.

## Reproduction (confirmed empirically)

Real load path `QImage.fromData -> QPixmap.fromImage -> toImage() -> img_qt_to_arr`, on `examples/tutorial/apc2016_obj3.jpg`, at a strongly-colored pixel:

```
QImage format      = Format_RGB32
PIL ground-truth RGB = [227  35  46]
img_qt_to_arr raw    = [ 46  35 227 255]   # BGRA
arr[:, :, :3]->model = [ 46  35 227]        # BGR, not RGB
matches RGB?           False
matches reversed BGR?  True
```

## Why this needs a human call (not auto-shipped)

The diagnosis is unambiguous, but the fix is a behavioral change to the core AI pipeline, so it wants maintainer sign-off:

1. **Blast radius**: it silently changes the output of every AI inference for all users. Channel order is cheaply testable, but whether the corrected color actually improves real-world segmentation/detection quality is not provable by a cheap unit test.
2. **Contract coordination**: in-flight PR #2328 deliberately documents the BGRA layout as `img_qt_to_arr`'s intended contract (characterization test `test_img_qt_to_arr_returns_bgra_pixels`). So the fix should go at the two callers, not in `img_qt_to_arr`. That intersection is worth a conscious decision.
3. **Prior history**: closed issue #1284 (2023) already asked whether the channel order into segment_anything was reversed. There may be context on why it was closed.

## Options / tradeoffs

- **(A) Fix at the two callers** (recommended): convert to true RGB before extracting the array, e.g. `.convertToFormat(QtGui.QImage.Format.Format_RGB888)`, then drop the now-redundant `[:, :, :3]`. Surgical, leaves `img_qt_to_arr`'s BGRA contract (and PR #2328) untouched.
- **(B) Add an RGB-returning helper** (e.g. `img_qt_to_rgb_arr`) used by the AI call sites, keeping the raw `img_qt_to_arr` for the hashing caller. Slightly more code, clearer intent at the call site.
- **(C) Change `img_qt_to_arr` to always return RGB**: collides with PR #2328 and any other raw-layout caller; not recommended.

## Verification

Cheap and additive: stub `propose_shapes` / `get_bboxes_from_texts` (existing tests already stub the AI path in `tests/unit/widgets/canvas_test.py`), load a pixmap of a known solid color, and assert the captured `image` kwarg's pixel equals true RGB rather than the swapped value.

wkentaro:
Closes #2356.

Every AI inference path (SAM, EfficientSAM, YOLO-World, SAM3) ran on BGR because both call sites sliced `[:, :, :3]` from the raw memory Qt keeps for its 32-bit image formats, which is BGRA on little-endian. The new `img_qt_to_rgb_arr` converts to RGB888 (byte-order defined on every platform) before extracting the array, so the fix also holds on big-endian, where the old slice would have produced ARG.

`img_qt_to_arr` itself is untouched: its raw layout is a pinned contract (#2328), and its remaining callers hash bytes or need alpha, where channel order is irrelevant.

Prior report #1284 (2023) asked the same channel-order question but was closed by its reporter as a misread, so the swap survived until now.

## Test plan

- [x] Unit regression tests for the channel order reaching both inference call sites and for the new helper
- [x] `make lint`
- [x] `make test` (1275 passed, 6 skipped)
- [x] Network e2e via CI


wkentaro:

CI fix attempt: the network-test failure was a latent bug this PR exposed, not a regression of the RGB change. `compute_oriented_rectangle_from_mask` fits a minimum-area rotated rectangle whose corners can overhang the image edges, and the RGB-corrected mask shifted the fit past the boundary (x=339.7 in a 338px image), violating the saved-annotation bounds contract. A new commit keeps AI oriented rectangles inside image bounds: rigid translation when the rectangle fits, coordinate clipping only for an oversized axis, with a unit regression test. The exact failing e2e case now passes locally against the real cached model. Awaiting fresh CI.

wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) found only the CHANGELOG link convention on the RGB change itself. The one red CI run exposed a latent bug (AI oriented rectangles never bounds-clamped), fixed here in its own commit with a regression test; the previously failing network e2e passes against the real model. All checks green.

wkentaro:

Rebased onto `main` (post-base overlap in `CHANGELOG.md` and `labelme/_app.py`, plus branch was `BEHIND`). Awaiting CI on the rebased branch before re-verdicting.

wkentaro:

Verdict: recommend-merge. CI is green on the rebased branch (all OS/Python cells, lint, cla, network-test), and the earlier recommend-merge verdict stands unconditionally now that the rebase is settled.

PR #2467, opened 2026-08-07
ci: run behavioral tests on every supported OS and Python

### LINKED ISSUE #2457: ci: run behavioral GUI tests on every supported OS and Python
> *This was generated by AI.*

## Problem

The test workflow currently runs `make test` only on Ubuntu. Windows and macOS matrix jobs build the project but do not run the functional or GUI test suite, leaving native Qt, filesystem, shortcut, menu, and focus behavior unchecked.

## Required behavior

- Run behavioral unit and end-to-end tests on Windows, macOS, and Linux for Python 3.12, 3.13, and 3.14.
- Keep rendering-sensitive pixel snapshots Linux-only if native output differs.
- Keep network tests in their existing isolated job.
- Split jobs only when needed to stay within practical CI time limits.

## Verification

All nine supported OS and Python cells must execute behavioral tests and report their results as required checks.

wkentaro:
Closes #2457.

The test workflow built the project on Windows and macOS but ran the suite only on Ubuntu, so platform regressions shipped untested. All nine OS × Python cells now run the behavioral suite with one plain `make test` invocation, and the matrix immediately earned its keep: it surfaced the dropped-file path-separator bug (split out as #2497, already merged) and the Windows settings-dialog sizing bug (split out as #2500, which this PR is stacked on).

Non-obvious bits:

1. Pixel snapshots are gated by a collection-time skip in `tests/e2e/conftest.py` keyed on the `pixel_snapshot` marker, not by workflow-level `-m` filters. The reference pixels are Linux-rendered, and putting the gate in code makes a plain `make test` correct on every OS — including a contributor's macOS or Windows machine, where the CI-only filter would have let the snapshots run and fail.

2. `tests/workflow_test.py` guards the one gap nothing else catches: the matrix's Python versions are derived from the `pyproject.toml` classifiers, so adding a 3.15 classifier fails the test until the workflow follows. Everything else it once asserted either fails loudly on its own or is visible in a diff, so it was trimmed to this single test.

3. The failed-write settings test now injects a `PermissionError` at the atomic-write boundary instead of chmod-ing the config directory read-only. The chmod trick skipped the test on Windows and covered nothing as root; the injection keeps the revert path tested on every cell.

4. `concurrency` with `cancel-in-progress` stops superseded runs, since each push now costs nine GUI cells. The `github.run_id` fallback keeps main pushes and `workflow_call` invocations (release.yml) in unique groups, so only obsolete PR runs are cancelled; safe.

## Test plan

- [x] `make lint` and `make test` (1246 passed; the 5 skips are the pixel snapshots, off-Linux by design)
- [x] Every commit passes lint and the full suite independently (verified in a detached worktree)
- [ ] The nine-cell run itself is confirmed by this PR's own CI

Stacked on #2500; merge that first (GitHub retargets this PR to main when it lands).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     strategy:
+      fail-fast: false
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
         python-version: ['3.12', '3.13', '3.14']
@@ -27,15 +28,18 @@ jobs:
         with:
           packages: xvfb libegl1 libxkbcommon0 libxcb-cursor0 libgl1
           version: 1.0
+      # Pixel snapshots skip themselves off Linux (tests/e2e/conftest.py), so
+      # one plain invocation runs the right set everywhere.
       - name: Test
         shell: bash
-        if: matrix.os == 'ubuntu-latest'
         env:
           MPLBACKEND: agg
         run: |
-          Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-          export DISPLAY=:99
-          make test
+          if [ "$RUNNER_OS" = Linux ]; then
+            xvfb-run -a make test
+          else
+            make test
+          fi
       - name: Run examples
         shell: bash
         if: matrix.os != 'windows-latest'
@@ -103,6 +107,4 @@ jobs:
         env:
           MPLBACKEND: agg
         run: |-
-          Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-          export DISPLAY=:99
-          make test PYTEST_ARGS="-m network --numprocesses=1"
+          xvfb-run -a make test PYTEST_ARGS="-m network --numprocesses=1"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,13 @@ on:
     branches: [main]
   pull_request:
   workflow_call:
+# Nine GUI cells per push are too expensive to keep running for superseded
+# commits. github.ref is unique per pull request (refs/pull/N/merge) even for
+# fork branches that share a name, and the event condition keeps main pushes
+# and workflow_call runs uncancelled, so only obsolete PR runs are cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 jobs:
   test:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     strategy:
+      fail-fast: false
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
         python-version: ['3.12', '3.13', '3.14']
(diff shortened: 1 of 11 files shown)


wkentaro:

CI fix attempt: all nine matrix cells failed at startup because the quoted `-m 'not network and not snapshot'` inside `PYTEST_ARGS` broke the Makefile's single-quoted echo macro (`SyntaxError` before any test ran). The marker filters now ride in the `PYTEST_ADDOPTS` environment variable, which pytest reads directly, so no Makefile quoting is involved on any OS. Reproduced the CI invocation locally: 1162 tests pass under the filter. Awaiting the fresh matrix run.

wkentaro:

Verdict: recommend-triage. The workflow itself now works: the quoting fix landed and all six macOS and Ubuntu cells run the behavioral suite green. The three Windows cells fail the same 11 pre-existing tests (inventory and cluster analysis in #2481); they are what this matrix was built to surface, not defects of this PR. The merge call is a scope decision:

1. Hold this PR until #2481 is burned down (matrix lands fully green, but Windows coverage stays off until then).
2. Merge now with the 11 tests marked xfail/skip on Windows, each linked to #2481, and burn them down behind green CI.
3. Merge with Windows cells non-required for now.

Option 2 keeps the new coverage protecting six cells immediately while making the Windows debt explicit; happy to implement it on request.

wkentaro:

Verdict: recommend-merge.

**Review.** Standards axis: clean — CHANGELOG entry correctly placed and linked, no hard violations. One minor judgement-call smell noted (the `HOME`/`USERPROFILE` steering pattern in `tests/e2e/conftest.py` duplicates a helper already extracted in `tests/unit/_config/api_test.py`), not blocking. Spec axis: the matrix genuinely covers all 9 OS×Python cells and pixel-snapshot/network isolation is correct (independently verified the `PYTEST_ADDOPTS` filters select the right tests). Two real app fixes (drag-drop path normalization, settings dialog width) are bundled in beyond the issue's literal CI-only scope, but both are legitimate byproducts of actually running the suite cross-platform for the first time and are explained in their own commits — not scope creep worth blocking on. Note: `main` has no branch-protection rule configured at all, so the spec's "report as required checks" language doesn't currently apply either way.

**Commits.** Already a clean, logical sequence (behavior fixes, then test prep, then the workflow change) — no reshaping needed.

**Verify.** Smoke-launched the app offscreen: `MainWindow` starts, a messy drag-and-drop path normalizes correctly into the file list, and the Settings dialog opens sized to its content (760px, matching `_preferred_width`). Screenshots looked correct, not blank.

**CI.** Green across all 9 matrix cells plus lint/cla/network-test, and this tick made no pushes, so the existing green stands.

wkentaro:

Verdict re-affirmed: recommend-merge.

The prior verdict comment predated the latest push (a history-only recommit — tree unchanged), which made the label stale per this tick's process. No code changes since the last review; all 19 CI checks re-ran green and merge state is CLEAN, so the verdict stands unchanged.

wkentaro:

Verdict: recommend-merge.

The prior verdict comment predated a real push (`bc0bb34e`, adding a `concurrency` block to cancel superseded PR runs) — re-reviewed that commit alone since the rest of the branch was already reviewed. Standards: clean, no repo-wide standards apply to CI YAML, no baseline smells. Spec: matches the PR description's stated goal (cancel obsolete PR runs, leave main pushes and `workflow_call` uncancelled); the mechanism (`github.ref` + `cancel-in-progress` gated on `github.event_name == 'pull_request'`) achieves that correctly. One non-blocking note: the PR body still describes an earlier `github.run_id`-fallback approach that predates this amend — the safety property it claims still holds, just via a different mechanism than described; worth a quick body edit but not a merge blocker.

Local recommit check: the 8-commit sequence is already clean — one purpose per commit, no rewrite needed. All 19 CI checks green, merge state CLEAN.

PR #2468, opened 2026-08-07
fix(app): preserve the current session when the next load fails

### LINKED ISSUE #2454: fix(app): preserve the current session when the next load fails
> *This was generated by AI.*

## Problem

`MainWindow._load_file` resets the active state before the replacement Image and Annotation have been read, decoded, and validated. A corrupt Annotation File or unreadable Image can therefore leave a blank disabled canvas and move File List state away from the last good session.

## Required behavior

- Stage the replacement Image and Annotation before changing the active session.
- Commit the new state only after reading, validation, and image decoding succeed.
- On failure, keep the current Image, Annotation, canvas, and File List selection active.
- A corrupt adjacent Annotation File blocks its Image from opening. Do not silently open an empty Annotation.

This implements `docs/adr/0006-annotation-transitions-preserve-last-good-state.md`.

## Verification

Add end-to-end cases for navigation from a valid session to a missing Image, unreadable Image, corrupt Annotation File, and semantically invalid Annotation File.

wkentaro:
Closes #2454.

`_load_file` reset the active session before the replacement had been read, decoded, or validated, so a corrupt or missing file left a blank disabled canvas and a dead File List selection. Loading is now a transaction: the Annotation, its shapes, and the decoded Image are staged first, the visible session is replaced only after everything is ready, and on failure `_load_file` returns `False` having touched nothing. Implements ADR 0006.

Non-obvious bits:

1. Qt moves the File List selection before the load starts, so the `currentItemChanged` handler keeps the previous item and restores it under blocked signals when the user cancels or the load fails; the rollback needs no path bookkeeping.

2. `_file_list_image_path` tracks which File List entry the session came from, because an Annotation File's embedded `imagePath` can rewrite `_image_path` away from the list entry. It feeds only selection restore; autosave and checkmark lookup still resolve from `_image_path` as on main — that divergence bug is pre-existing and filed as #2514.

3. Opening a directory commits its new File List even when the first Image fails to load: the previous session stays active while the remaining Images stay reachable (ADR 0006).

4. A rejected output-directory change restores the previous directory, so a corrupt candidate Annotation File never becomes the autosave target.

5. `read_image_file` wraps non-`OSError` decode failures as `OSError` because `tifffile` raises its own error type; without the wrap a corrupt TIFF bypasses the error dialog and the rollback never runs.

6. Consequence: failed next/prev navigation restores the selection, so keyboard navigation stops at a corrupt file; clicking a farther row still reaches the rest.

## Test plan

- [x] `uv run pytest tests/` (1292 passed, 6 skipped)
- [x] `make lint`


wkentaro:

Verdict: recommend-merge, with one ordering note: merge #2473 first. Independent review (gpt-5.6-sol) hardened this branch (direct file and directory loads no longer replace File List state before validation; unsupported Shape types take the load-error path with session restoration) and folded the fixes in. Type-specific Shape cardinality validation intentionally lives in #2473, so this PR fully satisfies ADR 0006 only once #2473 is in; merging in that order (or rebasing this branch afterward) closes the gap. All checks green.

wkentaro:

The merge-after-#2473 condition is now satisfied: #2473 merged and this branch is rebased onto it, dropping its duplicate Shape error handling in favor of the merged read-boundary validation. Full suite green after the rebase; the earlier recommend-merge verdict stands unconditionally.

wkentaro:

Verdict: **PASS — recommend merge.** Manually verified the current PR head (`99f42466`) with Computer Use in the running Labelme GUI.

- From a valid annotated session at `[1/2]`, Next Image into a corrupt JPEG showed the expected image-open error. After dismissal, the original image, all five annotations/shapes, enabled drawing controls, first File List selection, and `[1/2]` title remained intact.

- From the same valid session, Next Image into a valid JPEG with a corrupt adjacent JSON showed the expected label-open error. The candidate image was not opened without its Annotation; the original image, annotations/shapes, controls, File List selection, and title again remained intact.

No behavioral defects observed in the required rollback flow. The PR checks are also green.

Issue #2461, opened 2026-08-06
release: validate v7.1.0rc1 before final publication


wkentaro:
> *This was generated by AI.*

## Release gate

Status as of 2026-08-18. Gate issue: [#2461](https://github.com/wkentaro/labelme/issues/2461).

## Day 0 — merge before rc1

- [x] [PR #2496](https://github.com/wkentaro/labelme/pull/2496) cross-drive absolute imagePath (closes #1350) — **MERGED**
- [x] [PR #2477](https://github.com/wkentaro/labelme/pull/2477) atomic Annotation writes (closes #2451) — **MERGED**
- [x] [PR #2504](https://github.com/wkentaro/labelme/pull/2504) fractional Mask bounds (closes #2479) — **MERGED**
- [x] [PR #2474](https://github.com/wkentaro/labelme/pull/2474) filter without rescanning (closes #2455) — **MERGED**
- [x] [PR #2468](https://github.com/wkentaro/labelme/pull/2468) preserve session on failed load (closes #2454) — **MERGED**
- [x] [PR #2506](https://github.com/wkentaro/labelme/pull/2506) keep AI oriented rectangles inside Image bounds — **MERGED**
- [x] [PR #2512](https://github.com/wkentaro/labelme/pull/2512) keep bounded AI oriented rectangles valid — **MERGED**
- [x] [PR #2466](https://github.com/wkentaro/labelme/pull/2466) RGB to AI models (closes #2356) — **MERGED**
- [x] AI stack sequencing: #2506 / #2512 / #2466 merged serially with green CI
- [x] [PR #2508](https://github.com/wkentaro/labelme/pull/2508) migrate changelog workflow to towncrier fragments — **MERGED**
- [x] Confirm macOS 3.12 flake gone: prior run [32104527074](https://github.com/wkentaro/labelme/actions/runs/32104527074) red (xdist worker crash); current-main run [32136781258](https://github.com/wkentaro/labelme/actions/runs/32136781258) passed on `f8422a3f`
- [x] [PR #2502](https://github.com/wkentaro/labelme/pull/2502) copy dragged points into the moved shape (closes #2413) — **MERGED**
- [x] Defer [PR #2521](https://github.com/wkentaro/labelme/pull/2521) synchronize canvas mode status to the `next` milestone

## Cut rc1

- [x] `changelog.d/` has 60 pending files; `tools/release_notes.py` renders rc1 notes from them
[...]

PR #2469, opened 2026-08-07
ci: support prerelease tags and Unreleased notes

### LINKED ISSUE #2460: release: support prerelease tags and Unreleased notes
> *This was generated by AI.*

## Problem

The release workflow expects a changelog section whose version exactly matches the tag and always creates a normal GitHub release. The v7.1 plan calls for `v7.1.0rc1` to use the current Unreleased notes, publish to PyPI, and appear as a GitHub prerelease without permanently promoting an RC changelog section.

## Required behavior

- Detect PEP 440 prerelease tags such as `v7.1.0rc1`.
- For a prerelease, extract the current Unreleased changelog content.
- Publish the prerelease artifact to PyPI.
- Create the GitHub release with prerelease status.
- Keep the Unreleased section in place until final `v7.1.0` promotion.
- Preserve the current exact-version extraction for stable releases.

## Verification

Test note extraction for prerelease and stable tags without publishing, then exercise the real workflow with `v7.1.0rc1`.

wkentaro:
Closes #2460.

The release workflow required a changelog section exactly matching the tag and always created a normal GitHub release, so a `v7.1.0rc1` tag could not ship. A new tested extractor (`tools/release_notes.py`) detects PEP 440 prerelease tags: prereleases publish to PyPI with the current Unreleased notes and a GitHub prerelease, while stable tags keep the exact-version extraction unchanged. The Unreleased section stays in place until the final promotion.

## Test plan
- [x] Extractor regression tests for prerelease and stable tags (10 focused tests; suite at 1174 passed) plus `make lint`
- [x] Local extraction exercised with `v7.1.0rc1` and `v7.0.4`
- [ ] The real workflow runs when the actual `v7.1.0rc1` tag is pushed (release-gate step in #2461)


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) found only convention fixes (CHANGELOG links the PR), folded in during the rebase onto current main. Local extraction verified for prerelease and stable tags; all checks green. The real `v7.1.0rc1` workflow run stays a #2461 release-gate step.

wkentaro:
> *This was generated by AI during PR review.*

Final verdict: **merge**.

Reviewed the full diff against #2460's required behavior; all requirements are met and I found no blocking issues.

What I verified:

- **Extractor semantics** (`tools/release_notes.py`): PEP 440 prerelease detection via `packaging.Version.is_prerelease` correctly routes `a`/`b`/`rc`/`-rc`/`.dev` tags to the `Unreleased` section while stable tags keep the exact-version lookup, matching the old awk behavior including the fail-on-missing-section guard. It also adds a new guard the awk version lacked: an *empty* Unreleased section fails the release instead of shipping blank notes. Invalid tags exit non-zero via `parser.error`, so a bad tag still fails the workflow.
- **Workflow wiring** (`.github/workflows/release.yml`): I parsed the YAML — the folded `run` scalar resolves to `uv run tools/release_notes.py "$GITHUB_REF_NAME" CHANGELOG.md release-notes.md >> "$GITHUB_OUTPUT"`, and the script's only stdout is the `prerelease=true|false` line (uv sync chatter goes to stderr), so the step output is clean. The `--prerelease` flag is conditionally added from that output, which can only be `true`/`false` from our own script, so the `${{ }}` interpolation is injection-safe.
- **Dependency availability**: `packaging` was added to `[dependency-groups] dev`, which `uv run` includes by default, so the release job resolves the import without extra flags.
- **Tests**: 10 focused tests cover prerelease/stable routing, exact-version matching, missing-section rejection, end-to-end `main()` behavior (CHANGELOG left untouched, correct `prerelease=` stdout), plus a build→publish→release ordering check on the workflow file. CI is green across all 9 matrix jobs, lint, and network-test.
- **Changelog**: entry filed under `### Changed` with the PR linked, per convention.

Residual risk, already acknowledged in the test plan: the workflow itself only gets a true end-to-end run when the real `v7.1.0rc1` tag is pushed (release-gate step in #2461). Nothing in the diff suggests it will misbehave there.

Issue #2352, opened 2026-07-15
fix(app): auto-save branch of mark_dirty never clears the dirty state


wkentaro:
> *This was generated by AI during a kaizen pass.*

## Summary

When "Save Automatically" is toggled **on** after the annotation is already dirty, `mark_dirty()`'s auto-save branch writes the file to disk but never clears the dirty state. From then on the app keeps behaving as if there are unsaved changes even though the file is fully in sync.

## Where

`labelme/_app.py:1292-1307` — `mark_dirty()`:

```python
def mark_dirty(self) -> None:
    self._actions.undo.setEnabled(self._canvas_widgets.canvas.can_restore_shape)

    if self._actions.save_auto.isChecked():
        assert self._image_path is not None
        self.save_labels(
            label_path=_resolve_label_path(
                image_or_label_path=self._image_path,
                output_dir=self._output_dir,
            )
        )
        return                      # <-- saves, but never clears the dirty state
    self._is_changed = True
    self._actions.save.setEnabled(True)
    self.setWindowTitle(self._get_window_title(dirty=True))
```

The manual-save path already does the right thing — `_save_label_file()` at `labelme/_app.py:2380-2381`:

```python
if self.save_labels(label_path=label_path):
    self.mark_clean()
```

The auto-save branch is missing that `mark_clean()` after a successful `save_labels(...)`.

## Reproduce

1. Start with `auto_save=False`.
2. Edit a shape → `_is_changed` becomes `True`, Save action enabled, title shows the "modified" marker.
3. Toggle **Save Automatically** on (`_actions.save_auto`).
4. Edit again → this now takes the auto-save branch: the file is rewritten every time, but `_is_changed` stays `True` forever.

Result: the title stays "modified", the Save action stays permanently enabled, and `_can_continue()` (`_app.py:2602`, which only checks `if not self._is_changed`) keeps popping the *"Save annotations … before closing?"* dialog on every close — even though the file on disk is already fully synced. State-consistency / UX bug, not data loss.

## Why this needs a decision (not a blind fix)

The obvious one-line fix — mirror line 2380 and call `mark_clean()` in the auto-save branch — has a **subtle regression**, which is why I'm escalating instead of shipping:

- `mark_clean()` (`_app.py:1309-1319`) re-enables **all** draw actions (`for _, action in self._actions.draw: action.setEnabled(True)`, lines 1312-1313) and toggles `delete_file`.
[...]

PR #2472, opened 2026-08-07
fix(auto-save): keep edits dirty after a save failure

### LINKED ISSUE #2452: fix(auto-save): keep edits dirty after a save failure
> *This was generated by AI.*

## Problem

`MainWindow.mark_dirty` calls `save_labels` when auto-save is enabled but ignores its boolean result. If the write fails while the window was clean, the edit remains only in memory while the application still treats the Annotation as clean. Closing can then skip the unsaved-changes prompt.

## Required behavior

- A failed auto-save leaves the Annotation dirty.
- The window title and Save action show the dirty state.
- Closing still asks about unsaved changes.
- The first failure shows the existing save error.
- Repeated edits do not show repeated dialogs until a save succeeds or the target path changes.
- The auto-save Setting remains enabled.
- Manual Save remains available as a retry.

Successful auto-save state is tracked separately in #2352.

## Verification

Add an end-to-end test that edits a clean Annotation, forces auto-save failure, checks the dirty UI, checks the close prompt, and then retries successfully.

wkentaro:
Closes #2452.

A failed auto-save silently marked the annotation clean, so edits could be lost with no warning. A failed save now keeps the annotation dirty: Save stays enabled, closing prompts about unsaved changes, and repeated auto-save failures show one error per target path until a save succeeds or the path changes. This implements ADR 0006 (annotation transitions preserve the last good state); the success-path dirty-state fix is #2352 on its own branch.

## Test plan
- [x] End-to-end tests: save failure keeps dirty, retry after the failure clears it, repeated-failure dialog suppression, suppression reset on path change
- [x] `make lint` and `make test` (1166 passed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed failed auto-saves leaving an edited Annotation marked clean. The dirty title, Save action, and unsaved-changes prompt now remain active, while repeated failures for the same target show only one error until a save succeeds or the target changes ([#2472](https://github.com/wkentaro/labelme/pull/2472))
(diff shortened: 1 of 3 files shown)


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) caught a state-machine defect, fixed and folded in: auto-save failure suppression survived an Annotation reset, which could silence the first error on the next file. All checks green. Note: shares one `save_labels()` hunk with #2475; whichever merges second needs a trivial rebase.

PR #2473, opened 2026-08-07
fix(label-file): reject malformed Shape semantics during read

### LINKED ISSUE #2453: fix(label-file): reject malformed Shape semantics during read
> *This was generated by AI.*

## Problem

The Annotation File reader accepts some structurally valid Shape dictionaries that are not valid Shapes. For example, an unknown `shape_type` can pass the codec and later raise during GUI Shape construction. This turns malformed input into an application failure instead of one clear load error.

## Required behavior

- Validate supported Shape types and their semantic invariants at the Annotation File boundary.
- Cover point-count, Mask, and related Shape invariants that are required before GUI construction.
- Wrap failures as `LabelFileReadError` with the file and failing field identified.
- Do not partially install Shapes from a malformed Annotation File.

## Verification

Add codec and end-to-end cases for malformed but valid JSON, including an unknown Shape type and invalid geometry for representative Shape types.

wkentaro:
Closes #2453.

The label-file reader accepted structurally valid JSON whose shapes were semantically broken (unknown shape types, wrong point counts, non-finite coordinates, missing or non-2D mask data), installing partial state that could later corrupt a save. Reading now validates shape semantics at the boundary and raises `LabelFileReadError` naming the file and the failing shape index and field (e.g. `shapes[1]: points ...`); no shapes are partially installed. Together with #2454 this keeps a corrupt annotation file from silently opening as an empty annotation (ADR 0006).

Validation covers only the invariants the GUI needs before Shape construction. Degenerate geometry (zero-extent rectangles, coincident points, 2-point polygons written by v5.x ai_polygon) and mask/bbox extent drift from fractional whole-shape drags can all be produced by released versions, so such files stay loadable.

## Test plan

- [x] Unit tests for each rejected malformed-semantics class with indexed error messages, plus accept-cases locking in that degenerate-but-saveable geometry stays loadable
- [x] End-to-end tests confirming a rejected file shows one error naming the file and field and leaves the session without partially installed shapes
- [x] `make lint` and `make test` (1200 passed)


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) caught oversized integer coordinates escaping as a raw `OverflowError`; now a field-specific `LabelFileReadError` with a regression test, folded in. All checks green. Follow-ups filed, non-blocking: #2479 (fractional Mask bounds convention), #2480 (self-crossing oriented-rectangle validation strictness).

Issue #2462, opened 2026-08-06
fix(file-list): align Previous and Next with visible Images


wkentaro:
> *This was generated by AI.*

## Problem

Previous and Next are enabled without regard to the current File List position. At a boundary their handlers only log and return, so enabled actions can do nothing. In-memory filtering also needs one explicit navigation contract.

## Required behavior

- Previous and Next navigate only among Images visible under the active File List filter.
- If the active Image is hidden by the filter, Previous selects the nearest visible Image before it in the original directory order, and Next selects the nearest visible Image after it.
- Disable an action when no visible Image exists in that direction.
- Keep both actions disabled for an empty visible result set.
- Recompute action state after import, filtering, successful navigation, and clearing the filter.
- Apply the same behavior to menu, toolbar, and keyboard shortcut activation.

This builds on #2455, which changes search to an in-memory filter without navigating or replacing the active session.

## Verification

Add end-to-end cases for the first and last visible Images, one visible Image, no visible Images, a hidden active Image with matches on both sides, and clearing the filter.

wkentaro:
> *This was generated by AI during triage.*

Blocked by #2455, which must define the in-memory filter this navigation contract builds on. The `ready-for-agent` label is removed on purpose until #2455 merges. Restore the label after that merge.

wkentaro:
> *This was generated by AI during triage.*

Still blocked. #2455 is open and its PR #2474 has not merged, so the in-memory filter this navigation contract builds on does not exist yet.

Moving to `needs-triage` rather than leaving it unlabeled: with no state label the issue re-enters the untriaged queue on every triage pass and gets re-examined for the same answer. Parking it here stops that churn without implying any other verdict.

Maintainer action when #2474 merges: swap this label back to `ready-for-agent`. The body is already a complete brief and needs no rework.

PR #1384, opened 2023-12-31
Regex filename search

wkentaro:
Close https://github.com/labelmeai/labelme/pull/1359

Thanks @michaelitvin

Screenshot 2023-12-31 at 13 46 13

PR #2474, opened 2026-08-07
fix(file-list): filter loaded Images without rescanning or navigation

### LINKED ISSUE #2455: fix(file-list): filter loaded Images without rescanning or navigation

## Problem

Each edit in the File List search control calls the directory-import path again. Labelme rescans the directory, rebuilds the File List, and calls the continue check. As a result, typing in search can show a save or discard prompt for a dirty Annotation and can disturb the active Image and Annotation.

Regex search is intentional existing behavior. It was added by #1384 after the request in #884. This issue must preserve it.

## Agent Brief

**Category:** bug

**Summary:** Filter the loaded Images in memory without rescanning, navigating, or changing the active Annotation.

**Current behavior:**

Changing the File List search text runs the same path used to import a directory. This path can scan the filesystem, call the continue check, rebuild navigation state, and change the active session.

The search query is a case-sensitive regular expression applied to the stored Image path. This is released behavior and is not the bug.

**Desired behavior:**

A directory import must load the directory Image set once. Changes to the File List search text must derive the visible File List from that loaded set in memory.

Filtering must not scan the filesystem, ask whether to save a dirty Annotation, load another Image, or change the active Image or Annotation. If the filter hides the active Image, that Image and Annotation must remain active. Clearing the query must restore the full File List and reselect the active Image when it belongs to the loaded directory.

Keep the existing case-sensitive regex matching behavior. Do not replace it with literal or case-insensitive matching.

**Key interfaces:**

- The directory import path, which owns filesystem scanning and the complete ordered Image set
- The File List search-change path, which must only derive a visible subset in memory
- The current Image and Annotation session, which must not change during filtering

**Acceptance criteria:**

- [ ] A directory is scanned once when it is imported.
- [ ] Editing the search query does not rescan the directory or call the continue check.
- [ ] Existing case-sensitive regex matching against Image paths is preserved.
- [ ] Filtering a dirty Annotation does not show a save or discard prompt.
- [ ] If filtering hides the active Image, its Image and Annotation remain active.
- [ ] Clearing the query restores the full File List and its active-Image selection.
- [ ] Initial import still applies the configured `file_search` regex.
- [ ] End-to-end tests cover a dirty Annotation, valid regex matching, a hidden active Image, and clearing the query.

**Out of scope:**

- Changing regex search to literal or case-insensitive search
- Defining new UI behavior for an invalid regex
- Previous and Next navigation within a filtered File List, which is tracked by #2462

wkentaro:
Closes #2455.

Typing in the File List search re-ran the directory-import path, so each keystroke rescanned the filesystem, could prompt to save a dirty annotation, and could replace the active image. A directory is now scanned once into a loaded image set, and a single refresh derives the visible File List from that set plus the search box: no rescan, no continue check, no session change, and the active image stays active even while the filter hides it. Case-sensitive regex matching (released behavior from #1384) is preserved. #2462 will build its Previous/Next navigation contract on this; navigation logic is untouched here.

Non-obvious bits:

1. The refresh reads the search box directly instead of taking a `pattern` argument. The change-annotations-dir path used to rebuild with `pattern=None` while the box still held text, silently dropping the filter; making the widget the single source of truth deletes that hole along with the parameter. Safe at startup: the configured `file_search` value is written into the widget before any directory load.

2. Dropped images join the loaded set and go through the same refresh, so a drop now respects an active filter instead of appending rows that contradict the search box. Hidden drops stay loaded; clearing the search reveals them.

3. The rebuild runs under `QSignalBlocker`, which is what keeps the selection-changed load path (the only signal connected to the list widget) from firing; the old unblocked `clear()` could even prompt twice on one keystroke.

## Test plan

- [x] e2e: single scan per import; filtering while dirty triggers no prompt, no continue check, no load; hidden active image stays active; clearing restores the full list and its selection; drops respect the active filter; opening a label file resets the loaded set
- [x] `make lint`, `make test` (1251 passed)


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) caught three real defects, all fixed and folded in: search edits still ran Annotation File existence checks; separate output directories broke active-Image reselection on unnormalized paths; dropped-Image filtering could leave no active path. All checks green. #2462 unblocks when this merges.

wkentaro:

Rebased onto `main` (post-base changes to `labelme/_app.py` and `CHANGELOG.md` overlapped this PR's files). Two-axis code review: Standards found only judgement-call nits (some duplicated has-label lookup/cache-update logic, mild primitive obsession from keeping `_image_paths` and `_has_label_by_image_path` in sync by hand) — no hard violations, so `/review-fix` was skipped. Spec review confirmed all #2455 acceptance criteria are met with no scope creep and no defects.

Recommitted into two commits (implementation, then tests) without changing the final tree; validated each in isolation with lint + a targeted offscreen pytest run. Force-pushed, which re-triggers CI — will verdict once it's green.

wkentaro:

Verdict: recommend-merge. CI is green on the rebased, recommitted head (`d3d8e8eb`) across all platforms/Python versions, lint, and network-test. No further changes needed.

PR #2475, opened 2026-08-07
fix(app): clear dirty state after successful auto-save

### LINKED ISSUE #2352: fix(app): auto-save branch of mark_dirty never clears the dirty state

wkentaro:
Closes #2352.

The auto-save branch of `mark_dirty` saved to disk but never reset the dirty flag, so the title kept its "modified" marker and closing prompted about unsaved changes even though the file was in sync. Per the pinned triage decision, a successful auto-save now clears only the dirty flag, Save action, and title marker, without re-enabling the create-mode draw actions.

Rebased on #2482, which is merged. `mark_clean()` is now the exact opposite of `mark_dirty()`, so the private `_clear_dirty_indicators` helper is gone and the whole fix is one line:

```python
if self.save_labels(...):
    self.mark_clean()
    return
```

The regression test's last assertion (`polygon` draw action still disabled) is what pins the "without re-enabling create-mode draw actions" half, now that `mark_clean()` no longer touches those actions.

## Test plan

- [x] End-to-end regression: auto-save persists the file, clears the dirty indicators, keeps create mode intact, and closing shows no prompt
- [x] `make lint` and `make test` (1230 passed) after the rebase


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) deduplicated the dirty-indicator reset into a shared private method and fixed the CHANGELOG link; folded in. All checks green. Note: shares one `save_labels()` hunk with #2472; whichever merges second needs a trivial rebase.

PR #2476, opened 2026-08-07
fix(canvas): one coherent drag-bounds pass

### LINKED ISSUE #2365: Oriented-rectangle corner drag can leave a corner out of bounds even with out-of-bounds points disabled

## Summary

While reviewing #2364 (which fixes the single-vertex slide in `_compute_intersection_edges_image`), fuzzing surfaced a **pre-existing, best-effort limitation** in the *oriented-rectangle* consumer of that helper: dragging a corner far off-canvas can leave a corner **outside the image even when out-of-bounds points are disabled** (the default).

This is not introduced by #2364 and does not block it; both the pre-#2364 and post-#2364 code exhibit it. Filing so it isn't lost.

## Where

`_reproject_oriented_rectangle_corners` in `labelme/_widgets/canvas.py` clips a dragged oriented-rectangle corner by intersecting the two adjacent parallelogram edges against the image via `_compute_intersection_edges_image`, then projecting `moving` onto the segment `edge_a`–`edge_b` (`_utils.project_point_on_line`). When both adjacent start points clip to the **same** image-boundary coordinate, `edge_a == edge_b`; `project_point_on_line` hits its `length_sq == 0` degenerate branch and returns `moving` **unchanged** — i.e. fully unclipped. The result is used raw (`dataclasses.replace(current, points=new_corners)` / `shape.move_vertex`), so the corner is drawn outside the image.

## Evidence

Fuzzing 200k oriented-rectangle drags that *start* fully in-bounds, dragging one corner up to 60px past the edges (image 100×50), `allow_out_of_bounds=False`:

- corner left out-of-bounds: ~28% of drags (this is inherent to the best-effort parallelogram clipping, not specific to any one revision)
- degenerate `edge_a == edge_b` collapses occur in a large fraction of off-canvas drags

Example (post-#2364): `corners=[(41.5,17.4),(44.9,22.6),(27.6,34.1),(24.2,28.9)]`, drag corner 3 to `(21.7, 99.9)` → corner lands at `(21.9, 77.4)`, ~27px below the image bottom.

## Why it's hard (not a trivial clamp)

Keeping a dragged parallelogram both a valid parallelogram **and** fully inside the image is not always geometrically possible, so some best-effort compromise is unavoidable. The open question is whether the degenerate `edge_a == edge_b` case should fall back to a plain per-axis clamp of `moving` (as the single-vertex path effectively does) rather than returning it unclipped.

Originating PR: #2364

### LINKED ISSUE #2375: Whole-shape drag inverts direction for a shape larger than the image (out-of-bounds points off)

## Summary

`Canvas._drag_shapes` (`labelme/_widgets/canvas.py:1354-1376`) inverts the drag
direction when the dragged selection's bounding box is larger than the image and
`canvas.allow_out_of_bounds_points` is off. Dragging a too-wide shape to the
right teleports it to the left (and vice versa for a too-tall shape dragged
vertically).

## Reproduction

1. Enable the beta setting `canvas.allow_out_of_bounds_points`.
2. Draw a shape wider (or taller) than the whole image, e.g. a rectangle whose
   left edge is off the left side and right edge off the right side.
3. Disable `canvas.allow_out_of_bounds_points` again.
4. Left-drag the shape to move it (this also reproduces via arrow-key nudge,
   which routes through `_move_by_keyboard` -> `_drag_shapes`).

Observed: dragging right moves the shape left. A direct call proves it: with a
100x50 pixmap, a rectangle of width 160 (`points=[[-30,0],[130,10]]`), and a
rightward cursor delta of +5, the shape moves left by 60px.

## Root cause

```python
if not self._allow_out_of_bounds_points:
    pw = float(self.pixmap.width())
    ph = float(self.pixmap.height())
    target.setX(max(0.0, target.x()))
    target.setY(max(0.0, target.y()))
    target.setX(min(target.x(), pw - bounds.width()))   # pw - bounds.width() < 0 when too wide
    target.setY(min(target.y(), ph - bounds.height()))
```

The clamp keeps the top-left `target` inside `[0, pw - bounds.width()]` so the
whole bounding box stays within `[0, pw]`. When `bounds.width() > pw`, the upper
bound `pw - bounds.width()` is negative, so the trailing `min(...)` overrides the
preceding `max(0.0, ...)` and forces `target.x()` negative regardless of drag
direction. The `[0, pw - bounds.width()]` interval is empty (unsatisfiable): a
shape wider than the image can never have its whole bbox inside the image.

Such an over-sized, out-of-bounds shape is only reachable by drawing it with
`allow_out_of_bounds_points` on and then turning the setting off (per
ADR 0004, loading never clamps, and toggling the setting off does not clamp
existing points).

## Why this needs a decision

The current behavior is unambiguously wrong (drag inversion), but the correct
behavior when a shape cannot fit in the image is a product call with more than
one reasonable answer, and it interacts with ADR 0004 (out-of-bounds points):

- **Free-drag on the overflowing axis** - skip the upper clamp only on the axis
  where the shape does not fit; the user can move it anywhere on that axis.
- **Symmetric clamp (keep the image covered)** - clamp `target` into
  `[min(0, pw - w), max(0, pw - w)]`, so a too-wide shape always fully covers the
  image horizontally but cannot slide entirely off.
- **Block the drag entirely** when the shape cannot fit (like the existing
  off-image cursor guard).

## Affected files

- `labelme/_widgets/canvas.py` (`_drag_shapes`, lines 1354-1376)
- Tests: `tests/unit/widgets/canvas_test.py`
  (`test_drag_shapes_blocked_off_image_by_default` /
  `test_drag_shapes_keeps_out_of_bounds_when_enabled` only use bounds smaller
  than the pixmap, so this case is uncovered)

Once the desired semantics are chosen, the fix is a small, local change to the
clamp plus a regression test.

### LINKED ISSUE #2398: Keyboard-nudging a shape selected without a canvas click misbehaves (escapes bounds one way, silent no-op the other)

## Summary

When a shape is selected **without a canvas mouse click** — i.e. via Select-All (`Ctrl+A`), a label-list-sidebar click, or a programmatic/undo-redo selection — arrow-key nudging (`_move_by_keyboard` → `_drag_shapes`) misbehaves, because two pieces of drag state are only ever established by a real canvas click (`_select_shape_point` → `_record_drag_anchor`). With out-of-bounds points disabled (the default), the same user scenario produces two distinct, inconsistent failures depending on direction:

1. **Stale clamp anchor → shape escapes the image.** `_drag_shapes` clamps the target using `self._drag_anchor`'s recorded bounding box. On a clickless selection that anchor is still the zero-size default `(QPointF(), QRectF())` from `__init__` (`canvas.py:249`), so the edge clamp measures against a `0×0` box and the shape slides past the image edge. (Verified: a rectangle at x∈[85,95] on a 100px-wide image, selected without a click, nudged Right 3× ends at x=110 — 10px past the edge.)

2. **Synthetic-cursor pixmap gate → Left/Up nudges silently do nothing.** `_drag_shapes` first runs `if self._should_constrain_to_pixmap(cursor): return False`, where `cursor = self._prev_point + offset`. `_prev_point` is a mouse-tracking value, unrelated to the shape's position, and on a clickless selection can be at its `(0,0)` default. A Left (`offset=(-5,0)`) or Up (`offset=(0,-5)`) nudge then yields a negative `cursor`, which reads as out-of-pixmap, so the drag returns before moving anything — a silent no-op, even though moving the shape is perfectly in-bounds. (Verified via repro below.)

So in the identical scenario, nudging toward the far edge lets the shape escape (bug 1) while nudging toward the origin does nothing (bug 2). Related pre-existing variant (F4): `_prev_point` is never reset on image switch (`load_pixmap`), so a stale value from a larger previous image can wrongly gate a nudge on a smaller one.

## Reproduction

```python
import numpy as np
from PySide6 import QtGui, QtWidgets
from PySide6.QtCore import QPointF
from labelme._shape import Shape
from labelme._widgets.canvas import Canvas

app = QtWidgets.QApplication([])
c = Canvas(); c.pixmap = QtGui.QPixmap(100, 50)
s = Shape(shape_type="rectangle",
          points=np.array([(40, 20), (60, 30)], dtype=np.float64), closed=True)
c.shapes = [s]; c.selected_shapes = [s]   # clickless selection; _drag_anchor never recorded

c._move_by_keyboard(offset=QPointF(-5.0, 0.0))  # Left  -> shape does NOT move (bug 2)
c._move_by_keyboard(offset=QPointF(0.0, -5.0))  # Up    -> shape does NOT move (bug 2)
print(s.points.tolist())  # unchanged: [[40, 20], [60, 30]]
```

## Why this needs a human decision

The surgical one-liner I prototyped — call `self._record_drag_anchor(click=self._prev_point)` at the top of `_move_by_keyboard` — fixes bug 1 (and passes a regression test) but leaves bug 2 untouched, delivering an inconsistent half-fix within a single user scenario. The clean fix is a **shared-choke-point decision** in the drag machinery, with more than one reasonable answer:

- **(a)** Recompute the clamp anchor on demand inside `_drag_shapes` from the current `selected_shapes` bounds whenever it is unset/mismatched, so every present and future caller is correct by construction; and gate the pixmap check on the shape's own target bounds rather than the synthetic `_prev_point + offset` cursor.
- **(b)** Give `Canvas.selected_shapes` a real setter (it is currently a plain public attribute assigned directly from `_app.py`) that refreshes the anchor on every selection change.
- **(c)** Keep the per-gesture `_record_drag_anchor` convention but also seed `_prev_point` from the selection for the keyboard path, so both the anchor and the gate agree on "current position."

This is shared machinery also under active, coordinated work (#2375 whole-shape drag inversion, #2364 / #2365 corner-drag out-of-bounds), so the approach should be chosen coherently with those rather than piecemeal-patched here.

## Affected code

- `labelme/_widgets/canvas.py`: `_move_by_keyboard` (~1682), `_drag_shapes` (~1354, the `_should_constrain_to_pixmap(cursor)` gate at ~1355 and the anchor clamp at ~1358-1366), `_record_drag_anchor` (~1297), `_drag_anchor` default (249), `load_pixmap` (~1808, `_prev_point` not reset).
- `labelme/_app.py`: `_on_shape_selection_changed` (~1670) assigns `canvas.selected_shapes` directly without refreshing the anchor.

A prototype partial-fix patch (bug 1 only) plus its regression test is saved locally under `log/agents/2026-07-23_keyboard-nudge-clickless-selection/partial-anchor-fix.patch` as a starting point.

wkentaro:
Closes #2398. Closes #2375. Closes #2365.

One pass over the `_drag_shapes` bounds machinery, per the triage decision to fix these together instead of patching the same code three ways:

- **#2398**: shapes selected without a canvas click (Ctrl+A, label-list sidebar, undo/redo) left a stale drag anchor, so keyboard nudges either escaped the image bounds or silently did nothing. The anchor is now rebuilt from the current shape bounds on demand, and nudges are gated on shape target bounds instead of the synthetic cursor position.
- **#2375**: for a shape larger than the image the clamp interval was empty and the trailing `min()` forced the shape one direction regardless of drag direction. Oversized shapes now use a symmetric clamp, so a drag never moves the shape opposite to the drag direction.
- **#2365**: an oriented-rectangle corner drag could leave a corner out of bounds with out-of-bounds points disabled; degenerate projections are now clipped safely.

## Test plan
- [x] Regression tests: clickless-selection nudges stay in bounds and move in the pressed direction, oversized-shape drags never invert, oriented-rectangle corner drags keep every corner in bounds
- [x] `make lint` and `make test` (1178 passed)


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) caught a clamp edge case, fixed and folded in: symmetric clamping could reverse movement for a Shape already out of range. All checks green. Merging closes #2398, #2375, and #2365 together.

PR #2477, opened 2026-08-07
fix(label-file): write Annotation Files atomically

### LINKED ISSUE #2451: fix(label-file): write Annotation Files atomically
> *This was generated by AI.*

## Problem

`write_label_file` opens the target Annotation File with `"w"` and streams `json.dump` into it. A serialization or filesystem failure can therefore truncate the last good Annotation File before the replacement is complete.

## Required behavior

- Write the complete replacement to a temporary file in the target directory.
- Replace the target with `os.replace` only after the temporary file closes successfully.
- Leave the previous Annotation File intact after serialization, write, close, or replacement failure.
- Remove abandoned temporary files.
- Do not force `fsync` on every auto-save.
- Do not create persistent backup files.

This implements `docs/adr/0006-annotation-transitions-preserve-last-good-state.md`.

## Verification

Add tests for successful replacement, serialization failure, write failure, replacement failure, and temporary-file cleanup.

wkentaro:
Closes #2451.

Saving streamed `json.dump` straight into the target file, so a crash or full disk mid-write truncated the last good Annotation File. Saves now write a sibling `.tmp` file and `os.replace` it over the target once the temporary file has closed cleanly, per ADR 0006 (no forced fsync per auto-save). Any failure leaves the previous file intact and removes the temporary file.

Non-obvious bit: `os.replace` installs the temporary file's permissions, so the previous file's mode is copied onto the temporary file before the rename. POSIX-only: Windows has no POSIX modes, and a read-only Windows target cannot be replaced (or saved over before this change) anyway.

## Test plan

- [x] Unit tests: failure injected at serialization, write, close, and replace each preserves the prior file and leaves no `.tmp`; existing mode preserved, new files get the umask default
- [x] `make lint` and `make test` (1227 passed)


wkentaro:

Verdict: recommend-merge. Independent review (gpt-5.6-sol) found only the CHANGELOG link convention to fix, folded in during the rebase. Failure-stage tests all pass; all checks green.

wkentaro:

## Review-fix summary

Ran a full review-fix loop (3 rounds) on top of the existing atomic-write implementation. The prior version had a real permission-exposure bug: the temp file was `chmod`'d to match the target's mode *before* the annotation content was written, so for any annotation file with a non-default (looser) mode, the temp file sat at that looser mode for the entire write — the opposite of what `tempfile`'s private-by-default temp files are for.

Changes force-pushed:

- **`fix(label-file): write annotation files atomically`** — reworked so the temp file starts at `mkstemp`'s private `0o600` and stays there for the whole write, only widening to the target's real mode once the write is complete and closed (`chmod` moved after, not before, via `os.fchmod` on the still-open fd to also avoid a symlink-swap race). A filesystem that rejects `chmod` now degrades gracefully (keeps the private default) instead of losing a completed write.
- **`refactor(config): reuse the shared atomic-write helper`** — `_config/_writer.py` had its own independent atomic-write implementation that had already started drifting from `_label_file.py`'s; both now share `labelme/_fs.py::atomic_write`. Config File save behavior is unchanged (`preserve_mode=False`, as before).
- **`docs: note atomic annotation writes`** — CHANGELOG entry, worded to say "mode bits" rather than "permissions" (owner/group, ACLs, and symlink/hardlink identity are not preserved by `os.replace`; that's an inherent, pre-existing tradeoff of atomic replacement, not new here).

All verified: 1062 unit tests pass, `ruff`/`ty` clean, plus a real (non-mocked) smoke test of the write path confirming new-file mode, mode preservation across a re-save, and no leaked temp files.

**Known, deliberately out-of-scope tradeoffs of the atomic-replace pattern** (inherent to `os.replace`, not introduced by this PR, consistent with how `_config/_writer.py` already behaved): a symlinked or hardlinked annotation path loses that link identity on save; a read-only target's write-protection is no longer enforced by the OS (the app doesn't currently check this either way); a save into a non-writable directory now fails where an in-place `open(path, "w")` used to succeed. Filing follow-up issues for these separately so they don't block this PR.

This PR is now awaiting CI on the force-pushed branch.

wkentaro:

follow-ups filed from this review: #2485, #2486, #2487, #2488 (none block this PR's merge).

wkentaro:

Verdict: recommend-merge. Post-review-fix commits are clean (3 logical commits: fix, refactor, docs), CHANGELOG entry present, all checks green, merge state CLEAN. Follow-ups filed from this review: #2485, #2486, #2487, #2488 (none block this PR's merge).

PR #2484, opened 2026-08-17
fix: keep viewport position with previous zoom

wfzark:
## Summary

- carry the previous horizontal and vertical scroll positions to newly opened images when **Keep Previous Zoom** is enabled
- continue restoring an image's own saved viewport when revisiting it
- add a GUI regression test covering zoom and viewport preservation across image navigation

## Root cause

Labelme retained the zoom widget value across images, but scroll positions were keyed only by image path. A newly opened image therefore had no stored scroll values and reset its viewport even though the zoom level was preserved.

## User impact

Annotators working through aligned image sequences can keep the same magnified region centered while moving between frames, avoiding repetitive panning on every image.

## Validation

- `QT_QPA_PLATFORM=offscreen uv run pytest tests/e2e/zoom_test.py -q` (8 passed)
- `uv run ruff format --check labelme/_app.py tests/e2e/zoom_test.py`
- `uv run ruff check labelme/_app.py tests/e2e/zoom_test.py`
- `uv run ty check labelme/_app.py tests/e2e/zoom_test.py`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed Keep Previous Zoom preserving the scale while resetting the viewport position on each newly opened image. It now carries the previous horizontal and vertical scroll positions forward, while still restoring an image's own saved viewport when revisiting it ([#2484](https://github.com/wkentaro/labelme/pull/2484))
 - Fixed failed auto-saves leaving an edited Annotation marked clean. The dirty title, Save action, and unsaved-changes prompt now remain active, while repeated failures for the same target show only one error until a save succeeds or the target changes ([#2472](https://github.com/wkentaro/labelme/pull/2472))
 - Fixed a cancelled close still overwriting the Window State (window size, position, and dock layout); `closeEvent` ignored the close event but kept going, so cancelling the unsaved-changes prompt saved the current geometry anyway and the next session started from a layout the user never closed on. The Window State is now written only when the close goes through ([#2478](https://github.com/wkentaro/labelme/pull/2478))
 - Fixed Mask Shape construction for Model Session detections with fractional bounding boxes by using the same rounding for Shape points and Mask extents, preventing placement drift and Existing Shape Suppression errors ([#2464](https://github.com/wkentaro/labelme/pull/2464))
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -1943,6 +1943,13 @@ class MainWindow(QtWidgets.QMainWindow):
         if self._image_path is not None:
             self._scroll_values[orientation][self._image_path] = value
 
+    def _remember_current_viewport(self) -> None:
+        if self._image_path is None:
+            return
+        for orientation, bar in self._canvas_widgets.scroll_bars.items():
+            self._scroll_values[orientation][self._image_path] = bar.value()
+        self._prev_image_path = self._image_path
+
     def _set_zoom(self, value: float, pos: QtCore.QPointF | None = None) -> None:
         if self._image_path is None:
             logger.warning("image_path is None, cannot set zoom")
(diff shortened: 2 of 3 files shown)


wfzark:

wkentaro:

## Review

**Standards.** No repo-wide `CODING_STANDARDS.md`/`CONTRIBUTING.md`. Against the Fowler smell baseline: none apply. The new `elif` branch in `_paint_canvas`'s scroll-restore loop directly mirrors the existing `keep_prev_brightness_contrast` pattern a few lines up (`self._prev_image_path`-keyed fallback), so it's consistent with how this codebase already handles "carry forward from the previous image" for other `keep_prev_*` settings.

**Spec / correctness.** The change is scoped correctly: only applies when `keep_prev_scale` is on, only as a fallback when the *current* image has no stored scroll value yet (an image's own saved viewport still takes priority via the `if` branch above), and only when a previous image's scroll value actually exists. Matches the PR description (carry previous scroll on new images; still restore an image's own saved viewport on revisit).

**Tests.** `test_keep_previous_zoom_keeps_viewport_position` is a solid GUI regression test: zooms to 300%, sets scroll to the midpoint on both axes, advances to the next image, and asserts both zoom and scroll carried over. `CHANGELOG.md` entry is present and correctly cross-referenced.

No issues found. This PR is currently **blocked on CI**: only the `cla` check has run because it's a fork PR — the test/lint/network workflows are queued pending a maintainer's approval to run for first-time/external contributors. That approval, not a code change, is what's needed to move this forward.

wkentaro:

Verdict: recommend-merge. CI is now green across all checks (the workflows needed maintainer approval for this first-time contributor, which has since been granted). The prior review found no issues: the fallback correctly defers to an image's own saved viewport, only applies when `keep_prev_scale` is on and no prior value exists, and is covered by a solid GUI regression test.

wkentaro:

## Review

**Correctness.** Traced the zoom/scroll state machine in `labelme/_app.py`: the new `_remember_current_viewport()` correctly reads live scrollbar values before `reset_state()` and is wired into all four navigation entry points (`_load_file`, `close_file`, `import_dropped_image_files`, `import_folder_images`), so the carry-forward and per-image restore both work as described. No enforced invariant stops a future loader that sets `_image_path` directly from silently skipping this, but that's a latent risk, not a bug today.

One real, untested edge case: `target_scroll_values` is computed and applied via `set_scroll_value` *before* `_paint_canvas()` resizes the canvas and rescales the scrollbar's `maximum()` for the new image. For a differently-sized, previously-unvisited image, `QScrollBar.setValue` clamps to the stale (larger) previous maximum at assignment time; whether centering fidelity survives the later resize isn't verified. The test fixtures (`2011_000003.jpg` 500×338, `2011_000006.jpg` 500×375) do differ in size but the fractional scroll positions used never push the carried value past the second image's actual max, so this path is never exercised despite fixtures that could exercise it.

**Tests.** `tests/e2e/zoom_test.py` ran green (13 passed). Coverage is solid for next/prev, search-jump, drag-and-drop load, and close+reopen, but misses: the resize-before-clamp edge case above, and first-ever-opened image with `keep_prev_scale=True` and no `_prev_image_path` (falls through cleanly in the code, but untested).

**Standards.** No hard violations. One judgement call: `_remember_current_viewport` duplicates the per-orientation scroll-bar iteration shape already present in `set_scroll_value` and the old `_load_file` scroll block — three near-identical loops now exist. Not blocking.

**Changelog.** Correct — under `## [Unreleased]` → `### Fixed`, links `#2484`.

## CI

`Lint` and `Test` both show `action_required` on the latest push — GitHub is withholding Actions runs pending maintainer approval for this first-time contributor (this needs re-approval on every new push, not just the first). This agent can't approve workflow runs, so the review above is based on a local checkout, not a green CI run.

Verdict: parking on `needs-info` until CI is authorized and runs. @wkentaro: approving the pending workflow run in the Actions tab would unblock this.

wkentaro (approved):

PR #2490, opened 2026-08-17
ci: gate releases on the test workflow

wkentaro:
Stacked on #2471.

A tag push never triggered the Test workflow, so a release could publish from a commit whose tests never ran. This makes the Test workflow callable and has the release job depend on it, so publishing a tag requires the full OS and Python matrix, the artifact smoke jobs, and the network test to pass on the tagged commit first.

The release job's own artifact validation step stays: the called workflow builds its own distributions on its own runners, while that step checks the exact files uploaded to PyPI.

## Test plan

- [x] yamlfix and actionlint pass on both workflows
- [ ] The `workflow_call` path only truly executes on a tag push; the first release candidate tag is its live trial

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,10 @@ on:
   push:
     tags: [v*]
 jobs:
+  test:
+    uses: ./.github/workflows/test.yml
   release:
+    needs: test
     runs-on: ubuntu-latest
     environment: pypi
     permissions:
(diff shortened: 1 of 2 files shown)


wkentaro:
This PR was auto-marked as merged by GitHub during a stack reorder (its commit became reachable from the then-base branch). The change now lives at the bottom of the stack as #2492.

PR #2492, opened 2026-08-17
ci: gate releases on the test workflow

wkentaro:
Bottom of a stack; #2471 builds on this. Replaces #2490, which GitHub auto-marked as merged during the stack reorder.

A tag push never triggered the Test workflow, so a release could publish from a commit whose tests never ran. This makes the Test workflow callable and has the release job depend on it, so publishing a tag requires the full OS and Python matrix and the network test to pass on the tagged commit first.

## Test plan

- [x] yamlfix and actionlint pass on both workflows
- [ ] The `workflow_call` path only truly executes on a tag push; the first release candidate tag is its live trial

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,10 @@ on:
   push:
     tags: [v*]
 jobs:
+  test:
+    uses: ./.github/workflows/test.yml
   release:
+    needs: test
     runs-on: ubuntu-latest
     environment: pypi
     permissions:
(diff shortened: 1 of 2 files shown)


wkentaro:

Verdict: recommend-merge. Small, self-contained CI change: adds a `workflow_call` trigger to the Test workflow and makes the release job `needs: test`, exactly per the PR description. No runtime code touched, so no smoke test applies. Standards and spec review found nothing to flag. All checks green on the latest run (two earlier flaky failures on macos/ubuntu 3.13 cleared on rerun). Note the PR's own test-plan caveat: the `workflow_call` path only truly executes on a tag push, so its first real trial is the first release candidate tag after merge.

PR #2496, opened 2026-08-18
fix: save cross-drive annotations with an absolute imagePath

### LINKED ISSUE #1350: Labelme does not allow saving of files to a separate Mount (separate hard disk)

3.9.13

### What OS are you using?

Windows 11

### Describe the Bug

When trying to save a labelme file on another hard disk (D drive), it crashed.

### Expected Behavior

_No response_

### To Reproduce

_No response_

wkentaro:
Saving an annotation to a different drive than the image still loses the annotation: #2393 stopped the crash but only rerouted the `ValueError` from `os.path.relpath` into the save-error dialog. Two Windows drives have no relative path between them, so there is nothing for `imagePath` to hold in relative form.

When no relative path exists, store an absolute one. Every other save keeps the relative path.

Three non-obvious bits:

1. The reader needs no change: both join sites resolve `imagePath` against the label directory with pathlib `/`, an absolute right operand wins that join, and the reader's `PureWindowsPath(...).as_posix()` normalization preserves drive and UNC absoluteness. The reopen half of the new e2e test proves it.

2. UNC shares raise the same `ValueError` (share-vs-drive and share-vs-share) and take the same fallback; safe. The branch runs un-mocked on Windows CI via the new `skipif` unit test — `relpath` never raises across drives on POSIX, so it cannot be exercised here.

3. An absolute path does not survive moving the files to another machine, which only a relative path does — on another OS it fails loudly through the existing read-error dialog, not silently. A failed save survives nothing, so the fallback stays scoped to the pair that has no relative form.

The issue's triage note says this crashes through the global excepthook; true of the reported v5 traceback, stale for current `main`, so `__main__.py` needed no change.

Closes #1350


wkentaro:

Rebased onto `main` (touched files overlapped: `labelme/_app.py`, `CHANGELOG.md`) and resolved a changelog ordering conflict. Ran the medium-effort `/code-review` gate on both axes (Standards, Spec) — no meaningful findings, only judgement-call nits (a now-unreachable `ValueError` arm left as a documented backstop; the new e2e test exercises the absolute-path join via a POSIX path rather than a real Windows drive letter, which is a test-fidelity note, not a defect). Commit history was already clean (code commit + docs commit), so `/recommit` was a no-op. Re-ran the new unit and e2e tests locally — all pass. CI is green across the full OS/Python matrix after the rebase push.

**recommend-merge**

wkentaro:

Verdict stale (branch was rebased again after the prior verdict comment, main having advanced past that point); re-ran the finalize pipeline against the current head.

Re-ran the medium-effort `/code-review` gate on both axes:
- **Standards**: no hard violations. The extracted `_resolve_stored_image_path` helper is single-caller, `_`-prefixed, fully annotated, keyword-call-site — matches house conventions. No meaningful smells.
- **Spec**: the two verified non-obvious claims (absolute-path join precedence, cross-OS failure mode) check out against the actual code. One gap noted: the UNC-share fallback (claim 2 in the PR description) has zero test coverage on either platform — not a defect, but the claim is asserted, not proven. Worth a follow-up if UNC paths matter in practice; not a merge blocker.

Commit history already clean (code commit + docs commit); `/recommit` a no-op. CI is green across the full OS/Python matrix at the current head, merge state CLEAN.

**recommend-merge**

wkentaro:

follow-ups: #2505 (UNC-share fallback path has no test coverage on either platform)

PR #2497, opened 2026-08-18
fix: normalize the paths of dropped image files

wkentaro:
On Windows, `QUrl.toLocalFile` reports a dropped path with forward slashes while every other path in the File List carries the platform separator, so dropping an image whose directory was also opened listed the same file twice. Normalizing at the drop boundary keeps the File List's separator invariant in one place.

Split out of #2467, whose Windows matrix cells surfaced the bug; that PR is stacked on this one.

## Test plan

- [x] `make lint` and `make test`
- [x] The nine-cell matrix run in #2467 exercises this commit on Windows


wkentaro (commented):
**Verdict: recommend merge.** Two-axis review (Standards / Spec) of e2325ea against `main`: no hard violations on either axis; both agents verified the claimed mechanism against the code, not just the commit message. Findings below are judgement calls, none blocking.

## Standards

**CHANGELOG.md — compliant.** The entry sits under `## [Unreleased]` / `### Fixed` in Keep a Changelog format with the PR number linked, and the change is user-facing so it belongs there. Minor: the Fixed list is otherwise roughly newest-first and this entry was inserted second, but no ordering rule is documented.

**Code hunk — premise verified.** `os` is imported; the directory scan normalizes via `os.path.normpath` (`labelme/_app.py:3109`) and `import_dropped_image_files` dedupes by exact string membership (line 2870), so the separator mismatch was real and the fix closes it. The comment explains the why; naming `QUrl` (a stable external Qt class) is within the spirit of the comment conventions.

**Judgement call — possible shotgun surgery.** The "File List paths carry the platform separator" invariant is enforced by convention at scattered producers (scan at 3109, drop at 2306) rather than at one boundary. Sufficient for the stated bug, but the next producer must remember the rule; normalizing inside `import_dropped_image_files` would centralize it.

## Spec

**Mechanism confirmed as described.** A forward-slash drop path fails the `path not in already_loaded` membership test and gets `addItem`-ed a second time; both File List write paths (lines 2875 and 2919) are now normalized, so the stated invariant holds — CLI and dialog directory opens funnel through the normalizing scan.

**Partial vs. the body's framing.** "Keeps the File List's separator invariant in one place" — the normalization sits in `dropEvent`, not in `import_dropped_image_files`, the method that actually writes the list. No bypassing caller exists today, so this is robustness, not a behavior gap.

**Test-plan checkbox is externally verified only.** No test in this PR touches `dropEvent` or `import_dropped_image_files`; the Windows coverage lives entirely in the stacked #2467 matrix, as the body says.

**No scope creep; no defects.** `normpath` side effects (collapsing `..`, trailing slashes) are theoretical here since drag-drop `toLocalFile` yields clean absolute paths, and the scan side already applies identical normalization. Leaving `dragEnterEvent` raw is correct scoping (it only checks extensions).

**Adjacent, pre-existing, out of scope:** `self._image_path` set from a Qt file dialog (also forward slashes on Windows) is compared to File List text via `findItems(..., MatchExactly)` at lines 1789–1795, so the saved-checkmark update can silently miss on Windows. Not introduced or claimed fixed by this PR; worth a follow-up issue.

## Summary

Standards: 0 violations, 2 judgement calls (worst: invariant enforced by convention at scattered producers). Spec: 0 missing requirements, 2 notes (worst: in-PR test coverage is zero — verification rides on the stacked #2467 matrix).

PR #2499, opened 2026-08-18
fix: normalize the paths of images loaded via file dialogs

wkentaro:
Surfaced while reviewing #2497: the saved checkmark and the file list row selection both compare `_image_path` against File List items as exact strings, but Qt file dialogs report a chosen path with forward slashes on Windows, so an image opened through the dialog neither selected its row nor got the checkmark after a save. Normalizing where paths enter loading state closes the last un-normalized producers of the invariant #2497 states.

Two non-obvious bits:

1. The label-file branch also normalizes, because a stored relative image path can carry `.` or `..` components that survive the pathlib join; separators alone were already re-rendered by pathlib, so this only collapses dot components.

2. The regression test spells the mismatch with a redundant `.` component instead of forward slashes, since on POSIX the forward slash is the platform separator and the Windows spelling cannot mismatch there.

Stacked on #2497; merge that first.

## Test plan

- [x] New regression test in `tests/e2e/file_dialog_actions_test.py`, red without the fix
- [x] `make lint` and `make test`


wkentaro (commented):
**Verdict: LGTM.** Every claim in the PR body checked out against the code, and I reproduced the red/green cycle locally.

Verified:

- Both consumers named in the body really do compare exact strings: the saved checkmark uses `findItems(self._image_path, MatchExactly)` and row selection uses `current_image_path in self.image_list` / `.index(...)`. An unnormalized `_image_path` misses both.
- `_load_file` is the single choke point for the dialog flows: `_open_file_with_dialog` and `_open_dir_with_dialog` both route through `_load_from_file_or_dir`, whose non-label branches end in `_load_file`, and the file-list selection handler re-enters it with already-normalized item text (normpath is idempotent, so no harm). The only other `_image_path` producers — `_open_label_file_into_state`, `_open_image_into_state` (fed by `_load_file`), and the drop path from #2497 — are all normalized after this change.
- The label-file rationale holds: pathlib drops `.` and re-renders separators but keeps `..`, so the added normpath in `_open_label_file_into_state` is what collapses dot components surviving the join.
- Ran the new test on the PR head: passes in 5.6s. Reverted `labelme/_app.py` to the parent commit and reran: fails, and the failure log shows exactly the claimed bug — the app loads `.../raw/./2011_000006.jpg` directly instead of selecting the listed row.
- CI is green across the full matrix including Windows. Changelog entry is present in the right subsection with the PR link.

One non-blocking note: `os.path.normpath` collapses `..` lexically, so a label file whose relative image path climbs through a symlinked directory could resolve to a different file than the OS would. That spelling was broken before this PR too (it never matched the file list), so not a regression — just worth remembering if a symlink report ever comes in.

Could not verify directly (no Windows box here): that Qt dialogs report forward slashes on Windows. Consistent with documented Qt behavior (`QDir::toNativeSeparators` exists for exactly this), and the drop-path precedent in #2497.

PR #2502, opened 2026-08-18
refactor(canvas): copy the dragged points into the moved shape

### LINKED ISSUE #2413: refactor(canvas): _apply_in_place_move aliases the clone's points array into the original Shape

Noticed while reviewing #2410, which establishes the invariant that a `Shape` owns its `points` / `point_labels` arrays (the constructor now copies with `np.array` instead of `np.asarray`).

`Canvas._apply_in_place_move` bypasses that invariant by rebinding the array object directly:

```python
def _apply_in_place_move(self) -> None:
    for original, clone in zip(self.selected_shapes, self._selected_shapes_copy):
        original.points = clone.points
```

`labelme/_widgets/canvas.py:1221-1223`

After this runs, `original` and `clone` share one buffer. It is harmless today only because the caller clears `self._selected_shapes_copy` on the very next line, so the clone is dropped and no second live holder remains. That is a timing accident rather than a property of the code: any future change that keeps a clone alive (undo/redo snapshots, a longer-lived copy buffer, multi-step drag) would silently reintroduce exactly the aliasing bug #2410 just closed at the constructor.

Suggested fix: copy on assignment (`original.points = clone.points.copy()`), or route the move through a `Shape` method that owns the write, so the invariant is enforced in one place rather than depending on the caller discarding the clone.

No user-visible bug is known to result from this today; this is a latent-invariant cleanup.

wkentaro:
Preserves `Shape` point-array ownership when committing an in-place move. The copy-drag path already promotes the clone and is unchanged.

Closes #2413

## Test plan

- `make test` (1290 passed, 6 skipped)
- `make lint`

### DIFF
--- a/labelme/_widgets/canvas.py
+++ b/labelme/_widgets/canvas.py
@@ -1272,7 +1272,7 @@ class Canvas(QtWidgets.QWidget):
 
     def _apply_in_place_move(self) -> None:
         for original, clone in zip(self.selected_shapes, self._selected_shapes_copy):
-            original.points = clone.points
+            original.points = clone.points.copy()
 
     def _can_close_shape(self) -> bool:
         if self.mode != _CanvasMode.CREATE:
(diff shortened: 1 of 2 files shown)


wkentaro:

Rebased onto `main` (was `BEHIND`). Ran a report-only `/code-review` (Standards + Spec, medium effort): both axes clean, no findings — the fix is exactly the `.copy()` the linked issue suggested, correctly scoped, `point_labels` has no analogous live-aliasing hazard to fix. `/review-fix` skipped as unnecessary. Single commit was already clean, `/recommit` skipped. Refactor with no behavior change and full GUI-marked regression test coverage already green in CI, so no additional `/verify` smoke. CI green on all platforms post-rebase.

PR #2503, opened 2026-08-18
refactor(app): warn once per shape with a non-str label

### LINKED ISSUE #2415: refactor(app): hoist the non-str label guard out of the per-pattern loop in _shapes_from_dicts

Noticed while reviewing #2414, which adds test coverage for `_shapes_from_dicts`.

In `labelme/_app.py`, the `isinstance(shape.label, str)` guard sits inside the loop over `label_flags.items()`, but it does not depend on the loop variables:

```python
default_flags: dict[str, bool] = {}
if label_flags:
    for pattern, keys in label_flags.items():
        if not isinstance(shape.label, str):
            logger.warning("shape.label is not str: {}", shape.label)
            continue
        if re.match(pattern, shape.label):
            for key in keys:
                default_flags[key] = False
```

`Shape.label` is `str | None` (`labelme/_shape.py:32`), so the guard is load-bearing. But being loop-invariant, it re-evaluates once per configured pattern and emits one warning per pattern for a single offending Shape. A user with N `label_flags` patterns gets N identical log lines for one bad label.

Hoisting it above the loop keeps the same behavior (all patterns are skipped either way) and logs once:

```python
default_flags: dict[str, bool] = {}
if label_flags and isinstance(shape.label, str):
    for pattern, keys in label_flags.items():
        if re.match(pattern, shape.label):
            for key in keys:
                default_flags[key] = False
elif label_flags:
    logger.warning("shape.label is not str: {}", shape.label)
```

Purely a readability and log-noise cleanup; no behavior change to the resulting flags. The merge contract is now pinned by the tests in #2414, so the refactor is covered.

wkentaro:
Closes #2415.

The label type check did not depend on the pattern loop it sat in, so one shape with a non-str label logged one identical warning per configured pattern.

One deliberate behavior change: the hoisted guard warns even when `label_flags` is empty or every pattern was dropped, where `main` stayed silent. A non-str label is anomalous regardless of the flag config, and `read_label_file` raises `TypeError` on one anyway, so only a caller bypassing the reader reaches the warning at all — which is what the new test does via `cast`.

Worth a follow-up, not done here (#2415 scopes guard removal out): since the guard is unreachable through the app, it could be deleted outright.

## Test plan

- [x] `test_shapes_from_dicts_warns_once_per_shape_with_a_non_str_label` fails against `main` (4 warnings vs the 2 asserted)
- [x] `make test` (1251 passed) and `make lint`
- [x] Loading `examples/tutorial/apc2016_obj3.json` with real `label_flags` yields flags byte-identical to `main` (the str-label path is unchanged code motion)

### DIFF
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -2955,13 +2955,13 @@ def _shapes_from_dicts(
         )
 
         default_flags: dict[str, bool] = {}
-        for pattern, keys in compiled_label_flags.items():
-            if not isinstance(shape.label, str):
-                logger.warning("shape.label is not str: {}", shape.label)
-                continue
-            if pattern.match(shape.label):
-                for key in keys:
-                    default_flags[key] = False
+        if not isinstance(shape.label, str):
+            logger.warning("shape.label is not str: {}", shape.label)
+        else:
+            for pattern, keys in compiled_label_flags.items():
+                if pattern.match(shape.label):
+                    for key in keys:
+                        default_flags[key] = False
         shape.flags = default_flags
         shape.flags.update(shape_dict["flags"])
         shape.other_data = shape_dict["other_data"]
--- a/tests/unit/_app_test.py
+++ b/tests/unit/_app_test.py
@@ -4,9 +4,11 @@ import struct
 import zlib
 from collections.abc import Callable
 from pathlib import Path
+from typing import cast
 
 import numpy as np
 import pytest
+from loguru import logger
 from PySide6 import QtCore
 from PySide6 import QtGui
 from PySide6 import QtWidgets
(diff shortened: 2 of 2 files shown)


wkentaro:

Verdict: recommend-merge.

Standards and spec review (medium effort, no fixes) came back clean — no meaningful findings. Two notes, both non-blocking:
- The warn-guard's `elif compiled_label_flags:` check is a slightly indirect proxy ("would matching have been attempted") but is explained by its adjacent comment and mirrors pre-existing behavior.
- One acceptance-criterion bullet ("warns once for a non-str label") is met in substance, not letter: with an empty `compiled_label_flags`, the new code logs zero warnings, matching `main`'s actual behavior (not the issue's illustrative snippet). Deliberate and pinned by the `every-pattern-dropped` test case.

Single commit already, Conventional Commits format — no `/recommit` needed. Change is internal/unreachable through the app per the PR body, so no `/verify` smoke. All 19 CI checks green, merge state CLEAN.

PR #2506, opened 2026-08-18
fix: keep AI oriented rectangles inside image bounds

wkentaro:
First in a two-PR stack; #2466 follows with the RGB inference fix. Corrected RGB input changed the model mask fit and exposed a latent bounds bug: minimum-area rectangles and bounding-box fallbacks can overhang an Image edge when out-of-bounds points are disabled.

Out-of-bounds behavior follows the Canvas Setting: the default fits oriented rectangles to the Image, while enabling the beta Setting preserves the model-generated vertices. Passing `image_size=None` to the detection-to-shape conversion is the single opt-out; there is no separate flag.

Two safety details:

1. A rectangle that fits is translated rigidly, preserving its dimensions and rotation; clipping vertices individually would deform it into a non-rectangle the editor cannot represent.

2. Clipping applies only when an axis is larger than the Image, where translation alone cannot fit it.

## Test plan

- [x] Unit regression tests for rigid translation, oversized-axis clipping, and the bounding-box fallback; a Canvas test pins the Setting-to-`image_size` wiring
- [x] AI Text Prompt e2e matrix gains an edge-crossing oriented-rectangle case asserting in-bounds output
- [x] `make lint`
- [x] `make test` on the stacked tree (1268 passed, 6 skipped)
- [ ] Network e2e via CI (behavior-identical rework; the previous tree passed 10/10 against real cached models)


wkentaro:
> *This was generated by AI while finalizing the PR.*

Two competing versions of the edge-crossing geometry were force-pushed to this branch today; recording the resolution so it does not ping-pong.

**Kept:** rigid translation — slide the min-area rectangle inside the Image, clip only an axis larger than the Image. **Rejected:** replacing any edge-crossing fit with its clipped axis-aligned bounding box. The failure this PR fixes was a 1.7px overhang (x=339.7 in a 338px Image); the axis-aligned variant discards the rectangle's orientation for that sub-2px overhang, defeating the point of the oriented-rectangle output format, while rigid translation nudges it 1.7px and keeps the fit.

Both versions shared the same plumbing, which is kept: a single optional `image_size` on the detection-to-shape conversion (default `None` = no fitting), decided at the two entry points by the out-of-bounds Setting. The rejected variant also dropped detections lying fully outside the Image; with rigid translation they slide to the edge, consistent with partial overhangs.

Branch is rebased onto current `main`; #2466 is rebased on top.

PR #2536, opened 2026-08-18
fix: offset duplicated and pasted shapes

### LINKED ISSUE #2527: fix(canvas): offset duplicated and pasted shapes
## Problem

In the packaged v7.1.0rc1 GUI, both Duplicate and Copy/Paste create a shape at exactly the source coordinates.

The new annotation is hidden beneath the source even though a new saved row exists, so users can unknowingly create overlapping duplicate annotations.

## Reproduction

1. Open an image with a rectangle.
2. Select the rectangle and invoke Duplicate.
3. Copy and paste the selected rectangle.
4. Compare the canvas placement and serialized points of the created shapes.

Both creation paths preserve the source coordinates exactly.

## What to build

Place duplicated and pasted shapes at a small deterministic canvas-space offset and keep the new copy selected.

## Acceptance criteria

- [ ] Duplicate creates a visibly offset copy within the image bounds.
- [ ] Copy/Paste creates a visibly offset copy within the image bounds.
- [ ] Repeated duplication or pasting remains discoverable and does not place a copy outside the image.
- [ ] The new copy is selected so it can be moved or edited immediately.
- [ ] Saving and reloading preserves the offset geometry.
- [ ] Regression coverage exercises both Duplicate and Copy/Paste.

## QA provenance

- Finding: F-004
- Classification: functional defect
- Severity: P2
- Confidence: verified
- Environment: packaged PyPI wheel v7.1.0rc1, macOS 26.5.2 arm64
- Attempts: two consistent creation paths

wkentaro:
Closes #2527.

Duplicate and Paste placed the copy on the source coordinates, so the new shape sat hidden underneath and users could stack annotations without noticing. Both paths now search for a placement the copy can have to itself.

Four non-obvious bits:

1. The search starts at a zero offset, so pasting into a different image keeps the copied coordinates — transferring annotations between frames of one scene depends on that. In the same image the source occupies that spot, so Duplicate and Paste always move the copy; a copy carrying coordinates from a larger image is brought onto this one first.

2. The step is a fixed on-screen distance converted through the zoom, so the gap looks the same at any magnification and the saved geometry depends on the zoom level. Gaps under 3 screen pixels are refused rather than accepted as success.

3. The search widens the gap first, then narrows it once the image edge stops making room. It is finite by design: roughly 41 distinct placements on a 500x375 image and 103 on 1920x1080 before positions repeat under repeated Paste of one clipboard. When every placement is taken, or a shape spans the whole image, the copy stays inside the image rather than being pushed out of it — the issue asks for in-bounds copies, and this is the one case where a copy stays under its source.

4. `_compute_shapes_bounds` now unions the edges instead of calling `QRectF.united()`, which discards a zero-extent rect. A point shape draws with no width or height, so a mixed selection was dropping it from the group bounds and could carry it off the image. This also fixes the same latent bug on the group-drag path.

Verified against the packaged entry point on the annotated example, offscreen: clicking a shape then Duplicate and Copy/Paste each produced a copy 12 screen pixels clear of its source and left it selected, the saved JSON matched the canvas with every point inside the image, and Undo removed the copy.

three offset copies of one polygon

## Test plan

- [x] `make test` — 1330 passed, 1 skipped
- [x] `make lint` — ruff, ty, taplo, mdformat, yamlfix, typos
- [x] Regression coverage for both paths: repeated Duplicate and repeated Paste, cross-image paste, in-bounds after save and reload, group with a point shape, circle drawn extent, mask bounding boxes, shapes wider than the image, and the zoom-scaled gap

### DIFF
--- a/changelog.d/2536.fixed.md
+++ b/changelog.d/2536.fixed.md
@@ -1,0 +1,1 @@
+Fixed Duplicate and Paste placing a Shape on its source coordinates, where the copy was hidden underneath and easy to stack unnoticed. Both now offset the copy by a visible distance that stays inside the Image, while pasting into another Image keeps the copied coordinates
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -1823,6 +1823,11 @@ class MainWindow(QtWidgets.QMainWindow):
     def _insert_shapes(self, shapes: list[Shape]) -> None:
         if not shapes:
             return
+        # A copy landing on the shape it came from is invisible, and the user
+        # cannot tell that anything was created. This has to run before the
+        # shapes join the canvas, or they would read as occupying the very
+        # placement they are being offered.
+        self._canvas_widgets.canvas.offset_shapes_for_insertion(shapes=shapes)
         self._load_shapes(shapes=shapes, replace=False)
         self._canvas_widgets.canvas.select_shapes(shapes)
         self.mark_dirty()
--- a/labelme/_widgets/canvas.py
+++ b/labelme/_widgets/canvas.py
@@ -5,6 +5,7 @@ import dataclasses
 import enum
 import typing
 from collections.abc import Callable
+from collections.abc import Iterator
 from collections.abc import Sequence
 from typing import Any
 from typing import Final
@@ -12,6 +13,7 @@ from typing import Literal
 from typing import cast
 
 import numpy as np
+import numpy.typing as npt
 from loguru import logger
 from PySide6 import QtCore
 from PySide6 import QtGui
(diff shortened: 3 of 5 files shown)


wkentaro:

**Verdict: recommend-merge**

Cheap `/code-review` gate (Standards + Spec axes) ran on the diff vs `main`; no meaningful findings surfaced:

**Standards** — no hard violations. Two judgement-call duplication smells in `_iter_insertion_offsets`/`_compute_axis_containment` vs `_compute_axis_offset` (near-identical widen/narrow loops, a repeated low/high/extent data clump) — minor, not blocking.

**Spec** (issue #2527) — all acceptance criteria are met. One nuance worth naming for the record: paste only offsets when the target coordinates are already occupied by an identical shape; pasting into a different image intentionally keeps the exact copied coordinates (documented in the changelog fragment and covered by `test_paste_into_another_image_keeps_the_copied_coordinates`), a deliberate scope call rather than an oversight. There's also a theoretical edge case where >~199 repeated inserts on one image exhaust the offset search and settle on an occupied placement again — acknowledged in the code's own comment as an intentional inside-beats-outside tradeoff, not worth a fix at that scale.

No rebase needed (`main` hasn't advanced since the merge-base). Commit sequence is already clean (fix + changelog fragment). CI is green across all 20 checks.

wkentaro:
I think this is expected behavior because sometimes we want to just copy and then edit and make a group out of it, so I am just not sure if this is the right thing or should be a config or made more visible without shifting.

wkentaro:

Re-flagging as **recommend-triage**: the prior `recommend-merge` verdict is superseded by your own comment above raising an open product question — whether shifting pasted/duplicated shapes on overlap should be configurable, or made visible some other way instead of always shifting. That is a product call, not something this pass can decide. Code itself is unchanged and still green (CI, code-review gate from the prior tick). Landing this as-is versus reworking the UX is yours to make.

wkentaro:

**Verdict: wontfix (rejected)**

Closed on a product decision, not code quality: the code was sound and CI green. Duplicate and Paste keep the source coordinates on purpose. A common workflow is to duplicate a shape and edit the copy — label, flags, group id, a few vertices — while keeping the geometry identical to the source. An automatic offset makes that impossible, since the exact coordinates can never be recovered by hand.

The decision is recorded in `.out-of-scope/offset-duplicated-shapes.md` (#2607) so the request does not return as fresh code. Discoverability of a hidden copy is still a fair concern; acceptable fixes are ones that do not move the copy (highlight, list entry, status hint).

One piece of this PR is worth salvaging separately: the `QRectF.united()` fix for point shapes in group bounds, which is a real latent bug in the group-drag clamp on `main`. Filed as #2608.

PR #2551, opened 2026-08-23
feat: enforce AI model distribution policy

wkentaro:
## Why

Distribution packages need an explicit way to restrict optional AI models, while users need the model license and provenance before the first network download and afterward.

## What

- Add `LABELME_AI_MODEL_ALLOWLIST` and enforce it at the shared download and inference boundaries.
- Show a lightweight Cancel/Download confirmation with exact OSAM license and source links plus the bundled full license for offline reading.
- Keep model information accessible from both AI panels and disable unavailable choices without changing the settings schema.
- Require OSAM 0.6, which packages the canonical model license files.

## Test plan

- [x] `make lint`
- [x] `make test` (1,314 passed, 6 skipped)
- [x] Qt tests cover the Cancel default, Download continuation, clickable pinned links, bundled licenses, filtered GUI choices, and direct download/inference rejection.
- [x] Isolated app launch with a restricted allowlist.
- [ ] Pointer-level desktop QA: the Computer Use client/server version mismatch persisted after a full relaunch; proceeding with the explicitly accepted gap.


wkentaro:

CI's `lint` check was red: `check_translate` failed because this PR adds new `tr()` strings without regenerating the translation catalogs. Ran `make update_translate` and translated the 6 new source strings into all 20 target languages, then pushed the fix. Awaiting the re-triggered CI run before a merge verdict.

wkentaro:

**Verdict: recommend-close**

CI's `network-test` job timed out after 5m0s on the `Test real model download` step (`tests/unit/widgets/download_test.py::test_download_ai_model_from_network`).

Root cause traces to this PR's own change: `download_ai_model()` in `labelme/_widgets/download.py` now opens a blocking `_confirm_ai_model_download()` `QMessageBox.exec()` before downloading. No test or fixture dismisses this new dialog — `close_failed_download_dialog` in `tests/conftest.py` only closes a `QProgressDialog` labeled "Failed to download...", not the new confirm box — so the network test hangs until the job hits its 5-minute timeout.

A prior agent fix-attempt already landed on this branch (the `check_translate` fix), so per the standing process this goes to `recommend-close` rather than another auto-fix attempt. Fix direction: have the network test click through (or monkeypatch/bypass) the new confirm dialog before asserting the download, the same way the existing failure-dialog fixture handles the post-download error box.

PR #2581, opened 2026-09-02
refactor: build menus and toolbars from plain QAction lists

wkentaro:
Drops the private menu/toolbar filler that walked a mixed list where `None` meant a separator and a `QMenu` meant a submenu. Qt covers both natively: a `QAction` flagged as a separator renders as one, and `addActions` takes the whole list.

Two non-obvious bits:

1. Separator actions are owned by the main window, not the menu. The edit menu is cleared and rebuilt on every mode change, and `QMenu.clear()` deletes menu-owned actions, so window-owned separators survive the rebuild; verified by a unit test and an offscreen smoke run (file 2, view 6, edit 3 before and after rebuild, toolbars 4 and 1, all matching `main`).

2. `ToolBar.addAction` now routes separator actions to the base class the same way it already did for widget actions, so they are not wrapped in a `QToolButton`.

Submenu support is dropped from the helper because no caller passed a `QMenu`; `menu.addMenu` covers it if one ever does.

https://claude.ai/code/session_011sPH8v4aEj6KXCsaHCZeKy

### DIFF
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -165,8 +165,8 @@ class _Actions(NamedTuple):
     zoom: tuple[ZoomWidget | QtGui.QAction, ...]
     on_load_active: tuple[QtGui.QAction, ...]
     on_shapes_present: tuple[QtGui.QAction, ...]
-    context_menu: tuple[QtGui.QAction | None, ...]
-    edit_menu: tuple[QtGui.QAction | None, ...]
+    context_menu: tuple[QtGui.QAction, ...]
+    edit_menu: tuple[QtGui.QAction, ...]
 
 
 class _Menus(NamedTuple):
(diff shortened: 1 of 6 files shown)


wkentaro:
Visual QA verdict: **PASS** — no visual breakage found at `91f00a467`.

Checked on macOS with an isolated profile:

- Main-window launch and image rendering.
- Horizontal and vertical toolbar layout, action enablement, and separator rendering.
- File, Edit, View, Help, and macOS-relocated About/Settings/Quit actions.
- Edit-menu rebuild after switching Rectangle → Polygon; all actions remained present and correctly enabled.
- Rectangle and polygon creation, label acceptance, autosave, polygon undo, clean quit, and restart persistence. Saved JSON retained only the rectangle after undo.
- Targeted Qt utility/toolbar tests: `61 passed`.

Not covered: AI model inference/downloads, remaining shape types, theme, and accessibility sweeps.

PR #2596, opened 2026-09-02
refactor(canvas): share one click handler across open-path modes

wkentaro:
## What

Polygon, linestrip, and AI-Points each had their own branch in the canvas click handler for growing the draft. The three bodies differed only in how the draft ends: a polygon closes when the click lands on its first vertex, and the other two end on Ctrl. One handler now promotes the preview's cursor end into the draft and asks a small predicate whether that click was the last, so the per-mode branches are gone.

## Behavior

- Click, Ctrl+click, and polygon self-closing behave as before for all three modes; the annotation and canvas-interaction e2e tests pass unchanged.
- Linestrip now treats Ctrl as a modifier bit, so Ctrl+Shift+click also places the last point, matching AI-Points.
- After undoing a committed polygon or linestrip, the preview segment carries point labels for both ends, matching the invariant the paint path already asserts.

## Tests

The first commit adds canvas unit tests that pin the behavior the shared handler has to keep: a click grows a polygon or linestrip and restarts the preview from the new vertex, Ctrl finishes a linestrip but not a polygon, a click on the first vertex closes a polygon, an AI-Points click carries the preview's Shift label into the draft and the next preview, and a click right after undoing a committed polygon closes it again. All of them pass against the pre-refactor handler. The refactor commit extends the linestrip Ctrl test to Ctrl+Shift, since that is the one behavior this PR changes.

Each of the five one-line mutations of the new handler (drop the vertex label, drop the preview labels, exact-Ctrl compare, no autoclose, revert the undo label fix) now fails at least one unit test; before, all five passed the full suite.

`make lint` passes. `uv run --no-sync pytest tests/unit` passes (1127 passed), and the non-network annotation and canvas-interaction e2e tests pass (37 passed).

### DIFF
--- a/changelog.d/2596.changed.md
+++ b/changelog.d/2596.changed.md
@@ -1,0 +1,1 @@
+Polygon, linestrip, and AI-Points clicks share one canvas handler; Ctrl+Shift+click now also places the last linestrip point, matching AI-Points
--- a/labelme/_widgets/canvas.py
+++ b/labelme/_widgets/canvas.py
@@ -1073,45 +1073,45 @@ class Canvas(QtWidgets.QWidget):
         self, *, current: _DraftShape, event: QtGui.QMouseEvent
     ) -> None:
         mode = self.create_mode
-        modifiers = event.modifiers()
-        if mode == "polygon":
-            current = current.add_point(self._line.points[1], autoclose=True)
-            self._current = current
-            self._line = dataclasses.replace(
-                self._line, points=(current.points[-1],) + self._line.points[1:]
-            )
-            if current.closed:
-                self._finalize()
+        if mode in ("polygon", "linestrip", "ai_points_to_shape"):
+            self._commit_preview_vertex(current=current, event=event)
         elif mode == "oriented_rectangle":
             if len(current.points) == ORIENTED_RECTANGLE_POINT_COUNT:
                 self._finalize()
             else:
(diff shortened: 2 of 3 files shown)


wkentaro (commented):
Reviewed head f358820. Verdict: recommend-merge.

Verified against the code, not the description:

- The shared handler is behavior-preserving for polygon and linestrip. Every path that sets the preview segment for those modes (start, mouse move, undo, and now commit) leaves its cursor label at 1, so reading the label from the preview instead of defaulting it changes nothing for them. For AI-Points it is the same read the old branch already did.
- Skipping the preview update when the click finalizes is safe: the point-prompt rejection is checked before extending, and every other finalize path either resets the draft or cancels it, and the next start rewrites the preview.
- The only intended behavior change is linestrip Ctrl+Shift+click, which now finishes the strip. It is covered by the new parametrized test and named in the changelog fragment.
- The polygon undo path now fills preview labels from the reopened draft. Previously the stale labels happened to be length 2 already, so this fixes an invariant rather than a crash.

Ran locally on the PR head: the 8 new canvas tests pass, and the full unit suite passes (1127 passed). Spot-checked three of the five claimed mutations (exact-Ctrl compare, no autoclose, dropped vertex label); each fails at least one of the new tests.

CI is green apart from one Windows test job still pending at review time.

Issue #2598, opened 2026-09-02
ZeroDivisionError on window resize after Close File in fit-window or fit-width mode


wkentaro:
## Steps

1. Open an image, choose Fit Window (or Fit Width).

2. File > Close.

3. Resize the main window.

## Observed

`resizeEvent` still sees `_zoom_mode == FIT_WINDOW` and `_image.isNull()` False, calls `_adjust_scale()`, and `_fit_window_scale` divides by the cleared pixmap's width:

```
ZeroDivisionError: Error calling Python override of QMainWindow::resizeEvent(): float division by zero
  labelme/_app.py:2339 in _fit_window_scale
    scale_by_width = available_w / pixmap.width()
```

## Cause

`close_file` runs `reset_state`, which clears the canvas pixmap and `_image_path` but never resets `_image`, so the resize guard treats the window as still having an image. `_set_zoom` already uses `_image_path is None` as the "nothing open" predicate; the resize guard should use the same one (or `reset_state` should clear `_image`).

## Repro test

```python
def test_resize_after_close(*, qtbot, main_win, data_path) -> None:
    win = main_win(file_or_dir=str(data_path / "raw/2011_000003.jpg"))
    show_window_and_wait_for_imagedata(qtbot=qtbot, win=win)
    win.set_fit_window_mode(True)
    win.close_file()
    win.resize(win.width() + 50, win.height() + 50)
    qtbot.wait(50)
```

Reproduced on `main` at bf2d0085.

wkentaro:
Released in [v7.4.0](https://github.com/wkentaro/labelme/releases/tag/v7.4.0).

PR #2599, opened 2026-09-02
refactor: compute zoom scale in one place and simplify resize guard

wkentaro:
Percent-to-factor conversion for the zoom now lives in one place, and the canvas owns its own resize-and-repaint when its scale changes, so the main window just assigns one to the other.

Three non-obvious bits:

1. `resizeEvent` now calls `super()` first and early-returns. `QMainWindow` does not override `resizeEvent` and `QWidget`'s is empty, and the central widget already has its new geometry on entry, so the fit calculation reads the same viewport size either way; safe. The dropped canvas-truthiness conjunct was always true (a `QWidget` has no `__bool__`).

2. The `Canvas.scale` setter fires unconditionally, on purpose. The file-load path assigns an unchanged zoom to force a geometry refresh for a differently sized image, and a unit test pins that so an equality short-circuit cannot sneak in later.

3. The middle-drag e2e helper drops its manual `adjustSize()`/`update()` after writing `canvas.scale`; the setter does that now.

While verifying the resize guard I found a pre-existing crash on resize after Close File in fit mode; it is unchanged by this PR and tracked in #2598.

No changelog fragment: no user-facing behavior change.

https://claude.ai/code/session_012Tk3PrFziUKbanji3ok7zG

### DIFF
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -290,7 +290,9 @@ class MainWindow(QtWidgets.QMainWindow):
 
         self._setup_app_state(file_or_dir=file_or_dir, output_dir=output_dir)
 
-        self._canvas_widgets.zoom_widget.valueChanged.connect(self._paint_canvas)
+        self._canvas_widgets.zoom_widget.valueChanged.connect(
+            self._apply_zoom_to_canvas
+        )
 
         self.populate_mode_actions()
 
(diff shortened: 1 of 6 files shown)


wkentaro:
Verdict: recommend-merge.

Computer Use comparison against base e28103f6 and head 7d83adb6 found the zoom behavior preserved.

- Fit to Window started at 196.0% and reached 150.2% after the same window resize on both revisions, with the full image still visible.

- Switching between differently sized images at an unchanged 100% zoom resized the canvas correctly on both revisions.

- Fit to Width reached 143.4% at the same window size on both revisions, filled the horizontal viewport, and exposed only the expected vertical scrollbar.

- The focused canvas, zoom-widget, zoom e2e, and middle-drag suites passed: 204 tests. All required PR checks are green.

Non-blocking standards note: the new labelme/_app.py comment names the Qt signal valueChanged in prose, contrary to the AGENTS.md rule to keep identifiers out of comment prose. Rewording it to “Setting the value rescales the canvas synchronously.” would remove that drift risk; it does not affect behavior.

PR #2602, opened 2026-09-02
Keep Previous Scale: the previous image's zoom wins over the zoom stored for this image

### LINKED ISSUE #2601: Per-image zoom memory overrides "Keep Previous Scale" once an image has been visited

labelme 7.3.0 (PyPI), Python 3.13, PyQt; reproduced on current main (e28103f6).

### What OS are you using?

Ubuntu 24.04

### Describe the Bug

Zoom is a property of the viewer, not of the image: when I zoom in, I am moving closer to the picture, and I expect to stay at that distance as I step through a folder of same-sized frames.

Labelme remembers a zoom level per image, and once an image has been visited, going back to it snaps to whatever zoom it had at the time, ignoring the current zoom and the "Keep Previous Scale" setting. Stepping through 50 frames from one camera, the view jumps between fit-to-window and various magnifications on every Prev/Next, and I have to press Ctrl+F on each one.

In `load_file`, the viewport stored for the image being opened is looked up first, and the previous image's viewport is used only when none is stored for this image. So "Keep Previous Scale" only has an effect on images never opened before; with the whole folder visited once it does nothing.

### Expected Behavior

With "Keep Previous Scale" on, the current zoom applies to every image until I change it. The per-image memory should apply only when the setting is off (or there is no previous image).

### To Reproduce

1. Open a folder of same-sized images, tick View > Keep Previous Scale.
2. Zoom to 170 % on image 1. Next.
3. Ctrl+F (fit window) on image 2. Prev, Next, Prev.
4. Image 1 is always at 170 % and image 2 always fit-to-window, whatever zoom was current when Prev/Next was pressed.

Usimian:
Fixes #2601.

With `keep_prev_scale` on, `load_file` looked up the viewport stored for the image being opened first and used the previous image's only when none was stored. Once every image in a folder had been visited, Prev/Next always landed on each image's own remembered zoom and the setting had no effect.

This swaps the priority: when the setting is on, the previous image's viewport is preferred; the stored one is used when the setting is off or there is no previous image. Behaviour with the setting off is unchanged.

Eight lines in `labelme/_app.py`, no new state. We are running this patch locally while labelling a folder of camera frames.

### DIFF
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -2260,13 +2260,15 @@ class MainWindow(QtWidgets.QMainWindow):
         self._canvas_widgets.canvas.setEnabled(True)
         # Zoom changes the live scroll positions, so resolve the intended
         # viewport first.
-        target_viewport = self._viewport_states.get(self._image_path)
-        if (
-            target_viewport is None
-            and self._config["keep_prev_scale"]
-            and self._prev_image_path is not None
-        ):
+        # "Keep Previous Scale" means the zoom follows the viewer, not the
+        # image: the previous image's viewport wins over one stored for this
+        # image, otherwise every image visited once snaps back to its own
+        # remembered zoom on Prev/Next and the setting has no effect.
+        target_viewport = None
+        if self._config["keep_prev_scale"] and self._prev_image_path is not None:
             target_viewport = self._viewport_states.get(self._prev_image_path)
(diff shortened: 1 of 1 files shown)


Usimian:
Doesn't matter, I forked it and fixed it locally, good luck.

PR #2605, opened 2026-09-03
refactor(render): scale mask outline once, size pens at construction

wkentaro:
Finishes the mask-outline half of #2600: the contour is now one image-space path mapped through a single scale transform, like the two-point shapes already are, and the stroke width lives with the module that strokes rather than on the per-shape render context.

Three things the diff does not spell out:

1. `QPainterPath.addPolygon` adds an unclosed subpath, so it emits exactly the `moveTo` plus `lineTo` chain the old loop built; `find_contours` closes its loops by repeating the first point, so nothing visibly changes.

2. `QTransform.fromScale(s, s).map(path)` multiplies each coordinate the same way the old per-point `QPointF * s` did, so the outline lands on the same device pixels at every zoom.

3. `QPen(color, width, style)` yields the same cap and join defaults as `QPen(color)` followed by `setWidth`, and the orientation-arrow and negative-vertex pens were already solid, so passing no style there is not a change.

Old and new renderers were compared pixel for pixel across mask shapes, scales from 0.37 to 7, selection, labels, and solid and dashed pens: identical. No changelog fragment since nothing is user-facing.

https://claude.ai/code/session_01Pn6a8wABPxAiM6E45bSVW5

### DIFF
--- a/labelme/_widgets/_shape_render.py
+++ b/labelme/_widgets/_shape_render.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import dataclasses
 import typing
-from typing import ClassVar
 from typing import Final
 from typing import Literal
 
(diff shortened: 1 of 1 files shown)


wkentaro:
Verdict: recommend-merge.

Code/spec review found no behavioral defects. A 288-case antialiased pixel comparison between base 48463e6b and head 5c5e002c was identical across mask topologies (holes, disconnected regions, and border-touching regions), scales from 0.37 to 7, selection, labels, solid and dashed pens, the orientation arrow, and negative vertices. The focused renderer suite passed all 35 tests; all required CI checks are green.

Computer Use on the isolated PR-head desktop build loaded examples/primitives/primitives.json and confirmed the octagon mask fill and contour stayed aligned at Fit to Window (150.7%) and 37% zoom, including selected rendering.

Non-blocking standards note: commit 5c5e002c contains a Co-Authored-By trailer, contrary to the repository instruction to omit co-author trailers. The source diff itself has no standards findings.

Issue #2151, opened 2026-06-11
Restructure Canvas: drawing-state and painting decomposition


wkentaro:
## What to build

Follow-up to the interaction-internals work: extract the drawing-mode state machine and the painting pipeline from the Canvas monolith into separate, testable units.

## Acceptance criteria

- [ ] Every creation mode behaves identically: polygon, rectangle, circle, line, point, AI modes
- [ ] Undo behaves identically, including mid-draw undo
- [ ] Test suite passes; manual parity check against main

## Blocked by

- #2150 (Canvas interaction internals)

PR #2610, opened 2026-09-05
perf(canvas): reuse AI previews across repaints

wkentaro:
Repainting an unchanged AI point preview repeatedly runs inference while the painter is active. Painting now queues a deferred refresh, which memoizes the proposal inputs so an unchanged repaint performs no inference. The previous preview stays visible until its replacement arrives, avoiding blank frames during pointer motion.

Inference still runs synchronously on the GUI thread. Paint-driven refresh also handles showing a hidden canvas without caller-side scheduling hooks. Match highlights use one store; finalization computes a fresh proposal from committed prompt points. This addresses the repaint-related portion of #2151.

Validation: 187 canvas tests and 1,158 unit tests passed (one unit test skipped and one deselected); make lint passed. The retained click/move/repaint regression checks inference runs outside the active painter and the preview remains available during refresh. Additional coverage checks hidden Settings changes, Shape-list changes, brightness updates, and observable empty/error/cancel/reset outcomes. Desktop QA with cached SAM2 verified preview generation, brightness adjustment during an active draft, and Escape cancellation. All 21 GitHub checks pass on `a8a5d24d`.

### DIFF
--- a/labelme/_widgets/canvas.py
+++ b/labelme/_widgets/canvas.py
@@ -204,6 +204,8 @@ class Canvas(QtWidgets.QWidget):
     _ai_assist_session: _automation.AiAssistSession
     _ai_suppress_existing_shape_matches: bool
     _ai_existing_shape_highlights: list[Shape]
+    _ai_points_preview: list[Shape]
+    _ai_points_preview_key: tuple[object, ...] | None
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:  # noqa: ANN401
         self._epsilon: float = kwargs.pop("epsilon", 10.0)
@@ -232,6 +234,9 @@ class Canvas(QtWidgets.QWidget):
         )
         super().__init__(*args, **kwargs)
 
+        self._ai_preview_timer = QtCore.QTimer(self)
+        self._ai_preview_timer.setSingleShot(True)
+        self._ai_preview_timer.timeout.connect(self._refresh_ai_points_preview)
         self._cursor = CursorRole.DEFAULT
         self.reset_state()
 
@@ -277,6 +282,7 @@ class Canvas(QtWidgets.QWidget):
 
     def set_allow_out_of_bounds_points(self, *, value: bool) -> None:
         self._allow_out_of_bounds_points = value
+        self.update()
 
     def pan_view(self, *, step: QPointF, constrain_to_center: bool = True) -> None:
         viewport = self._scroll_viewport()
@@ -1707,12 +1713,15 @@ class Canvas(QtWidgets.QWidget):
         render_shape(painter=painter, shape=shape, context=context)
 
     def _build_preview_shapes(self) -> list[Shape]:
+        if self.create_mode == "ai_points_to_shape":
+            # Defer inference until painting ends; unchanged inputs stop the loop.
+            self._ai_preview_timer.start(0)
         if self._current is None:
             return []
         if self.create_mode == "polygon":
             return [self._build_polygon_preview(current=self._current)]
         if self.create_mode == "ai_points_to_shape":
-            return self._build_ai_points_preview(current=self._current)
+            return self._ai_points_preview
         return []
 
     def _build_polygon_preview(self, *, current: _DraftShape) -> Shape:
@@ -1724,9 +1733,31 @@ class Canvas(QtWidgets.QWidget):
             preview = preview.add_point(self._line.points[1], autoclose=True)
         return _draft_to_shape(preview)
 
-    def _build_ai_points_preview(self, *, current: _DraftShape) -> list[Shape]:
+    def _refresh_ai_points_preview(self) -> None:
+        current = self._current
+        if current is None or self.create_mode != "ai_points_to_shape":
+            self._ai_points_preview = []
+            self._ai_points_preview_key = None
+            return
+        key = (
+            current,
+            self._line,
+            self._ai_assist_session.model_name,
+            self._ai_assist_session.output_format,
+            self._ai_assist_session.polygon_detail,
+            self._ai_suppress_existing_shape_matches,
+            self._allow_out_of_bounds_points,
+            self._pixmap_hash,
+            tuple(self.shapes),
+        )
+        if key == self._ai_points_preview_key:
+            return
+        self._ai_points_preview_key = key
         if not _ai_models.supports_point_prompts(model_name=self.get_ai_model_name()):
-            return []
+            self._ai_points_preview = []
+            self._set_ai_existing_shape_highlights(shapes=[])
+            self.update()
+            return
         preview = current.add_point(
             self._line.points[1],
             label=self._line.point_labels[1],
@@ -1738,16 +1769,18 @@ class Canvas(QtWidgets.QWidget):
                 point_labels=preview.point_labels,
             )
         except Exception as e:
-            # This runs inside paintEvent on every repaint, so a persistently
-            # failing model would report on every frame. Report once; a later
-            # success re-arms the report.
+            self._ai_points_preview = []
+            self._set_ai_existing_shape_highlights(shapes=[])
+            # Repeated failed prompts report once; a success re-arms the report.
             if not self._ai_inference_failed:
                 self._report_inference_failure(error=e)
-            self._set_ai_existing_shape_highlights(shapes=[])
-            return []
-        self._ai_inference_failed = False
-        self._set_ai_existing_shape_highlights(shapes=proposal.matching_existing_shapes)
-        return proposal.new_shapes
+        else:
+            self._ai_inference_failed = False
+            self._ai_points_preview = proposal.new_shapes
+            self._set_ai_existing_shape_highlights(
+                shapes=proposal.matching_existing_shapes
+            )
+        self.update()
 
     def transform_widget_point_to_image(self, point: QPointF, /) -> QPointF:
         origin = self._compute_image_origin_offset(area=None)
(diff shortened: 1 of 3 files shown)


labelme/_widgets/canvas.py
```diff
@@ -1724,30 +1742,46 @@ def _build_polygon_preview(self, *, current: _DraftShape) -> Shape:
             preview = preview.add_point(self._line.points[1], autoclose=True)
         return _draft_to_shape(preview)
 
-    def _build_ai_points_preview(self, *, current: _DraftShape) -> list[Shape]:
+    def _schedule_ai_points_preview(self) -> None:
```
wkentaro: Suggestion: pull-based instead of push-based. Let the paint path arm the timer and let the refresh memoize on its inputs, so no caller has to remember to schedule:

```python
# _build_preview_shapes
if self.create_mode == "ai_points_to_shape":
    # Inference runs once this paint returns and repaints on a change;
    # a repaint with unchanged inputs finds the key equal and stops.
    self._ai_preview_timer.start(0)
    return self._ai_points_preview

# _refresh_ai_points_preview
current = self._current
if current is None or self.create_mode != "ai_points_to_shape":
    self._ai_points_preview = []
    return
key = (
    current, self._line,
    self._ai_assist_session.model_name,
    self._ai_assist_session.output_format,
    self._ai_assist_session.polygon_detail,
    self._ai_suppress_existing_shape_matches,
    self._allow_out_of_bounds_points,
    self._pixmap_hash,
    tuple(self.shapes),
)
if key == self._ai_points_preview_key:
    return
self._ai_points_preview_key = key
...  # inference as here, then self.update()
```

Every state change that matters already calls `update()`, which is what main relied on; only `set_allow_out_of_bounds_points` needs one added. That removes all fourteen scheduling sites, the `isVisible()` guard (hidden widgets do not paint, and showing one paints), and the dual highlight source. The loop paint -> refresh -> update -> paint converges because the refresh only repaints when the key changed.

Verified: 185 canvas tests, 1156 unit tests, ruff and ty clean. The only test edits needed: `test_points_preview_hides_failed_and_empty_predictions` varies the prompt point between calls instead of calling schedule and refresh, and the empty/error outcome below calls `update()` and waits on the preview instead of the timer.

labelme/_widgets/canvas.py
```diff
@@ -1724,30 +1742,46 @@ def _build_polygon_preview(self, *, current: _DraftShape) -> Shape:
...
+        self._ai_preview_timer.stop()
+        if (
+            self.isVisible()
```
wkentaro: A canvas hidden mid-draw clears its preview here and nothing re-arms it on show, so the preview stays missing until the next input change. Not blocking, and it disappears with the pull-based version since showing the widget paints it.

labelme/_widgets/canvas.py
```diff
@@ -1691,7 +1704,12 @@ def _draw_ai_existing_match_layer(self, painter: QtGui.QPainter, /) -> None:
             rotation_highlight=None,
             line_style=Qt.PenStyle.DashLine,
         )
-        for shape in self._ai_existing_shape_highlights:
+        highlights = (
```
wkentaro: Two sources of truth for highlights now: the stored proposal while a preview exists, `_ai_existing_shape_highlights` otherwise. Having the refresh call `_set_ai_existing_shape_highlights` as main did keeps one source and lets `_ai_points_preview` be a plain `list[Shape]`.

labelme/_widgets/canvas.py
```diff
@@ -518,6 +527,7 @@ def _report_inference_failure(self, *, error: Exception) -> None:
 
     def backup_shapes(self) -> None:
         self.shape_backups.append([s.copy() for s in self.shapes])
+        self._schedule_ai_points_preview()
```
wkentaro: Using `backup_shapes` as a proxy for "shapes changed" is a heuristic: it fires before the mutation and misses shape changes that do not go through it. With the memo key including `tuple(self.shapes)`, the shapes input is covered without this call.

tests/unit/widgets/canvas_test.py
```diff
@@ -2233,3 +2240,156 @@ def test_end_move_in_place_copies_points(*, canvas: Canvas) -> None:
 
     assert np.array_equal(shape.points, clone.points)
     assert not np.shares_memory(shape.points, clone.points)
+
+
+@pytest.mark.gui
+def test_ai_preview_follows_input_without_inference_during_paint(
```
wkentaro, line 2251: This is the meaningful one: real clicks and moves, `grab()` repaints, and the assertion that inference never runs with the painter active. Keep it as is.

tests/unit/widgets/canvas_test.py
```diff
@@ -2233,3 +2240,156 @@ def test_end_move_in_place_copies_points(*, canvas: Canvas) -> None:
...
+    qtbot.keyClick(canvas, Qt.Key.Key_Escape)
+    QtCore.QCoreApplication.processEvents()
+    canvas.grab()
+    assert len(prompts) == 4
+    assert canvas._build_preview_shapes() == []
+
+
+@pytest.mark.gui
+@pytest.mark.parametrize(
+    "change", ["model", "format", "detail", "suppression", "bounds", "image"]
```
wkentaro: The `image` case exercises `load_pixmap` mid-draft with the old prompt points on the new image. The app never reaches that state because `reset_state` runs before a load, so I would drop it as speculative.

tests/unit/widgets/canvas_test.py
```diff
@@ -2233,3 +2240,156 @@ def test_end_move_in_place_copies_points(*, canvas: Canvas) -> None:
...
+    if outcome == "cancel":
+        qtbot.keyClick(canvas, Qt.Key.Key_Escape)
+    elif outcome == "image":
+        pixmap = QtGui.QPixmap(20, 20)
+        pixmap.fill(Qt.GlobalColor.white)
+        canvas.load_pixmap(pixmap=pixmap, clear_shapes=False)
+    else:
+        fail = outcome == "error"
+        proposal = AiAssistProposal(new_shapes=[], matching_existing_shapes=[])
+        qtbot.waitUntil(lambda: not canvas._ai_preview_timer.isActive())
```
wkentaro: Waiting on the private timer couples the test to the scheduling mechanism. `canvas.update()` followed by `qtbot.waitUntil(lambda: not canvas._build_preview_shapes())` asserts the visible outcome and survives either design.

wkentaro (commented):
Verdict: worth adding and no verified defect, but not minimal. Verified on `06b56263` in an isolated worktree: 185 canvas tests and ruff pass.

Worth adding: main runs inference inside every paintEvent and emits the failure signal from inside the painter; moving it to the event loop and reusing the stored proposal on repaint is a real fix.

Not minimal: fourteen `_schedule_ai_points_preview` sites are push-based invalidation with no guard. Every future `_current` or `_line` assignment needs one, and a missed one silently leaves a stale preview. Details and a verified pull-based alternative in the inline comments; its diff against main is +45/-11 in canvas.py versus about 100 changed lines here, and this PR's own tests pass on it with two small edits.

labelme/_widgets/canvas.py
wkentaro: Implemented in a8a5d24d: painting arms the single-shot timer, the refresh compares its input key before inference, and unchanged inputs produce no further repaint. Removed the scheduling helper and all caller-side scheduling sites. The previous preview remains visible during refresh. Validation: 187 canvas tests, 1,158 unit tests, and make lint passed.

labelme/_widgets/canvas.py
wkentaro: Fixed in a8a5d24d. Added a regression that hides the canvas, changes polygon detail, and shows it again. It timed out on 06b56263 and passes with paint-driven refresh.

labelme/_widgets/canvas.py
wkentaro: Implemented in a8a5d24d. The preview is a list of Shapes again; the refresh updates the existing match-highlight store. Rendering now reads that single highlight source.

labelme/_widgets/canvas.py
wkentaro: Removed the backup_shapes scheduling hook in a8a5d24d. The memo key includes the current Shape sequence, with a regression that adds a Shape and requests a repaint without going through backup_shapes.

tests/unit/widgets/canvas_test.py
wkentaro: Kept this test unchanged in a8a5d24d, including real clicks/moves, repeated grab() repaints, the inactive-painter assertion, and the pending-preview retention check.

tests/unit/widgets/canvas_test.py
wkentaro: Normal file navigation does reset first, but brightness/contrast calls load_pixmap(clear_shapes=False) while preserving an active draft through MainWindow._on_brightness_contrast_changed. I verified that desktop flow with cached SAM2. In a8a5d24d the refresh case is renamed brightness and keeps the image dimensions unchanged; the separate unavailable-preview case now exercises reset_state instead of loading an unrelated image mid-draft.

tests/unit/widgets/canvas_test.py
wkentaro: Changed in a8a5d24d to canvas.update() followed by waiting for an empty preview. The test no longer inspects whether the private timer is active.

wkentaro (commented):
Re-reviewed `a8a5d24d`. All seven threads are addressed as described, and I verified the changes rather than the replies.

Verified locally in an isolated worktree: 1158 unit tests pass, ruff check, ruff format, and ty are clean. All 21 GitHub checks are green.

The paint-armed timer plus memo key now replaces every explicit scheduling site, the highlight store is single-source again, and hidden-then-shown canvases refresh without a hook. Arming the timer before the `_current is None` check is a good improvement over my sketch: the cancel frame itself clears the stale preview.

I withdraw the objection to the image case. The brightness/contrast dialog does call `load_pixmap(clear_shapes=False)` with the draft alive, so the renamed `brightness` case covers a real app path.

Verdict: ready to merge.

PR #2611, opened 2026-09-05
fix(ai): keep the app open when text Shape generation fails

wkentaro:
Only the model call in the AI Text Prompt handler was guarded. Any error raised afterwards, in suppression, mask indexing, Shape construction, or Group allocation, escaped the slot and the global excepthook closed the app with a critical dialog.

MainWindow owned that numeric protocol: synthetic scores for existing Shapes, aligned box/mask indices, and conversion into new Shapes. Moving it behind one Qt-free proposal call puts the whole pipeline under the existing failure handler and gives it a direct test surface.

Text suppression intentionally keeps its existing semantics, which differ from AI Assist point matching. Missing-mask output retains its specific user warning; other inference and postprocessing failures use the existing failure status message.

Validation: 138 automation and text-workflow tests passed before the additional regressions; the final text-focused run passed 20 tests, including real NMS with mask/index alignment and preservation of existing Shapes on failures. `make lint` passed, and a fresh-process import confirmed the proposal module loads without Qt. The desktop text-model check was not completed because computer-use model selection returned an accessibility error; no desktop parity claim is made.

https://claude.ai/code/session_01T3geU2ZzSYoc5sjxpXXTDJ

### DIFF
--- a/changelog.d/2611.fixed.md
+++ b/changelog.d/2611.fixed.md
@@ -1,0 +1,1 @@
+AI Text Prompt reports errors during Shape generation instead of closing the app.
--- a/labelme/_app.py
+++ b/labelme/_app.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import enum
 import functools
-import json
 import math
 import os
 import platform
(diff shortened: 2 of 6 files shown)


wkentaro:
**Verdict: approve** (recorded as a comment; GitHub blocks approving one's own PR). Worth adding, correct, minimal, and the tests earn their place.

**Worth adding.** This is a real fix, not only a refactor. On `main`, only the model call sits inside the try block. Anything raised after it (NMS, mask polygonization, group-id assignment) escapes the slot, hits the global excepthook in `labelme/__main__.py`, and closes the app with a critical dialog. Moving the whole pipeline under one call puts it all inside the existing catch.

**Correct.** Verified against the branch head, not the description:

- The moved body is a line-for-line copy of the old window code. Ordering is preserved: mask check, existing-shape injection, NMS, `is_new` filtering, greedy suppression, group ids.

- The mask-unavailable branch still shows the warning dialog and is caught before the generic handler.

- The extracted module imports without loading any Qt module (checked in a fresh process).

- No other callers of the two removed re-exports remain outside the unit test file, which imports them from the private module directly.

- `make lint` passes; 10 unit and 10 e2e text-prompt tests pass locally; all CI checks are green.

**Minimal.** `_app.py` is net −63 lines. The only new abstraction is one exception class, needed to route the mask case to a different dialog.

**Tests.** The e2e test is the primary guard: narrowing the window's catch clause to a different exception type makes both parametrized cases fail with the injected error, so it detects exactly the regression this PR fixes. Asserting the annotated image's shape list is unchanged is stronger than the old empty-canvas check. The alignment unit test runs real osam NMS, real shape building, and real group-id assignment against a fake session, so it is an integrated test of the extracted function rather than a mock-heavy unit; the one-pixel mask difference catches an index-mapping mistake the e2e suite cannot see, and that mapping had no test before. The mask-required test covers a guard that had no test on `main`.

**Nits, non-blocking:**

- The changelog line says the fix avoids "changing the current Annotation". That was already true on `main` since shapes were never loaded on failure. "without closing the app" is the actual user-facing change.

- The `nullcontext` vs `pytest.raises` pattern with an assert inside the `with` is slightly clever; two plain tests would read faster.

Issue #2419, opened 2026-07-25
refactor(examples): stop hand-vendoring shape_to_mask / shapes_to_label into examples/utils.py


wkentaro:
> *This was generated by AI during triage.*

Noticed while processing [#2327](https://github.com/wkentaro/labelme/pull/2327). Filed as a standalone idea; it does not block that PR.

## Observation

`shape_to_mask` and `shapes_to_label` exist twice, in `labelme/_utils/shape.py` and in `examples/utils.py`, and the two copies are kept in sync by hand:

- `shape_to_mask` — **byte-identical** in both (43 lines each).
- `shapes_to_label` — identical except for type annotations (the package version uses the `ShapeDict` TypedDict, the examples version uses `dict[str, Any]`) and one blank line.

The cost is that a rasterization bug has to be found and fixed twice, in lockstep, or the copies silently diverge. #2327 is exactly that: the same 7-line out-of-bounds clip had to be applied to both files, and ADR 0004 names the examples copy explicitly, so missing it would have left `labelme2voc` / `labelme2coco` still wrapping pixels onto the opposite edge. Nothing enforces the pairing today, so the next such fix can land in one copy only.

## Why the duplication exists

Deliberate, not accidental: the five example scripts import it as a plain sibling module so they stay runnable standalone.

```
examples/tutorial/draw_json.py:11:import utils  # examples/utils.py, vendored alongside this script
examples/tutorial/export_json.py:14:import utils
examples/instance_segmentation/labelme2coco.py:24:import utils
examples/instance_segmentation/labelme2voc.py:12:import utils
examples/bbox_detection/labelme2voc.py:18:import utils
```

So the question is whether standalone-runnability is still worth a hand-synced copy, given the scripts already live in a repo where `labelme` is installable.

## Options

1. **Import from the package** — `examples/utils.py` re-exports from `labelme._utils.shape`. Removes the divergence entirely; costs the scripts their standalone property and reaches into a private module.
2. **Promote to public API** — expose these two under a supported path and have the examples import that. Cleanest, but makes them a compatibility surface that has to be maintained.
3. **Keep the copy, enforce it** — a test asserting the two function sources match, so a one-sided fix fails CI. Cheapest, preserves standalone scripts, but ossifies the annotation drift noted above (the test would need to normalize it).
4. **Leave as is** — accept the hand-sync.

Which trade-off is right is a maintainer call; recording the options rather than assuming one.

wkentaro:
[...]

PR #2617, opened 2026-09-06
refactor: simplify standalone dataset exporters

wkentaro:
Keep the standalone VOC and COCO converters easy to adapt while preserving optional output controls and grouped-instance behavior.

Circle export now picks its vertex count by doubling from 12 until the chord-to-arc deviation is within one pixel. The previous closed-form count floored the result and slightly exceeded the tolerance (about 1.01 to 1.09 px for radii of 50 px and up), so circles of that size gain vertices. For example, radius 2,000 changes from 99 to 192 vertices (+94%), and radius 10,000 changes from 222 to 384 (+73%).

Validation: VOC output on `tests/data/annotated` is byte-identical to `main`, and COCO output differs only in the creation timestamp. 36 exporter tests pass with pycocotools installed; CI now installs it before pytest on the non-Windows legs so the COCO tests execute there instead of skipping. The circle accuracy test reads the real CLI’s exported JSON and checks every edge, including the closing edge; it rejects both a semicircle mutation and the old floor-rounding defect. `make lint` passes with an existing unused-type-ignore warning on the optional pycocotools import.

Stack 2/8. Prerequisite #2616 is merged. Next: #2618.

https://claude.ai/code/session_016EEs8Vh1FEmZiiwJL36AfR

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
         shell: bash
         if: matrix.os != 'windows-latest'
         run: |
-          uv pip install cython && uv pip install pycocotools
+          uv pip install cython pycocotools
       # Pixel snapshots skip themselves off Linux (tests/e2e/conftest.py), so
       # one plain invocation runs the right set everywhere.
       - name: Test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,14 @@ jobs:
         with:
           packages: xvfb libegl1 libxkbcommon0 libxcb-cursor0 libgl1
           version: 1.0
+      # The COCO exporter tests skip themselves without pycocotools, so it has
+      # to be present before pytest runs. The pip install survives the inexact
+      # sync that uv run performs.
+      - name: Install optional example dependencies
+        shell: bash
+        if: matrix.os != 'windows-latest'
+        run: |
+          uv pip install cython && uv pip install pycocotools
       # Pixel snapshots skip themselves off Linux (tests/e2e/conftest.py), so
       # one plain invocation runs the right set everywhere.
       - name: Test
@@ -64,7 +72,6 @@ jobs:
           (cd examples/video_annotation && rm -rf data_dataset_voc && ./labelme2voc.py data_annotated data_dataset_voc --labels labels.txt && git checkout -- .)
           uv pip install lxml  # for bbox_detection/labelme2voc.py
           (cd examples/bbox_detection && rm -rf data_dataset_voc && ./labelme2voc.py data_annotated data_dataset_voc --labels labels.txt && git checkout -- .)
-          uv pip install cython && uv pip install pycocotools  # for instance_segmentation/labelme2coco.py
           (cd examples/instance_segmentation && rm -rf data_dataset_coco && ./labelme2coco.py data_annotated data_dataset_coco --labels labels.txt && git checkout -- .)
   artifact-install-smoke:
     runs-on: ${{ matrix.os }}
--- a/examples/instance_segmentation/labelme2coco.py
+++ b/examples/instance_segmentation/labelme2coco.py
@@ -14,14 +14,21 @@ from typing import Final
 import imgviz
 import numpy as np
 
+try:
+    import pycocotools.mask  # type: ignore
+except ImportError:
+    print("Please install pycocotools:\n\n    pip install pycocotools\n")
+    sys.exit(1)
+
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
 import utils  # noqa: E402  # examples/utils.py, vendored alongside this script
 
 
-def _polygon_vertex_count_for_circle(radius: float, /) -> int:
+def _count_polygon_vertices_for_circle(radius: float, /) -> int:
     CIRCLE_APPROXIMATION_TOLERANCE_PX: Final = 1.0
     CIRCLE_MIN_VERTEX_COUNT: Final = 12
 
+    # Doubling reduces the chord-to-arc deviation until it is within one pixel.
     vertex_count = CIRCLE_MIN_VERTEX_COUNT
     while radius * (1.0 - math.cos(math.pi / vertex_count)) > (
         CIRCLE_APPROXIMATION_TOLERANCE_PX
@@ -38,7 +45,7 @@ def _circle_to_polygon_segmentation(
     if radius == 0.0:
         raise ValueError("Degenerate circle: center and edge are the same point.")
 
-    vertex_count = _polygon_vertex_count_for_circle(radius)
+    vertex_count = _count_polygon_vertices_for_circle(radius)
     angles = np.linspace(0.0, 2.0 * math.pi, num=vertex_count, endpoint=False)
     vertices = np.column_stack(
         (cx + radius * np.cos(angles), cy + radius * np.sin(angles))
@@ -47,14 +54,6 @@ def _circle_to_polygon_segmentation(
 
 
 def main() -> None:
-    # Imported lazily so importing this module for its polygon geometry (as
-    # the unit tests do) never requires the optional pycocotools dependency.
-    try:
-        import pycocotools.mask  # type: ignore  # noqa: PLC0415
-    except ImportError:
-        print("Please install pycocotools:\n\n    pip install pycocotools\n")
-        sys.exit(1)
-
     parser = argparse.ArgumentParser(
         formatter_class=argparse.ArgumentDefaultsHelpFormatter
     )
--- a/examples/instance_segmentation/labelme2coco.py
+++ b/examples/instance_segmentation/labelme2coco.py
@@ -14,46 +14,47 @@ from typing import Final
 import imgviz
 import numpy as np
 
-try:
-    import pycocotools.mask  # type: ignore
-except ImportError:
-    print("Please install pycocotools:\n\n    pip install pycocotools\n")
-    sys.exit(1)
-
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
 import utils  # noqa: E402  # examples/utils.py, vendored alongside this script
+
+
+def _polygon_vertex_count_for_circle(radius: float, /) -> int:
+    CIRCLE_APPROXIMATION_TOLERANCE_PX: Final = 1.0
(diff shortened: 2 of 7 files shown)


wkentaro (commented):
Two-axis review (Standards vs. Spec) of `1e565b62...65ee30d6`. Both axes reviewed independently; details inline.

**Standards** — 3 hard violations (deferred `pycocotools` import with `noqa: PLC0415`, three noun-phrase function names, deleted `examples/utils.py` module docstring that carried the why) plus 3 judgement calls.

**Spec** — behavior preservation verified: all CLI flags, output layout, `__ignore__`/`_background_` handling, duplicate-output-dir exit, and `group_id` merging are intact; `shape_to_mask`/`shapes_to_label` untouched so ADR 0004 clipping and the #2419 hand-sync are unaffected; 29 exporter tests pass. The circle vertex-count change is a legitimate fix, not a regression: the old `int()` truncation could exceed its own 1px tolerance by ~0.016px, the doubling loop cannot. One finding: the commit body claims coverage of grouped instances and optional VOC outputs that the tests do not provide.

examples/instance_segmentation/labelme2coco.py
```diff
@@ -14,46 +14,47 @@
...
+        (cx + radius * np.cos(angles), cy + radius * np.sin(angles))
+    )
+    return vertices.ravel().tolist()
 
 
 def main() -> None:
+    # Imported lazily so importing this module for its polygon geometry (as
+    # the unit tests do) never requires the optional pycocotools dependency.
+    try:
+        import pycocotools.mask  # type: ignore  # noqa: PLC0415
```
wkentaro: **Standards (hard):** deferred import with `noqa: PLC0415`. The house convention allows a deferred import only to break a circular dependency; this one exists so tests can import the module without `pycocotools` installed, and it opts out of an enforced ruff rule rather than resolving it. Options: keep the top-level import and mark the COCO tests `importorskip("pycocotools")`, or move the circle helpers into `utils.py` where the tests can reach them without importing the script.

examples/instance_segmentation/labelme2coco.py
```diff
@@ -14,46 +14,47 @@
...
-    import pycocotools.mask  # type: ignore
-except ImportError:
-    print("Please install pycocotools:\n\n    pip install pycocotools\n")
-    sys.exit(1)
-
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
 import utils  # noqa: E402  # examples/utils.py, vendored alongside this script
 
 
+def _polygon_vertex_count_for_circle(radius: float, /) -> int:
```
wkentaro: **Standards (hard):** noun-phrase function name; house rule is a leading verb. Suggest `_count_polygon_vertices_for_circle`.

**Judgement call:** the old derivation comment (`r*(1-cos(pi/n)) <= tol` ⇒ `n >= pi/acos(1-tol/r)`) is gone and nothing explains why doubling from 12 lands within tolerance. Worth one line, since the vertex counts now differ from the previous release (r=100: 22 → 24).

examples/utils.py
```diff
@@ -38,30 +24,35 @@ class LabeledImage:
     shapes: list[dict[str, Any]]
 
 
-def load_label_file(filename: str, /) -> LabeledImage:
-    with open(filename, encoding="utf-8") as f:
-        data = json.load(f)
+def _image_bytes_from_record(record: dict[str, Any], /, *, json_dir: Path) -> bytes:
```
wkentaro: **Standards (hard):** noun-phrase function names here and at `_shape_from_record` below; house rule is a leading verb. Suggest `_read_image_bytes` / `_build_shape`.

examples/utils.py
```diff
@@ -1,18 +1,4 @@
 #!/usr/bin/env python
```
wkentaro, line 1: **Standards (hard):** the removed module docstring explained *why* this file re-implements `shape_to_mask` / `shapes_to_label` instead of importing them (different owners and lifecycles, stdlib+numpy+PIL only). Nothing in the diff replaces that rationale, and #2419 shows it is exactly the question a reader will ask. Please keep at least the two sentences on ownership and dependencies.

examples/instance_segmentation/labelme2voc.py
```diff
@@ -7,11 +7,79 @@
...
+    if not include_objects:
+        return
+    (output_dir / "SegmentationObject").mkdir()
+    if include_npy:
+        (output_dir / "SegmentationObjectNpy").mkdir()
+    if include_visualizations:
+        (output_dir / "SegmentationObjectVisualization").mkdir()
+
+
+def _save_label_layer(
```
wkentaro, line 40: **Judgement call (Data Clumps):** 8 keyword parameters, several of which (`output_dir`, the `include_*` flags) also travel together through `_create_output_directories` and `main()`. A small options value would carry them once. Not blocking.

tests/unit/examples/labelme2voc_test.py
```diff
@@ -0,0 +1,104 @@
...
+    buf = io.BytesIO()
+    img.save(buf, format="PNG")
+    record = {
+        "version": "5.0.0",
+        "flags": {},
+        "shapes": [
+            {
+                "label": "cat",
+                "points": [[1, 1], [6, 6]],
+                "group_id": None,
```
wkentaro, line 29: **Spec:** the commit body says the tests "cover ... grouped instances", but no test builds two shapes sharing a `group_id`. I verified manually that two same-`group_id` rectangles still merge to one instance, so the behavior is fine, but the claim is untested. A two-shape fixture here would make it true.

tests/unit/examples/labelme2voc_test.py
```diff
@@ -0,0 +1,104 @@
...
+    _run(input_dir=annotated_dir, output_dir=output_dir, labels_file=labels_file)
+
+    assert (output_dir / "SegmentationClass" / "0001.png").is_file()
+    assert (output_dir / "SegmentationClassNpy" / "0001.npy").is_file()
+    assert (output_dir / "SegmentationClassVisualization" / "0001.jpg").is_file()
+    assert (output_dir / "SegmentationObject" / "0001.png").is_file()
+    assert (output_dir / "SegmentationObjectVisualization" / "0001.jpg").is_file()
+
+
+def test_labelme2voc_noviz_omits_visualization_outputs_only(
```
wkentaro: **Spec:** "optional VOC output files" is only partly covered: `--noviz` is tested, `--noobject` and `--nonpy` are not.

tests/unit/examples/labelme2voc_test.py
```diff
@@ -0,0 +1,104 @@
...
+        "imageHeight": 8,
+        "imageWidth": 8,
+    }
+    (directory / "0001.json").write_text(json.dumps(record))
+
+
+def _run(
+    *extra_args: str, input_dir: Path, output_dir: Path, labels_file: Path
+) -> None:
+    script = _REPO_ROOT / "examples" / "semantic_segmentation" / "labelme2voc.py"
```
wkentaro: **Nit:** this runs the `semantic_segmentation` path, which is a symlink to the `instance_segmentation` script this PR rewrote. It works today, but pointing the test at the real file would keep it honest if the symlink is ever replaced by a copy.

wkentaro (commented):
**Verdict: recommend merge**, subject to remaining CI checks. No blocking correctness defects found at `d187c87b`.

Worth adding: the shared VOC writer removes real duplication, the circle change fixes the one-pixel accuracy defect, and the new tests provide useful regression coverage. The implementation is mostly minimal; rounding the previous closed-form vertex count upward would be a smaller circle fix and produce fewer vertices.

The tests follow the preference for integrated coverage: VOC invokes the real CLI across all eight output-flag combinations and checks actual grouped/ungrouped PNG and NPY mask pixels. Loader tests use real files. The small geometry suite catches the original rounding defect and is not excessive unit testing; one nonblocking assertion improvement is inline.

Validation: all 36 exporter tests passed with pycocotools installed, and `make lint` passed. Against the PR base, all 22 VOC reference-output files were byte-identical and all seven COCO output files matched after removing the creation timestamp. A separate real COCO CLI check passed circle serialization, grouped instances, and deviation measured from actual polygon edges.

Nonblocking PR-description correction: “up to roughly 40 percent” understates vertex growth. Radius 2,000 changes from 99 to 192 vertices (+94%); radius 10,000 changes from 222 to 384 (+73%). Please correct that claim and update the reported test count from 32 to 36.

tests/unit/examples/labelme2coco_test.py
```diff
@@ -0,0 +1,31 @@
...
+    coords = labelme2coco._circle_to_polygon_segmentation(
+        center=(3.0, -2.0), edge=(3.0 + radius, -2.0)
+    )
+    points = np.array(coords).reshape(-1, 2)
+    np.testing.assert_allclose(np.hypot(points[:, 0] - 3.0, points[:, 1] + 2.0), radius)
+    n_vertices = len(points)
+    assert n_vertices >= 12
+    # Sagitta of a chord spanning 2*pi/n radians: the maximum distance the
+    # polygon edge deviates from the true arc.
+    assert radius * (1.0 - math.cos(math.pi / n_vertices)) <= 1.0
```
wkentaro: Nonblocking: this calculates deviation from the vertex count, assuming the returned vertices are evenly spaced around a complete circle. I verified that changing the angle range from `2.0 * math.pi` to `math.pi` still passes all five radius cases, although the polygon covers only a semicircle.

Measure deviation using the actual consecutive vertices, including the closing edge. Prefer checking a circle read from the real CLI's exported JSON so the same test covers serialization too; strengthen or replace the existing check rather than adding a suite for each helper. The current test still has value: it catches the previous floor-rounding defect.

.github/workflows/test.yml
```diff
@@ -35,6 +35,14 @@ jobs:
...
           packages: xvfb libegl1 libxkbcommon0 libxcb-cursor0 libgl1
           version: 1.0
+      # The COCO exporter tests skip themselves without pycocotools, so it has
+      # to be present before pytest runs. The pip install survives the inexact
+      # sync that uv run performs.
+      - name: Install optional example dependencies
+        shell: bash
+        if: matrix.os != 'windows-latest'
+        run: |
+          uv pip install cython && uv pip install pycocotools
```
wkentaro: can't this be oneline without &&?

PR #2627, opened 2026-09-06
refactor: separate shape model and geometry modules

wkentaro:
Separate shape state and topology edits from hit testing and oriented-rectangle calculations. The package exports the existing interface used by the canvas and renderer.

Review 1 of 7. Start here. Next: #2628.

Validation: full test suite — 1,412 passed, 7 skipped; full lint passed. The wheel builds and includes all four package modules. Function and class syntax trees match the previous implementation apart from private helper names.

Focused test: `uv run pytest -q tests/unit/_shape_test.py`.

Manual check: Insert/remove polygon vertices, then resize and rotate an oriented rectangle; confirm its center and arrow stay aligned.

Review and merge order: #2627 → #2628 → #2629 → #2630 → #2631 → #2632 → #2619.


labelme/_shape.py
```diff
@@ -164,26 +182,37 @@ def nearest_rotation_point_index(
...
-    distances = np.linalg.norm(handles - point, axis=1)
-    return _nearest_index_within_epsilon(distances=distances, epsilon=image_epsilon)
+    distances = np.linalg.norm(_rotation_handles(shape.points) - point, axis=1)
+    return _closest_index_within(distances=distances, epsilon=image_epsilon)
+
+
+# ---------------------------------------------------------------------------
+# Oriented-rectangle geometry: rotation handles, center, and the orientation
+# arrow drawn through it.
+# ---------------------------------------------------------------------------
```
wkentaro: if you add this kind of section title, maybe time to split the module into multiple files?

wkentaro: Split the module into `model.py`, `hit_testing.py`, and `oriented_rectangle.py`, and removed the section headings. The package exports the existing interface, so callers keep their imports. Full lint and 1,412 tests pass; the dependent PRs were rebased and pushed with `gh stack` and pass 1,549 tests.

PR #2629, opened 2026-09-06
refactor: isolate mask painting and shape outlines

wkentaro:
Keep mask pixel picking, contour painting, label anchors, and outline bounds in focused rendering helpers.

Review 3 of 7. After #2628. Next: #2630.

Validation: full test suite — 1,428 passed, 7 skipped; Ruff and type checks passed.

Focused test: `uv run pytest -q tests/unit/widgets/_shape_render_test.py`.

Manual check: Open a mask annotation, toggle labels, and zoom; the contour, filled mask, and label anchor should stay aligned.

PR #2630, opened 2026-09-06
test: preserve shape handle group highlighting

wkentaro:
Add regression tests for the existing shape handle highlighting: hovering or dragging a vertex highlights that shape's vertex group on polygons and open line strips, and hovering a rotation handle highlights the rotation-handle group without recoloring ordinary vertices. No production change; the renderer stays as on main.

Review 4 of 7. After #2629. Next: #2631.

Validation: full suite — 1,431 passed, 7 skipped; Ruff and type checks passed.

Focused test: `uv run pytest -q tests/unit/widgets/_shape_render_test.py`.

### DIFF
--- a/tests/unit/widgets/_shape_render_test.py
+++ b/tests/unit/widgets/_shape_render_test.py
@@ -10,6 +10,7 @@ from labelme._shape import Shape
 from labelme._shape import ShapeType
 from labelme._widgets._shape_render import Palette
 from labelme._widgets._shape_render import ShapeRenderContext
+from labelme._widgets._shape_render import VertexHighlight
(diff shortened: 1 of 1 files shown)


wkentaro:
Reduced this PR to the regression tests at head `8587105f` to keep the current main rendering behavior unchanged. The paint-order change, its pixel test, and the release note were dropped; the renderer is identical to main. Restacked the seven dependent PRs, none of which touch the renderer anymore. Full suite at this head: 1,431 passed, 7 skipped. Stack tip: 1,563 passed, 7 skipped, full lint passed.

PR #2640, opened 2026-09-07
fix: preserve labels until a suggestion is selected

wkentaro:
Opening a shape whose label differs only in case from a listed suggestion replaced it with the list's spelling (`cat` became `Cat`), and tabbing from the text field into the suggestion list overwrote custom text with the first row. Both came from applying the list's current item to the text field, which changes on open and on focus-in without the user choosing anything.

The dialog now applies a suggestion only on an explicit selection, click, or arrow key. The stored label is still highlighted and scrolled into view on open, with the list's signals blocked so the highlight cannot rewrite the label. Double-click still applies the item and accepts.

Before, on main at `55a360e6`, with this PR's regression tests:

```text
FAILED test_tabbing_into_choices_does_not_replace_custom_text
FAILED test_popup_preserves_label_case  (assert 'Cat' == 'cat')
```

After:

```text
uv run pytest -q tests/unit/widgets/label_dialog tests/e2e/label_dialog_validation_test.py
100 passed
make lint
All checks passed
```

An offscreen 500-label popup check at main and at this head highlighted `Label499` and kept it visible in both runs; accepting `label499` returned `Label499` on main and `label499` here. Native desktop verification was not performed.

https://claude.ai/code/session_016sFqszUNzkhKTDCbnDeyRg

### DIFF
--- a/labelme/_widgets/label_dialog.py
+++ b/labelme/_widgets/label_dialog.py
@@ -166,8 +166,9 @@ class LabelDialog(QtWidgets.QDialog):
 
         # Connect signals
         self.edit.textChanged.connect(self._on_text_changed)
-        self.label_list.currentItemChanged.connect(self._on_label_selected)
-        self.label_list.itemDoubleClicked.connect(self._submit_item)
+        self.label_list.itemSelectionChanged.connect(self._select_current_label)
+        self.label_list.itemClicked.connect(self._on_label_selected)
+        self.label_list.itemDoubleClicked.connect(self._accept_clicked_label)
 
         # Populate initial labels
         for label in dict.fromkeys([*(labels or []), *self._label_history]):
@@ -216,31 +217,28 @@ class LabelDialog(QtWidgets.QDialog):
         )
 
     def eventFilter(self, watched: QtCore.QObject, event: QtCore.QEvent, /) -> bool:
-        if watched is self.edit and event.type() == QtCore.QEvent.Type.KeyPress:
+        if event.type() == QtCore.QEvent.Type.KeyPress and watched is self.edit:
             assert isinstance(event, QtGui.QKeyEvent)
-            step = {QtCore.Qt.Key.Key_Up: -1, QtCore.Qt.Key.Key_Down: 1}.get(
-                QtCore.Qt.Key(event.key())
-            )
-            if step is not None:
-                row = self.label_list.currentRow() + step
-                self.label_list.setCurrentRow(
-                    min(max(row, 0), self.label_list.count() - 1)
-                )
+            if QtCore.Qt.Key(event.key()) in (
+                QtCore.Qt.Key.Key_Up,
+                QtCore.Qt.Key.Key_Down,
+            ):
+                self.label_list.keyPressEvent(event)
                 return True
         return super().eventFilter(watched, event)
 
-    def _on_label_selected(
-        self,
-        current: QtWidgets.QListWidgetItem | None,
-        _previous: QtWidgets.QListWidgetItem | None,
-        /,
-    ) -> None:
-        if current is None:
-            return
-        self.edit.setText(current.text())
+    def _select_current_label(self) -> None:
+        # Tabbing into the list sets a current row without selecting a label.
+        # Only an explicit selection should replace the typed text.
+        item = self.label_list.currentItem()
+        if item is not None and item.isSelected():
+            self._on_label_selected(item)
 
-    def _submit_item(self, item: QtWidgets.QListWidgetItem, /) -> None:
-        self.label_list.setCurrentItem(item)
+    def _on_label_selected(self, item: QtWidgets.QListWidgetItem, /) -> None:
+        self.edit.setText(item.text())
+
+    def _accept_clicked_label(self, item: QtWidgets.QListWidgetItem, /) -> None:
+        self._on_label_selected(item)
         self._ok_button.click()
 
     def _clear_flag_checkboxes(self) -> None:
@@ -347,7 +345,9 @@ class LabelDialog(QtWidgets.QDialog):
         else:
             self._set_flag_checkboxes(flags=flags)
 
-        self.label_list.setCurrentRow(self._find_label_row(text))
+        # Highlight the suggestion without applying its spelling to the label.
+        with QtCore.QSignalBlocker(self.label_list):
+            self.label_list.setCurrentRow(self._find_label_row(text))
 
         self._fit_label_list_to_content()
         self._refresh_ok_button()
--- a/tests/unit/widgets/label_dialog/interaction_test.py
+++ b/tests/unit/widgets/label_dialog/interaction_test.py
@@ -813,26 +813,6 @@ def test_popup_highlights_matching_label_at_show(*, qtbot: QtBot) -> None:
     assert seen["cur"] == "dog"
 
 
-def test_popup_highlights_matching_label_case_insensitively(*, qtbot: QtBot) -> None:
-    seen: dict[str, object] = {}
-    dialog = _add_dialog(qtbot, dialog=LabelDialog(labels=["Cat", "Dog"]))
-    _run_popup(
-        dialog=dialog,
-        accept=True,
-        text="cat",
-        at_show=lambda d: seen.update(
-            cur=d.label_list.currentItem().text()
-            if d.label_list.currentItem()
-            else None
-        ),
-        flags=None,
-        group_id=None,
-        description=None,
-        locked=(),
-    )
-    assert seen["cur"] == "Cat"
-
-
 def test_popup_clears_stale_highlight_when_nothing_matches(*, qtbot: QtBot) -> None:
     seen: dict[str, object] = {}
     dialog = _add_dialog(qtbot, dialog=LabelDialog(labels=["cat", "dog"]))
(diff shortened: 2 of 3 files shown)


labelme/_widgets/label_dialog.py
```diff
@@ -347,11 +334,11 @@ def popup(
         else:
             self._set_flag_checkboxes(flags=flags)
 
-        self.label_list.setCurrentRow(self._find_label_row(text))
+        self.label_list.setCurrentRow(-1)
```
wkentaro: This removes the highlight and scroll-to of the stored label on open, which has been there since the original dialog and which #2620 preserved. The bug was the signal applying the list's spelling, not the highlight itself. Keeping `_find_label_row` and doing

```python
with QtCore.QSignalBlocker(self.label_list):
    self.label_list.setCurrentRow(self._find_label_row(text))
```

passes everything on this branch except `test_popup_leaves_suggestion_list_unselected` and the `current: None` assertion in the leading-whitespace test, which exist only to pin the new behavior. Suggest keeping the highlight and restoring those two assertions.

labelme/_widgets/label_dialog.py
```diff
@@ -347,11 +334,11 @@ def popup(
...
 
         self._fit_label_list_to_content()
         self._refresh_ok_button()
-        self.edit.setFocus(QtCore.Qt.FocusReason.PopupFocusReason)
+        self.edit.setFocus(QtCore.Qt.FocusReason.OtherFocusReason)
```
wkentaro: Why `OtherFocusReason`? Neither the PR nor the commit body says, and all 107 dialog tests pass with `PopupFocusReason` restored. Drop it or add the reason to the commit body.

labelme/_widgets/label_dialog.py
```diff
@@ -216,32 +217,30 @@ def _refresh_ok_button(self) -> None:
...
+        if item is not None and item.isSelected():
+            self._on_label_selected(item)
 
-    def _submit_item(self, item: QtWidgets.QListWidgetItem, /) -> None:
-        self.label_list.setCurrentItem(item)
-        self._ok_button.click()
+    def _on_label_selected(self, item: QtWidgets.QListWidgetItem, /) -> None:
+        self.edit.setText(item.text())
+
+    def _accept_clicked_label(self, item: QtWidgets.QListWidgetItem, /) -> None:
```
wkentaro, line 240: Nit: `self._ok_button.click()` already no-ops when the button is disabled, so this can be

```python
self._on_label_selected(item)
self._ok_button.click()
```

tests/unit/widgets/label_dialog/interaction_test.py
```diff
@@ -1020,3 +996,116 @@ def edit_then_inspect(d: LabelDialog) -> None:
...
+
+@pytest.mark.parametrize(
+    "modifier",
+    [
+        QtCore.Qt.KeyboardModifier.ControlModifier,
+        QtCore.Qt.KeyboardModifier.MetaModifier,
+        QtCore.Qt.KeyboardModifier.AltModifier,
+    ],
+)
+def test_list_shortcuts_preserve_typed_label(
```
wkentaro: This asserts that Qt's `keyboardSearch` ignores modifier keys. No dialog code runs on this path, so it cannot catch a dialog regression. Suggest dropping it and the unmatched-search test below.

tests/unit/widgets/label_dialog/interaction_test.py
```diff
@@ -1020,3 +996,116 @@ def edit_then_inspect(d: LabelDialog) -> None:
...
+    dialog = _show_dialog_with_labels(qtbot, labels=["cat", "dog"])
+    dialog.label_list.setFocus()
+    dialog.label_list.setCurrentRow(0)
+    dialog.edit.setText("custom")
+    qtbot.keyClick(dialog.label_list, QtCore.Qt.Key.Key_C, modifier)
+    assert dialog.edit.text() == "custom"
+
+
+@pytest.mark.parametrize("selected", [False, True])
+def test_unmatched_list_search_preserves_typed_label(
```
wkentaro: Same as above: this pins Qt's behavior when typeahead finds no match, not anything in the dialog.

tests/unit/widgets/label_dialog/interaction_test.py
```diff
@@ -1020,3 +996,116 @@ def edit_then_inspect(d: LabelDialog) -> None:
...
+        dialog.show()
+    dialog.edit.setFocus()
+    for _ in range(4):
+        qtbot.keyClick(dialog.focusWidget(), QtCore.Qt.Key.Key_Tab)
+    assert dialog.focusWidget() is dialog.label_list
+    assert dialog.label_list.currentRow() == 0
+    assert dialog.edit.text() == "custom"
+
+
+def test_arrow_key_in_focused_label_list_selects_label(*, qtbot: QtBot) -> None:
```
wkentaro: Subsumed by the `Key_Down` case of `test_list_keyboard_choice_is_accepted_immediately`, which additionally checks the result reaches the entry.

tests/unit/widgets/label_dialog/interaction_test.py
```diff
@@ -1020,3 +996,116 @@ def edit_then_inspect(d: LabelDialog) -> None:
...
+    dialog = _show_dialog_with_labels(qtbot, labels=["cat", "dog"])
+    dialog.edit.setText("custom")
+    dialog.label_list.setCurrentRow(0)
+    dialog.label_list.setFocus(QtCore.Qt.FocusReason.TabFocusReason)
+    qtbot.keyClick(dialog.label_list, QtCore.Qt.Key.Key_Down)
+    qtbot.waitUntil(lambda: dialog.edit.text() == "dog")
+
+
+def test_popup_preserves_exact_label_across_casefold_collision(*, qtbot: QtBot) -> None:
```
wkentaro: This passes on main unchanged, so it guards nothing. Drop it.

tests/unit/widgets/label_dialog/interaction_test.py
```diff
@@ -1020,3 +996,116 @@ def edit_then_inspect(d: LabelDialog) -> None:
...
+        group_id=None,
+        description=None,
+        locked=(),
+    )
+
+    assert entry is not None
+    assert entry.label == "ß"
+
+
+def test_popup_does_not_guess_between_casefold_collisions(*, qtbot: QtBot) -> None:
```
wkentaro: This is the real regression and it fails on main, but the `ß`/`ss` fixture hides what the user actually hit. `LabelDialog(labels=["Cat"])` with `popup(text="cat")` fails on main the same way (`'Cat' == 'cat'`) and reads at a glance. Suggest that fixture instead.

wkentaro (commented):
Both bugs reproduce on main and are fixed here; lint and the full unit suite pass on the branch. Two changes requested before merge, plus a test trim.

1. The popup no longer highlights and scrolls to the stored label. That drops long-standing behavior rather than fixing it, and #2620 kept it. Wrapping the existing lookup in a signal blocker keeps the highlight and passes every test except the two that assert the new no-highlight behavior (see inline).

2. The focus-reason change has no rationale in the PR or commit body, and all 107 dialog tests pass with it reverted.

Tests: the tabbing test, the keyboard-choice test, and the `SS` casefold test are the ones that guard dialog logic. The other four assert Qt behavior or duplicate those (inline).

labelme/_widgets/label_dialog.py
wkentaro: Restored the existing lookup and wrapped the initial selection in QSignalBlocker in b733656b. Restored the highlight assertions; an offscreen 500-label popup check confirms the stored label is visible while its original spelling is retained.

labelme/_widgets/label_dialog.py
wkentaro: Restored PopupFocusReason in b733656b. All 100 remaining dialog and validation tests pass.

labelme/_widgets/label_dialog.py
wkentaro: Applied in b733656b: reuse the label-selection handler, then click the OK button. The dialog tests pass.

tests/unit/widgets/label_dialog/interaction_test.py
wkentaro: Removed the modifier-key and unmatched-search tests in b733656b.

tests/unit/widgets/label_dialog/interaction_test.py
wkentaro: Removed in b733656b; retained the tests covering explicit selection and label preservation.

tests/unit/widgets/label_dialog/interaction_test.py
wkentaro: Removed the duplicate arrow-key test in b733656b. The keyboard-choice test still covers Down and the accepted entry.

tests/unit/widgets/label_dialog/interaction_test.py
wkentaro: Removed in b733656b. This case also passed in the earlier baseline run.

tests/unit/widgets/label_dialog/interaction_test.py
wkentaro: Changed the fixture to Cat/cat in b733656b. The regression now checks both that Cat stays highlighted and that the accepted entry preserves cat.

wkentaro:
Verified at `12188129fa5128e692808b5e3ecd8ef9abe849a8` using Computer Use against the native macOS app (Python 3.12, PySide6 6.11.1), with disposable annotations and isolated window settings. The tested behavior matches the intended change.

- Opening `cat` preserves its lowercase spelling while highlighting the predefined `Cat` suggestion; accepting keeps `cat`.
- Tabbing into suggestions preserves `custom`; pressing Enter accepts and saves `custom` unchanged.
- Explicit arrow navigation applies suggestions from both the text field and the focused list. Down selects `Dog`, and Enter accepts it immediately.
- Typing `p` in the focused list selects `person`; Enter accepts it immediately.
- With 500 numbered labels, opening `label499` scrolls `Label499` into view and highlights it while retaining lowercase `label499` in the text field.
- Saved JSON was checked, and a clean restart retained the saved labels. Test app exits were clean.

Supplemental verification: `QT_QPA_PLATFORM=offscreen .venv/bin/python -m pytest -q tests/unit/widgets/label_dialog tests/e2e/label_dialog_validation_test.py` — **100 passed in 4.53s**.

Coverage gap: native suggestion click/double-click behavior remains unverified because Computer Use coordinate clicks returned `AXError.notImplemented`. This is an automation limitation, not a verified application defect. This was focused dialog verification, not a full application regression sweep.

PR #2667, opened 2026-09-09
fix: preserve saved shape metadata when undoing later creation

### LINKED ISSUE #2638: fix: preserve saved shape metadata when undoing later creation
Undoing a newly created polygon can also revert an earlier saved label/flag edit on an existing shape. This was observed through the native macOS desktop app and confirmed in saved annotation JSON on main at `2d333245b30c15bafef9b1c9cc03cb5a97ad594b`, as well as the candidate for #2620 at `eb17cfc648f6e65f309d6c32833dc700654a9bce`. It predates #2620.

Steps to reproduce:

1. Create a rectangle with label `dog` and save.

2. Edit the rectangle label to `person` (or another label with different flags) and save.

3. Create and save a polygon.

4. Undo polygon creation and save again.

Expected: only the polygon disappears; the rectangle retains the most recently saved label and flags.

Observed on the main baseline: before Undo, the JSON contains a `person` rectangle and a `person` polygon with empty flags. After Undo, only the rectangle remains, but its label is `dog` and its previous `occluded`/`furry` flags are restored. The candidate run independently showed the same pattern with `cat` reverting to `dog`; rectangle geometry was unchanged. Restart preserved the reverted metadata.

The label-edit and undo paths are unchanged between these two refs. Investigate how the snapshot history records the state after label/flag edits before a subsequent creation and Undo. A regression check should exercise edit → save → create → undo, including flags and unaffected geometry.

Environment: macOS, PySide6 / Qt 6.11.1. Windows/Linux were not exercised.

L4XB:
Fixes #2638

## What was wrong

`_edit_label` called `canvas.backup_shapes()` *before* applying the edit, while every other edit path (`set_last_label`, `end_move`, `_apply_in_place_move`, visibility toggling, `load_shapes`) snapshots the *resulting* state. The undo history is one snapshot per state with the last one being the current state (`restore_last_shape`: `[..., A, B] -> [...] -> [..., A]`), so after a label edit the "current" snapshot was actually the pre-edit state:

1. create rectangle `dog`, save → history `[…, S(rect dog)]`

2. edit label to `person`, save → history `[…, S(rect dog), S(rect dog)]` (snapshot taken before the mutation), canvas shows `person`

3. create polygon, save → history `[…, S(rect dog), S(rect dog), S(rect person + polygon)]`

4. undo → pops the current state and restores `S(rect dog)`: the polygon disappears **and** the rectangle goes back to `dog` with its old flags.

An undo directly after the edit was not affected: the stale snapshot is a copy of the pre-edit shapes, so restoring the entry below it reverted the edit correctly by accident.

## Fix

Take the snapshot after the edit, like the other paths. The pre-edit state is still in the history from whatever change produced it, so undo right after an edit still reverts exactly the edit.

The same bug class exists on the shape-creation path, where group id and description are assigned after the snapshot. That is tracked separately in #2676.

## Tests

`tests/e2e/undo_after_label_edit_test.py::test_undo_creation_keeps_earlier_label_edit`: edit label (with a `label_flags` pattern so flags change too) → create polygon → undo. Asserts only the polygon is gone, the first shape keeps `person` and its flags, and the other shapes' geometry is untouched. Saving is left out because it does not touch the undo history. Fails on `main`, passes with the fix.

Also ran `tests/unit` and the `shape_editing`, `last_label_and_restore`, `label_dialog_flags`, `action_availability`, `auto_save`, `label_list_canvas_selection` e2e suites (`QT_QPA_PLATFORM=offscreen`), plus `ruff check`, `ruff format --check`, and `ty check`.


L4XB:

L4XB:

wkentaro:
thanks!

PR #2669, opened 2026-09-10
refactor: inline end_move helpers and simplify shape-copy application

wkentaro:
Canvas.end_move delegated to two tiny private helpers, _apply_copy_move and _apply_in_place_move, that existed only to hold one loop each; folding them back into end_move removes a layer of indirection without losing readability, and the copy branch now reads as an extend-then-reassign instead of a manual enumerate loop, which is closer to what it is actually doing.

The two leading asserts collapsed into one since a non-empty selection with matching list lengths already guarantees the preview list is non-empty, and the trailing backup_shapes/update calls were reordered to match the bookkeeping-before-repaint order used elsewhere in this file.

Validated with `uv run pytest tests -x -q` (QT_QPA_PLATFORM=offscreen) and `make lint`, both passing unchanged.

### DIFF
--- a/labelme/_widgets/canvas.py
+++ b/labelme/_widgets/canvas.py
@@ -1362,26 +1362,18 @@ class Canvas(QtWidgets.QWidget):
         self._is_moving_shape = False
 
     def end_move(self, *, copy: bool) -> bool:
-        assert self.selected_shapes and self._selected_shapes_copy
-        assert len(self._selected_shapes_copy) == len(self.selected_shapes)
+        assert len(self.selected_shapes) == len(self._selected_shapes_copy) > 0
         if copy:
-            self._apply_copy_move()
+            self.shapes.extend(self._selected_shapes_copy)
+            self.selected_shapes[:] = self._selected_shapes_copy
         else:
-            self._apply_in_place_move()
+            for shape, preview in zip(self.selected_shapes, self._selected_shapes_copy):
+                shape.points = preview.points.copy()
         self._selected_shapes_copy.clear()
         self._drag_anchor = None
-        self.update()
         self.backup_shapes()
+        self.update()
         return True
-
-    def _apply_copy_move(self) -> None:
-        for i, clone in enumerate(self._selected_shapes_copy):
-            self.shapes.append(clone)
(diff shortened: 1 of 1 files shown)


wkentaro:
Verdict: **recommend-merge** at `5bb75e8dbf8977adc6025cea17d802076a7fe6fc`.

Worth adding as a small clean-room rewrite: removing two private, single-caller helpers puts the complete copy/move operation in one place. The copy branch uses ordinary list operations, removes indirection, and introduces no abstraction or dependency. Unchanged application behavior is expected here.

**Standards:** No verified defects. The change is confined to one function and removal of its helpers; an internal refactor does not need a user-facing changelog fragment.

**Spec / correctness:** No verified defects against the stated rewrite intent. Traced both callers and the drag-preview lifecycle. Copy preserves preview-object ordering and selection-list identity; move preserves original shape identity and independent point arrays. The combined assertion enforces the same invariants. Backup only copies shapes, so moving it before the queued repaint preserves normal behavior.

**Validation:** On this head, 319 focused canvas/interaction/app tests passed with 1 skipped; all 7 shape-editing end-to-end tests passed, including single- and multi-shape Copy Here, persistence, and copy independence. GitHub's reported checks are all green. Clean-room provenance itself was not audited.

Standards: 0 verified findings. Spec: 0 verified findings. Ready for maintainer review.

wkentaro:
Updated review: **recommend-merge** at `2b089115fc3fab7a5db617348d4b28f7e8ce66a2`, superseding my earlier-head review.

The new `len(selection) == len(preview) > 0` assertion states both invariants in one readable line and removes the unnecessary explanation. The helper inlining remains meaningful and minimal; unchanged copy/move behavior is expected for this clean-room rewrite.

**Standards:** 0 verified findings. **Spec:** 0 verified findings. No further concrete code simplifications suggested.

Validation on this head: 326 focused and shape-editing end-to-end tests passed, 1 skipped; changed-file Ruff lint/format checks passed. Also checked the actual assertion expression against the original invariant for 25 empty, matching, and mismatched list-length combinations. GitHub's reported CI checks are green.

PR #2677, opened 2026-09-14
fix: keep a new shape's group id and description across a later undo

### LINKED ISSUE #2676: fix: undo history snapshots a new shape before its group id and description are set
Same bug class as #2638, on the shape-creation path instead of the label-edit path. #2667 fixes the label-edit path only.

**Where**

In `_on_new_shape`, `canvas.set_last_label(...)` replaces the top undo snapshot with a copy of the shapes, and only afterwards does the loop assign `shape.group_id` and `shape.description` from the dialog entry. The snapshot that becomes the "current" state therefore has the new shape with `group_id=None` and `description=""`.

**Steps to reproduce**

1. Create a polygon and give it a group id or description in the label dialog. Save.

2. Create a second polygon. Save.

3. Undo.

Expected: only the second polygon disappears.

Observed: the first polygon also loses its group id and description, because the restored snapshot predates their assignment.

**Fix direction**

Snapshot after the group id and description are applied, matching the rule established in #2667: the last entry in the undo history must equal the current state. Either move the group id and description assignment into `set_last_label`, or re-take the snapshot at the end of `_on_new_shape`. A regression test should exercise create-with-group-id → save → create → undo and assert the first shape keeps its metadata.

wkentaro:
Fixes #2676

## What was wrong

Same bug class as #2638, on the shape-creation path. When a drawn shape was committed, `set_last_label` replaced the draft undo snapshot right after assigning the label and flags, and only afterwards did `_on_new_shape` assign the group id and description from the dialog. The "current" undo snapshot therefore had the new shape with no group id and no description, so undoing the next change (for example another new shape) also wiped them from the earlier shape.

## Fix

`set_last_label` no longer touches the undo history. `_on_new_shape` replaces the draft snapshot once, after every dialog field is applied, so the last history entry equals the current state as on the other edit paths.

## Tests

`tests/e2e/undo_after_new_shape_test.py`: draw a polygon and give it group id `7` and description `first` in the dialog → draw another polygon → undo. Asserts only the second polygon is gone and the first keeps its group id and description. Fails on `main`, passes with the fix.

Also ran `tests/unit` and the `undo_after_label_edit`, `shape_editing`, `last_label_and_restore`, `label_dialog_flags`, `action_availability`, `auto_save`, `annotation`, `ai_text_to_annotation` e2e suites (`QT_QPA_PLATFORM=offscreen`), plus `make lint` and `ty check`.


wkentaro:
Agent review (standards + spec axes), verdict: recommend-merge.

Verified against the code at 79f9652, not the description: `set_last_label` has exactly one caller, so dropping its pop/backup is safe. Undo-stack depth per creation is unchanged (one pop + one push, just moved after the group id and description are assigned). The dialog-cancel branch still pops exactly the draft snapshot from finalize. The AI Assist batch path calls `_on_new_shape` once for the whole batch, so the single snapshot covers every proposed shape.

The regression test fails on `main` and passes here. Lint, `ty check`, and the undo, label-edit, shape-editing, and last-label e2e suites pass locally; all CI checks are green.

No findings. One pre-existing judgement call left as is: `_on_new_shape` reaches into the canvas backup deque directly, but the cancel branch a few lines up already does the same, so no new abstraction was added.

wkentaro:
Second agent review (standards + spec axes) at e2e8ba24, verdict: recommend-merge.

Rebased onto main to resolve the conflict from the private-submodule rename in #2678; the diff is otherwise the same as before except one comment reworded to state why the snapshot is retaken instead of restating the two calls beneath it.

Verified against the code, not the description: the regression test fails with the fix reverted and passes with it. Unit tests plus the undo, label-edit, shape-editing, last-label, dialog-flags, action-availability, auto-save, annotation, and AI text-to-annotation e2e suites pass locally (1273 passed). Lint and ty check pass.

Spec axis: no missing or wrong requirements, no scope creep. The issue's repro says "Save" between steps and the test skips that; saving never touches the undo stack, so the omission does not reduce coverage.

Standards axis: no remaining findings. The direct pop on the canvas backup deque mirrors the pre-existing dialog-cancel branch a few lines up, so no new abstraction was added.

Issue #783, opened 2020-10-06
[BUG] Dragged Polygon disappears in "Polygon Labels" dock


gladcolor:
**Describe the bug**
When a shape was dragged to the first row, it will disappear. 

**To Reproduce**
1. Draw several polygons.
2. In the  "Polygon Labels" dock, drag a polygon to the first row.

wkentaro:
> *This was generated by AI during triage.*

Moving to **needs-triage** (same root cause as #1225, which now appears resolved). Dragging a row in the Polygon Labels dock used to overwrite/delete the dragged shape when it landed on another item (e.g. the first row). The list now uses a custom model whose `_ItemModel.dropMimeData` (`labelme/widgets/label_list_widget.py`) detects a drop *onto* an item and rewrites it to insert *after* that item instead of overwriting it, so the dragged label is preserved rather than disappearing. Recommend closing unless a specific drop target still drops a shape.

PR #2683, opened 2026-09-14
fix: keep the dragged shape when reordering the label list

### LINKED ISSUE #2681: Label List drag and drop replaces the moved Shape with a copy
## Summary

Dragging a row in the Label List swaps the moved item's `Shape` for a deserialized copy. When the reorder handler runs, it passes that copy to the canvas. So the canvas ends up holding a `Shape` that isn't the original object, and code that finds shapes by identity misses it.

## Evidence

I ran a probe with Qt's offscreen display on the #2679 branch head. It added three items to a `LabelListWidget`, then made the model calls a drop makes: `mimeData([index(0, 0)])`, then `dropMimeData(..., MoveAction, row=3, ...)`, then `removeRows(0, 1)`.

- The rows reorder correctly: `['b', 'c', 'a']`.
- The two rows that didn't move keep their original `Shape`s (`is` is True).
- The moved row now holds a different `Shape` with the same field values. Both `is` and `==` against the original are False.

Cause: `QStandardItemModel` stores the dragged item's data in the MIME payload through `QDataStream` and rebuilds it on drop. The new item comes from the prototype's `clone()`, and its `UserRole` data is read back from the stream, so a new Python object is built instead of reusing the original.

## Impact

- `_on_label_order_changed` builds the new order from `item.shape()`. After a real drag, the moved shape is the copy. On `main`, that list goes to `canvas.load_shapes`. On #2679, it goes through `_commit_shapes`.
- Anything that still holds the original object (the canvas selection that the reorder handler saves and restores, `find_item_by_shape`, and undo history) refers to a shape that is no longer in the list.

I haven't confirmed the selection or undo effect by dragging in the full app. I only ran the model-level probe above.

## Why tests miss it

`test_reorder_label_list_keeps_selection_and_enables_undo` in `tests/e2e/shape_editing_test.py` moves the row with `takeRow`/`appendRow`. That keeps the original item, so the MIME round-trip and `clone()` never run.

## Related

- The drop path also calls `clone()` on the empty item prototype, which reads `shape()` while it is `None`. Because of that, `shape()` can't simply return a non-optional `Shape`.
- This is different from #783 (a dragged row disappears) and #1191 (reordering doesn't work on Windows).

wkentaro:
Fixes #2681

Dragging a Label List row replaced the moved item's `Shape` with a deserialized copy. `QListView::dropEvent` tries `model()->moveRow(...)` first, and `QStandardItemModel` doesn't implement `moveRows`, so every drag used the MIME round-trip instead, which rebuilds the dropped rows through `clone()`.

## Changes

- `LabelListWidget.dropEvent` moves the selected rows itself with `takeRow` + `insertRow`, accepts the event, then calls the base handler. Qt's `dropOn` returns early for an accepted event and `startDrag` skips the source-row removal, so the same `LabelListWidgetItem` and `Shape` objects stay in the list and `item_dropped` fires once per drop.

- Dropping onto a row still inserts after it (#783). The `dropMimeData` override that did this is unreachable now and is removed with its tests.

- New unit test `test_drop_moves_items_and_emits_item_dropped_once` feeds synthetic `QDropEvent`s to the view: multi-row drop to the end, drop above a row, drop onto a row, and drop into the selection. Every shape is the same object and the signal fires once.

## Why not `moveRows` on the model

An earlier revision implemented `_ItemModel.moveRows` with `takeRow` + `insertRow`. Qt's drop loop computes the next insertion row from the persistent index of the row it just moved, and `takeRow` invalidates that index, so a multi-row drag left every row after the first in place. Verified with a real X11 drag under Xvfb: dragging rows `a, b` below `e` gave `b c d e a` instead of `c d e a b`.

## Verification

Real drags under Xvfb (Qt 6.11.1, `xdotool`), checking row order, `shape() is original`, and signal count:

- `main`: order correct, moved shapes are copies.

- this branch: rows `a, b` to the end → `c d e a b`, all shapes identical, one signal. `e` onto `a` → `a e b c d`. `e` to the top edge of `a` → `e a b c d`.

https://claude.ai/code/session_01JYgyidLKcc2A2hgw7ySFL8


wkentaro:
QA on `302d96bcbc7a341bc9ac4a883543b37040b76358`: no verified defect found in the tested paths. Native drag coverage is partial; this is not a complete cross-platform sign-off.

Environment: local macOS 26.6.2, Python 3.12.14, PySide6/Qt 6.11.1. Ran the PR's `MainWindow` from an isolated worktree with a disposable copy of `tests/data/annotated/2011_000003.json`, isolated settings, and read-only logging of shape identity, order, selection, and drop-signal count. Desktop interactions used Computer Use.

**Computer Use results**

- Dragged `amber_kite` from the first row into blank space below the last row: order became `blue_block, green_hexagon, red_triangle, purple_diamond, amber_kite`; canvas and list agreed, all five original Shape identities survived, and exactly one drop signal fired.

- Clicked Undo once: original five-row order returned and Undo became disabled.

- Repeated the end move, then dragged the last `amber_kite` onto the first `blue_block` row: it inserted after `blue_block`, with five shapes remaining and one additional drop signal.

- Selecting a row after reordering highlighted the corresponding canvas shape. Reordering itself clears selection; inspection of the base revision confirms the shared commit path already did this, so this is not a new regression finding.

**Automated results**

`python -m pytest -q tests/unit/widgets/_label_list_widget_test.py tests/e2e/shape_editing_test.py tests/e2e/label_list_canvas_selection_test.py` — **35 passed**. This includes synthetic multi-row, above-row, on-row, into-selection, identity, and undo checks. Configured Ruff formatting/lint, Gruff, and ty checks passed for the changed Python files.

**Coverage limits**

Successful native drags above used Qt's native macOS style. I also launched with the application's normal Fusion style, but Computer Use did not reliably complete the attempted drags. Native multi-row, top-edge, and no-op drops therefore remain unverified; the corresponding synthetic checks are not substitutes for native drag coverage. Save/reopen and Windows/Linux behavior were not manually tested in this session.

Standards review: no verified defects. Spec review against #2681: no verified defects; the native coverage limits above remain open.

PR #2692, opened 2026-09-15
feat: select neighboring shapes with Ctrl/Cmd+Arrow

kimchioverfit:
Selecting small annotations by mouse or finding them in list order interrupts keyboard editing. Ctrl+Arrow (Command+Arrow on macOS) selects a visible neighbor in the requested direction and keeps the Polygon Labels selection synchronized.

The shortcut follows [tldraw’s directional selection convention](https://tldraw.dev/sdk-features/selection#adjacent-selection). Tab/Shift+Tab in Figma and PowerPoint traverses objects sequentially, which is a different operation. The bindings can be changed or disabled in configuration.

Selection compares bounding-box centers within a 90° directional cone and favors candidates aligned with the direction of travel. Shapes with identical centers cannot be cycled through with these commands. With no selection, navigation starts near the image’s top-left corner; with no eligible neighbor, selection stays unchanged.

Validation on macOS: `uv run --no-sync pytest -q tests -n 4` — 1,518 passed, 7 skipped; `make lint` passed; `make check_translate` passed for all 20 languages. The focused GUI tests exercise successful navigation in all four directions, sidebar synchronization, hidden shapes, no-candidate behavior, and initial selection.


wkentaro:
the idea sounds solid and interesting.
i think the only concern i have is how we can decide the hotkey. like how does the other app do like figma/tldr/ppt

wkentaro:
macOS:

- excalidraw: cmd+arrow (do nothing)
- tldir: cmd+arrow (switch highlight to the neighbor)
- ppt
- figma

wkentaro:
[tldraw uses Ctrl/Cmd+Arrow for spatial neighbor selection](https://tldraw.dev/sdk-features/selection#adjacent-selection), so the proposed binding has a direct precedent.

[Figma](https://help.figma.com/hc/en-us/articles/360040449873-Select-layers-and-objects) and [PowerPoint](https://support.microsoft.com/en-us/accessibility/powerpoint/use-keyboard-shortcuts-to-create-powerpoint-presentations) use Tab/Shift+Tab for sequential selection, which is a different operation.

wkentaro (commented):
Verdict: recommend-merge (head 97ceeaf).

Meaningful: yes. Keyboard-only neighbor selection is a real workflow gap for dense or tiny annotations, and Ctrl/Cmd+Arrow matches tldraw's convention, which was the open question in the thread.

Correct: verified against the head, not the description. The 90-degree cone plus sideways penalty picks the expected neighbor on the test fixture in all four directions, hidden shapes are skipped, no-candidate leaves the selection alone, and the label list stays in sync. Ctrl+Arrow never reaches the canvas's own arrow-key nudge because the window-level shortcut consumes the event first, and QLineEdit's shortcut override keeps Cmd+Left/Right working in the file-search box, so there is no text-editing regression. Actions disable with the visibility actions when no shapes exist.

Minimal: acceptable. Four actions and four config keys are the floor for four configurable shortcuts. The only duplication is a second direction-to-vector table next to the existing arrow-key one, which is a nit, not a blocker. Tests are e2e through real key events, which is the right level. The Edit-menu enablement test overlaps the existing action-availability test and could have been folded in, but it is cheap.

Ran locally on macOS: the new test module plus the neighboring e2e modules (27 passed) and `make lint` (clean).

Issue #2696, opened 2026-09-16
feat: manage AI models and background downloads in Settings


wkentaro:
## Problem

AI Assist and AI Text Prompt list models as selectable before their files are downloaded. First use then starts a blocking download dialog. Users need to see what is ready, prepare models in Settings, and keep annotating during downloads.

## Confirmed behavior

- Settings → AI Models lists the currently supported models with capabilities, download size, status, license/details links, and relevant use restrictions. Require acknowledgment only where the actual terms require it.

- Both annotation-panel model pickers contain downloaded models plus Manage models…. An empty picker offers Download a model…. Completing a download makes the model selectable, without selecting it or running inference.

- Both AI Assist and AI Text Prompt remember model selections across restarts. Implement AI Text Prompt persistence as the first small PR. When remembered files are missing, show the unavailable state and offer restoration or another choice, without silently downloading or switching models.

- Download one model at a time with a visible queue. Downloads continue while Settings is closed and while users annotate manually or use another downloaded model. Users can cancel the current download or remove queued items.

- After failure, retain Failed and Retry for that model and continue the queue. Retry appends the model to the end.

- On quit, stop safely and discard the queue. On relaunch, downloads start only on explicit user request. Reuse completed files, restarting an unfinished file. Byte-level resume is deferred.

- Delete requires confirmation, releases any loaded session, and clears affected selections. A shared model such as SAM3 updates both panels. Annotation files are untouched.

- A completed download includes shared dependencies required by the model’s supported workflows, avoiding a hidden additional first-use download. Recognize and check existing local/offline files. Downloaded describes verified local files, not a promise of hardware compatibility.

- Model management remains accessible with command-line overrides while existing restrictions on editing Settings remain enforced.

## Scope and implementation order

1. Persist AI Text Prompt model choice in a separate small PR, using the existing configuration/apply path.

2. Add model management, background queue, accurate picker states, and model details. Reuse the existing Settings surface and download dependencies; introduce no second cache or alternate model catalog.

Cache-directory customization, new model families, downloads after quitting, automatic queue restart, and byte-level resume are outside this work. Related but distinct: #1428 (custom cache location), #2353 (large-file progress counts), and #2537 (custom ONNX models).

## Acceptance checks

[...]

PR #2698, opened 2026-09-16
feat: manage AI models and background downloads in Settings

wkentaro:
Models currently appear ready before their files exist, so first use interrupts annotation with a download. Implements #2696: prepare models in Settings, keep annotating while they download, and select a model separately once its files are present.

| Preview |
|:---:|
| Preview |

Downloads run through osam's pull on a worker thread, so the catalog, cache, mirror endpoints, retries, and hash checks stay in osam; labelme owns the one-at-a-time queue, cancel, Failed/Retry, and deletion. A model counts as downloaded when its files are present, because every writer verifies the hash before moving a file into the cache. The osam floor moves to 0.8.0 for the packaged non-maximum-suppression graph and the cancellable, time-limited pull.

Download sizes before transfer, offline license copies, byte-level resume, and downloads after quitting remain deferred. The transfer displays its reported size, and license names link to the full terms. SAM3 keeps its restrictions notice beside Download. Commercial-use summaries and badges are deferred; source-repository links are omitted from model rows.

Validation at eaaf868e: the default suite passed 1587 tests with 7 platform skips, and `make lint` passed, covering Ruff, type checks, and all 20 translation catalogs. The deletion test loads a Shape before removing a selected model and asserts it survives. Earlier macOS Computer-Use QA verified real download → selection → inference → saved annotation, queue cancellation and renumbering, retry/error details, background annotation, theme contrast, and restart persistence.


wkentaro (commented):
## Review: not mergeable yet

One verified defect blocks this. The rest is sound and worth adding.

### Blocking: the `network` e2e suite is broken

The `network-test` CI job is red (failed at 9%, then hung until the 5-minute timeout). Reproduced locally with an empty model cache:

```
FAILED tests/e2e/annotation_test.py::test_annotate_shape_types[ai_points-polygon]
  assert len(shapes) >= 1  ->  assert 0 >= 1
tests/e2e/annotation_test.py::test_annotate_shape_types[ai_points-mask]  (hangs)
```

Root cause: the canvas AI click now routes to `_ensure_ai_model_ready`, which opens Settings instead of downloading. The 11 `network`-marked tests still assume the old blocking download on first click, so no shape is created. The second test hangs because the first test's leftover `enter_label_when_visible` timer raises on the closed window and the global excepthook shows a modal `QMessageBox.critical`.

The product behavior is by design, so the fix belongs in the tests: enqueue via `win._model_manager` and `qtbot.waitUntil(lambda: manager.is_ready(...))` before clicking, and drop `close_failed_download_dialog` from those tests. The "1,532 passed" in the description never ran `-m network`. The `recommend-merge` label should come off until CI is green.

### Verified

- Default suite: 1530 passed, 7 skipped. Lint (ruff, gruff, ty, typos) clean.

- Implements #2696 as specified. osam 0.6 API usage (`_resolve_endpoints`, `_build_endpoint_url`, `Blob.attachments`, `get_model_metadata`, per-module `LICENSE`) checked for all 10 models.

- Startup hash scan of a real 7.1 GB cache: 2.6 s on NVMe. `ModelsWidget.refresh` per download chunk: ~0.15 ms, negligible.

### Non-blocking nits

- `_widgets/_download.py` blocking dialog is now dead in the app (only the canvas default, which the app always overrides). Delete it and its tests, or note a follow-up.

- `_model_download_sizes.json` is a hand-maintained second catalog keyed by hash and goes stale on every osam bump. The `get_download_size` assertion in the catalog test would catch that, so acceptable.

- Verified stamps live only in memory, so every launch re-hashes all cached files. On a slow disk with several GB, pickers stay empty for a while. Persisting `(size, mtime)` stamps would fix this cheaply. Follow-up.

- Private reach-ins: canvas sets `_ai_assist_session._session`; `ModelsWidget.refresh` reads `manager._file_index` and `manager._files`.

- `ModelManager.refresh()` silently no-ops during an active download, so "Check local model files" does nothing then.

- `assert len(sam3) == 6` in the catalog test couples to osam's catalog shape.

### Tests

Meaningful and integrated, not excessive. The `ModelManager` tests hit a real local HTTP server and real files under a temp HOME, each mapping to an acceptance check in the issue. The e2e test drives the full download → select → delete → relaunch flow through `MainWindow`.

wkentaro:
Rebased onto `main` after #2697 and addressed the review in `a60c64ed`.

- The network annotation tests now enqueue the model, wait for verified readiness or a reported download failure, and explicitly select it before drawing. The label timer is stopped in `finally`, so a failed test cannot leave a callback running against a closed window. The separate network suite passes locally: 10 passed, 1 skipped. The default suite passes: 1,530 passed, 5 skipped. Lint and all 20 translation catalogs pass.

- Removed the obsolete blocking downloader and its tests, retaining the byte-formatting checks beside the model widget. Session replacement and download file progress now have explicit methods instead of private-field access. The cache-check button is disabled while a scan or download is active and re-enables afterward. The catalog test no longer assumes exactly six SAM3 files.

- Retained the digest-keyed size data and its completeness check. Persistent verification stamps remain deferred to a separate performance follow-up, as suggested; this PR continues to verify cached files on launch.

The PR description now distinguishes default and network validation and no longer describes this as stacked. The existing screenshots still show the unchanged idle catalog and model-terms presentation.

All 21 CI checks now pass on `a60c64ed`, including `network-test`; restored `recommend-merge`.

wkentaro:
we should update gdown and osam first.

wkentaro:
**Verdict: recommend-merge.** Meaningful, matches #2696 item 2, and the diff is close to minimal for the scope: the old blocking `_download.py` dialog is deleted, the queue lives in one small `ModelManager`, and osam 0.8.0 keeps catalog, cache, retries, cancel, and hash checks upstream instead of duplicating them here.

Verified against the head (2d794920), not the description:

- osam 0.8.0 `pull(progress=, cancel=, timeout=)`, `PullCancelledError`, `remove()`, and `get_model_metadata` exist and behave as used. Cancel is checked per gdown chunk and during retry back-off, so an in-flight transfer stops promptly.
- The "downloaded means files present" claim holds: gdown downloads to a temp dir, asserts the hash, then moves the file under a lock. No unverified file can land in the cache.
- Signal ordering is safe: progress and `finished` are queued from the worker thread in order, and the next pull only starts after `finished`, so a stale progress event can never be attributed to the next model.
- Settings dialog deletion (open as text) followed by manager `changed`/`progress_changed` emits does not hit a dead `ModelsWidget` slot. Probed with a throwaway e2e test; passes.
- Delete clears both config keys, the text session, and the canvas model before removing files, and `_apply_to_live_widgets` resets the picker's remembered name so the placeholder does not keep saying Unavailable.
- Point-prompt gating survives repopulation: `set_point_prompt_mode` is re-run after every availability change, and the Manage row resolves to no option and stays enabled.

Local run in a worktree: lint clean; default suite 1581 passed, 5 skipped, 2 failed. The 2 failures are `test_enter_in_editor_reaches_default_button[*]` and reproduce identically on `origin/main` in my environment, so they are unrelated to this PR. Network suite not rerun locally; CI covers it.

Tests are minimal for the coverage: four manager tests each exercise several transitions (failure continuing the queue, retry appending, cancel keeping completed files, shutdown discarding, >2 GiB totals), one e2e test covers download-without-select, Settings-close, shared SAM3 deletion, and restart, and the `_format_bytes` table is a straight move from the deleted test file.

Non-blocking observations, no change requested:

- `ModelsWidget.refresh` runs on every progress chunk (512 KiB) and stats every blob of every model each time. Cheap in practice, but it is the first place to look if the Settings page ever feels sluggish during a fast download.
- `shutdown()` waits on the worker; if a server stalls, quit can block up to the 30 s read timeout. The `ponytail:` note in `_start_next` describes the sibling delay for the next queued download but not this one.
- A remembered model whose files were missing is not selected when its download completes in-session, but it will be selected on the next launch because the config still names it. Consistent with the "never select on download" rule in #2696, just worth knowing.
- Download size before transfer and the offline license text from #2696 are deliberately omitted, with the reasoning in the description. That is a scope call the maintainer already made as author.

wkentaro:
Rebased onto main at acd257c3 (head now 561276d9). Only conflict was the import rename to `apply_default_flags` in `_app.py` plus the binary `.qm` catalogs, which were regenerated from the merged `.ts` sources. No behaviour change. Re-verified on the new head: lint clean, default suite 1595 passed, 5 skipped, same 2 pre-existing failures as before. Keeping `recommend-merge`.

labelme/_widgets/_model_picker.py
```diff
@@ -0,0 +1,77 @@
...
+                    self.addItem(display, name)
+            if self._managed:
+                self.addItem(
+                    self.tr("Manage models…")
+                    if self._available
+                    else self.tr("Download a model…"),
+                    "manage",
+                )
+            self._show_missing_selection()
+            self.setCurrentIndex(self.findData(self._selected))
```
wkentaro, line 35: **Defect:** restoring a remembered model leaves it unselected until restart. `_selected` is only set from availability at call time and `set_available_models` never re-promotes `_requested` once its files arrive. With `text_model: sam3:latest` in config, downloading SAM3 leaves both pickers at index -1 while config still names it, so Run calls `_ensure_ai_model_ready("")` and opens Settings instead of running. Relaunching selects it. The e2e test at `tests/e2e/model_management_test.py:56` asserts this -1 state.

Fix: fall back to `_requested` here when it is now in `_available`. "Download never selects" is about other models; this selection already exists in config.

labelme/_app.py
```diff
@@ -2895,9 +2905,58 @@ def _read_flag_dock_states(self) -> dict[str, bool]:
...
+        )
+
+    def _remove_ai_model(self, model_name: str, /) -> None:
+        option = _ai_models.find_ai_assist_model_option(model_name=model_name)
+        overrides = []
+        if option is not None and self._config["ai"]["default"] == option.display_name:
+            overrides.append((("ai", "default"), None))
+        if self._config["ai"]["text_model"] == model_name:
+            overrides.append((("ai", "text_model"), None))
+        if self._is_settings_editable and not self._try_set_overrides(
```
wkentaro: **Defect:** deleting an unselected model rewrites the config file for nothing. `overrides` is empty here, and `set_overrides` always migrates and atomically writes. On a read-only labelmerc, deleting a never-selected model fails with a Configuration Error dialog.

```python
if overrides and self._is_settings_editable and not self._try_set_overrides(...)
```

labelme/_widgets/_models_widget.py
```diff
@@ -0,0 +1,180 @@
...
+            )
+            buttons = QtWidgets.QHBoxLayout()
+            buttons.addWidget(download)
+            buttons.addWidget(remove)
+            buttons.addStretch()
+            row_layout.addLayout(buttons, 5, 0, 1, 2)
+            layout.addWidget(row)
+            self._rows[name] = status, download, remove
+        manager.changed.connect(self.refresh)
+        manager.progress_changed.connect(self.refresh)
```
wkentaro: **GUI-thread hot loop:** gdown reports every 512 KiB and each report re-runs `get_state` for all 10 models plus the `_blobs` existence scan: 23 files, ~46 stat calls and 30 widget updates per tick, several hundred per second on a fast link.

Connect `progress_changed` to a slot that updates only the active row's status label; keep `refresh` on `changed`.

labelme/_widgets/_models_widget.py
```diff
@@ -0,0 +1,180 @@
...
+        self.refresh()
+
+    def refresh(self) -> None:
+        manager = self._manager
+        for name, (status, download, remove) in self._rows.items():
+            state = manager.get_state(name)
+            labels = {
```
wkentaro: Rebuilt per row per tick, and the `queued`/`downloading` entries are dead since the branches below overwrite `text`. Hoist to module level and drop the two entries.

labelme/_widgets/_models_widget.py
```diff
@@ -0,0 +1,180 @@
...
+                else self.tr("Retry")
+                if state == "failed"
+                else self.tr("Download")
+            )
+            download.setEnabled(state != "downloaded")
+            remove.setEnabled(
+                state not in {"queued", "downloading"}
+                and any(
+                    Path(blob.path).exists()
+                    for blob in osam.apis.get_model_type_by_name(name)._blobs.values()
```
wkentaro: Reaches into osam's private `_blobs`. `state == "downloaded"` is enough unless you specifically want to clear partial multi-file leftovers.

labelme/_widgets/_models_widget.py
```diff
@@ -0,0 +1,180 @@
...
+                )
+            )
+
+    def _download_or_cancel(self, model_name: str, /) -> None:
```
wkentaro, line 175: Re-derives `get_state`:

```python
if self._manager.get_state(model_name) in {"queued", "downloading"}:
```

labelme/_automation/_ai_assist.py
```diff
@@ -43,6 +43,11 @@ def __init__(
         self.polygon_detail = polygon_detail
         self._session = None
 
+    def set_model_name(self, model_name: str, /) -> None:
+        if self.model_name != model_name:
+            self._session = None
+            self.model_name = model_name
+
     def _get_session(self) -> OsamSession:
         if self._session is None or self._session.model_name != self.model_name:
```
wkentaro: `set_model_name` now owns invalidation, so the name comparison is dead: `if self._session is None:`.

labelme/_widgets/_model_picker.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        if self._requested and self._requested not in self._available:
+            display = dict(self._options)[self._requested]
+            self.setPlaceholderText(
+                self.tr("Unavailable: {model}").format(model=display)
+            )
+
+    def set_available_models(self, *, available: set[str]) -> None:
+        available = available.intersection(name for name, _ in self._options)
+        if self._managed and self._available == available:
```
wkentaro: `_managed` exists only to hide the manage entry before the manager's first push, which the app does in `__init__`. Always adding the entry removes the flag and this early return.

wkentaro (commented):
Lean for its scope (net +330 source lines after deleting the old modal downloader), lint clean, affected tests pass. Two verified behavioral defects and a GUI-thread hot loop should be fixed before merge; the rest are small cuts, take or leave.

Validation: ruff check/format clean; 377 passed, 3 skipped across the affected unit and e2e files. Two failures (`test_enter_in_editor_reaches_default_button`, `test_canvas_inference_failed_signal_surfaces_status_message`) fail identically on origin/main in the same order, so they are pre-existing. Network-marked real downloads not run.

labelme/_app.py
```diff
@@ -2895,9 +2905,58 @@ def _read_flag_dock_states(self) -> dict[str, bool]:
...
+        self._open_models()
+        return False
+
+    def _refresh_model_availability(self) -> None:
+        available = {
+            name
+            for name in self._model_manager.models
+            if self._model_manager.is_ready(name)
+        }
+        self._ai_annotation._model_combo.set_available_models(available=available)
```
wkentaro: main had no reach-ins to `_model_combo`. A one-line `set_available_models` passthrough on each widget keeps the boundary.

labelme/_model_manager.py
```diff
@@ -0,0 +1,148 @@
...
+            self.errors.pop(model_name, None)
+            self.changed.emit()
+
+    def shutdown(self) -> None:
+        self._closed = True
+        self.queue.clear()
+        self.active = None
+        if self._pull is not None:
+            self._pull.cancel.set()
+            self._pull.wait()
```
wkentaro, line 148: Note only: this blocks the GUI thread on quit. osam checks cancel per chunk so it is fast normally, but a stalled socket holds quit for the 30 s read timeout.

wkentaro:
Reviewed `561276d9008d7cdbda65cd9fd7ea68e2a491386a` against `acd257c34d9b49bc12bd917c40b730df286bb34c`, including an **ask-exemplar** evaluation against Qt's official documentation and examples. I would fix the cache-validation defect before merging; the overall structure is appropriately small.

**Spec — one verified defect**

- **[P2] Check existing cache contents before declaring a model downloaded.** [`_model_manager.py:62–67`](https://github.com/wkentaro/labelme/blob/561276d9008d7cdbda65cd9fd7ea68e2a491386a/labelme/_model_manager.py#L62-L67) checks file existence/size, not hashes. Reproduced with a real osam blob containing corrupt bytes: the manager reports `downloaded`, and `enqueue()` refuses repair. Loading through the real `OsamSession` then fetches replacement bytes from a local HTTP server **on MainThread**, recreating the hidden blocking download. The reproduction bypassed ONNX loading for the tiny artifact; osam/gdown verification and transfer were real. #2696 requires checking existing local/offline files and explicitly describes downloaded files as verified. Validate existing files outside the GUI thread, keep invalid files unavailable, and require an explicit download to repair them. Add this corrupt-cache regression alongside the existing local-HTTP tests.

Pre-download size remains a disclosed, incomplete issue requirement. Resolve that scope decision in #2696; duplicating upstream size metadata in labelme would add maintenance burden.

**Standards / coding structure — two small cuts**

1. [`_model_picker.py:19–55`](https://github.com/wkentaro/labelme/blob/561276d9008d7cdbda65cd9fd7ea68e2a491386a/labelme/_widgets/_model_picker.py#L19-L55): delete `_managed`. Always append the management item and compare availability directly. Both production owners receive availability before display. An in-memory comparison matched selection, placeholders, items, and signals across 48 transitions. Approximately **−3 lines**, with no observed production behavior change.

2. [`_models_widget.py:44–47`](https://github.com/wkentaro/labelme/blob/561276d9008d7cdbda65cd9fd7ea68e2a491386a/labelme/_widgets/_models_widget.py#L44-L47): reuse `find_ai_assist_model_option(model_name=name)` instead of repeating its lookup. **−3 lines**; identical lookup semantics.

**Ask-exemplar — native Qt opportunity**

3. [`_models_widget.py:171–180`](https://github.com/wkentaro/labelme/blob/561276d9008d7cdbda65cd9fd7ea68e2a491386a/labelme/_widgets/_models_widget.py#L171-L180): replace the carried-over byte formatter with `self.locale().formattedDataSize(value, 1)`. [QLocale already provides localized sizes and units](https://doc.qt.io/qt-6.8/qlocale.html#formattedDataSize), verified locally for English/German and values above 2 GiB. **Confidence: high. Tradeoff:** accept Qt's precision, units, and boundary behavior; this is an optional simplification, not an equivalent replacement. It removes roughly 38 lines including the formatter-specific test file; retain the transfer/progress integration test.

The application-owned manager, blocking-download `QThread`, shared picker, and Settings widget have distinct responsibilities. Qt's [QThread example](https://doc.qt.io/qt-6.8/qthread.html#details) and [PySide thread/signals example](https://doc.qt.io/qtforpython-6/examples/example_widgets_thread_signals.html) support this use of `run()` and signals. This is a supported convention, not a requirement to add a worker-object layer. **Confidence: high:** the download operation needs no worker event loop. Keep the current structure and HTTP-backed lifecycle tests.

Validation: **601 focused tests passed**, 10 network tests deselected; Ruff formatting/lint, gruff, and ty passed. Actual model inference and platform-specific loaded-session deletion were not exercised. No repository edits.

Net: **−6 behavior-preserving lines**, approximately **38 more** if adopting Qt's formatting policy.

labelme/_widgets/_model_picker.py
wkentaro: Fixed in 0ab76e53: `set_available_models` now selects `_requested` whenever it is in the new availability set, so a remembered model is restored the moment its files land. The e2e test asserts both pickers select it and that no session is created.

labelme/_app.py
wkentaro: Fixed in 0ab76e53: the config write is skipped when `overrides` is empty.

labelme/_widgets/_models_widget.py
wkentaro: Fixed in 0ab76e53: `progress_changed` now drives `_refresh_progress`, which updates only the active row's status label; `refresh` stays on `changed`.

labelme/_widgets/_models_widget.py
wkentaro: Fixed in 0ab76e53: the label map is built once per refresh with the three terminal states only; queued and downloading take their own branches.

labelme/_widgets/_models_widget.py
wkentaro: Fixed in 0ab76e53: Delete is enabled iff the state is `downloaded`. Partial multi-file leftovers are not deletable from the UI; they are overwritten by the next download of that model.

labelme/_widgets/_models_widget.py
wkentaro: Fixed in 0ab76e53.

labelme/_automation/_ai_assist.py
wkentaro: Fixed in 0ab76e53. The one test that assigned `model_name` directly now goes through `set_model_name`.

labelme/_widgets/_model_picker.py
wkentaro: Fixed in 0ab76e53: `_managed` is gone and the management entry is always present.

labelme/_app.py
wkentaro: Fixed in 0ab76e53: both widgets expose `set_available_models`; the app no longer touches `_model_combo`.

labelme/_model_manager.py
wkentaro: Acknowledged; the `ponytail:` note in `_start_next` now names the quit wait alongside the queue delay.

wkentaro:
Addressed both reviews in `0ab76e53`, folded into the feature commit so the history stays at two commits.

Fixed: remembered model restored when its files land (with the e2e assertion flipped to match), no config write when deleting an unselected model, per-chunk progress updates touch only the active row, label map hoisted, `_managed` removed, `find_ai_assist_model_option` reused, Delete enabled from state alone, `_download_or_cancel` uses `get_state`, `_get_session` relies on `set_model_name` for invalidation, `set_available_models` passthroughs replace the `_model_combo` reach-ins, and `QLocale.formattedDataSize` replaces the carried-over byte formatter and its test file.

Not changed: the P2 on cache validation. Existence-only readiness is the maintainer's decision from osam#96, on the grounds that every writer verifies the hash before moving a file into place. The residual case is a file corrupted after landing, which osam re-verifies and re-fetches on load. That fetch does happen on the loading thread, so if the maintainer wants it covered, the fix is either a background hash scan here or, better, an osam-side check that labelme can call before load.

Validation on `0ab76e53`: lint clean; default suite 1,582 passed, 5 skipped, plus the two `test_enter_in_editor_reaches_default_button` failures that reproduce on `origin/main`; network suite 10 passed, 1 skipped; translation catalogs regenerated and checked.

tests/e2e/model_management_test.py
```diff
@@ -0,0 +1,204 @@
...
+    assert not Path(blob.path).exists()
+    for key in ("default", "text_model"):
+        editor = dialog._editors[("ai", key)]
+        assert isinstance(editor, QtWidgets.QComboBox)
+        assert editor.currentData() is None
+        assert editor.currentText() == "(none)"
+    assert win._ai_text.get_model_name() == ""
+    assert win._canvas_widgets.canvas.get_ai_model_name() == ""
+    assert safe_load(config.read_text())["ai"] == {"default": None, "text_model": None}
+    assert not win._canvas_widgets.canvas.shapes
```
wkentaro: This is vacuously true: the test never creates a shape, so it passes whether or not deletion touches annotations. It is also the only coverage of the issue's "Annotation files are untouched" line. Either add a shape before the delete and assert it survives, or drop the assertion.

wkentaro (commented):
Review at 42f1a95c. Verdict: approve after one small test fix. Worth adding, correct as far as verified, close to minimal.

**Verified locally**

- `make lint` passes; `pytest tests/` gives 1587 passed, 7 skipped.

- The "present means hash-verified" premise holds: gdown verifies the hash before moving the temp file into the cache, so a corrupt transfer never lands.

- QThread subclass plus a `threading.Event` is the right shape here, since osam 0.8.0's `pull` takes exactly that Event. Cross-thread signal delivery is queued, so state stays on the main thread.

- Retry appends to the end, quit clears the queue and cancels the active pull, Prompt Compatibility is checked before the readiness hook, and with CLI overrides the setting groups are disabled while AI Models stays enabled. The open-as-text button is disabled in that mode, so the assert in that path is unreachable.

- Known ceiling, already marked in code: quit or the next download can block up to the 30 s read timeout if a server stalls after cancel.

**Tests** are integrated as preferred: real osam against a local HTTP server for failure, retry order, cancel, shutdown, offline recognition, and over-2 GiB progress, with no MagicMock; e2e tests drive the real main window. The pixel-scan link-contrast test is brittle by construction but catches the real dark-theme bug found in QA, so keep it.

**Required before merge:** one vacuous assertion, see the inline comment.

**Non-blocking follow-ups**

- SAM3 shows a passive notice only. The issue says acknowledge "only where the actual terms require it"; Hugging Face gates these weights behind an explicit click. Maintainer's call whether the license text itself demands one.

- Pre-transfer download size is omitted and disclosed in the PR body; it stays a literal gap against the issue.

- `Canvas.ensure_ai_model_ready` is a bare public attribute assigned from the app, while the same class injects its color resolver through a setter. A setter would match.

- The SAM3 row in AI Models reads "SAM3 (smart)" because the text-model list overwrites the AI Assist display name during the dict merge, while the AI Assist picker says "Sam3".

- The e2e tests index the widget's private row tuples by position; a small named tuple would make both sides readable.

- The deletion path in the app re-implements the two-line body of the single-setting apply path in a loop.

Skipped on purpose: a per-row progress bar. The byte text already carries the information and nothing in the issue asks for it.

tests/e2e/model_management_test.py
wkentaro: Fixed in eaaf868e: the test now loads a shape before the delete and asserts it survives. Re-verified at that head: lint passes and the default suite gives 1587 passed, 7 skipped.

PR #2700, opened 2026-09-16
feat: add sidebar search to Settings

### LINKED ISSUE #2699: feat: add sidebar search to Settings
## Outcome

Add Apple-style sidebar search inside the existing Settings dialog so users can quickly find and jump to a Setting Control using familiar words, config keys, or curated related terms. Search runs locally and updates as the user types, without AI, model downloads, network requests, or a new search dependency.

## Scope

- Search every Setting Control currently exposed in the dialog, regardless of the current section or scroll position. Include disabled controls, preserving their restrictions.

- Search definitions and available choices, including choices that are not currently selected. Do not index user-entered values such as predefined label lists.

- Config-only settings, main-window search, search history, typo correction, and arbitrary natural-language understanding are outside this change. Common intent phrases are supported only through curated aliases.

## Layout and navigation

- Place a standard search field with the placeholder and accessible name “Search settings” above the left sidebar. Provide its native clear button.

- With an empty or whitespace-only query, display the existing section navigation. With a nonempty query, replace it with matching settings; show each setting once, with its localized name and a smaller secondary section subtitle. Pane section headings are also smaller than setting labels; controls retain their normal text size.

- Update results as the user types. Typing or moving selection among results does not scroll the right pane or change a Setting.

- Activating a result by click or Enter places its row near the top of the existing right pane, including settings near the end. Use a clearly visible accent outline and tinted background; keep the highlight until the query changes or another result is activated. Keep surrounding settings visible and retain the query and result list. When navigating to the first setting in a section, keep the section heading fully visible above it.

- Clearing the query restores section navigation at the current page position, with the corresponding section selected. Do not restore an earlier scroll position.

- Closing and reopening Settings clears the query, preserves the last viewed page position, and focuses search.

## Matching and ordering

- Match setting names, descriptions, section names, config key paths, available choice values and labels, and a small curated set of aliases and common phrases.

- Search both English and the current UI language, plus config keys. Translate aliases through the existing translation mechanism; retain English matches alongside translated ones.

- Ignore case and surrounding or repeated whitespace. Allow partial-word matches. Every whitespace-separated query term must occur in the setting's searchable text or aliases, in any order and potentially across different metadata fields. Do not interpret regular expressions or require typo tolerance.

- Rank direct setting-name and config-key matches before matches found only through other metadata or aliases. Prefer exact name/key matches over partial matches. Preserve the existing settings order for ties.

- Keep matching deterministic and immediate at the dialog's scale. Reuse the existing settings definitions and sidebar with standard Qt controls; a dedicated search engine or completion popup is unnecessary.

## Keyboard and accessibility

- Opening Settings focuses search. Cmd+F on macOS and Ctrl+F on other platforms focus search and select the existing query.

- From search, Up/Down navigate result selection. Enter activates the selected result; with no result selected, activate the first result if one exists. Enter with no matches does nothing. Search activation must not trigger the dialog's Close button or change a Setting value.

- After activating an enabled destination, keep focus on the result without entering the editor or changing its value. Editing requires an explicit click or keyboard focus navigation. Cmd/Ctrl+F returns to search.

- For a disabled destination, scroll to and highlight its row, keep keyboard focus on the result, and expose the reason it is unavailable visually and to assistive technology. Search never enables the control or changes its prerequisite settings.

- Escape clears a nonempty query before closing the dialog; with an empty query, retain the existing close behavior. Preserve native handling of input-method composition and open child editors or popups before handling search shortcuts.

- Preserve logical Tab/Shift+Tab navigation, visible focus, accessible field and result names, and programmatic selection state. Each result's accessible name includes its setting name and section.

- Expose result status, including no matches, to assistive technology without stealing focus or interrupting typing. The row highlight is supplementary; keyboard focus on the result remains visibly identifiable.

- Preserve the existing immediate-apply behavior and normal pending-edit/save-failure handling. Searching and navigation do not themselves change Setting values.

## Empty results

Show “No matching settings in this dialog. Try another term or clear search.” in the sidebar when nothing matches. Keep the right pane at its current position, keep search usable, and provide the native clear button. The message must not imply that no corresponding config-only setting exists.

## Search acceptance examples

These are expected inclusions unless a first-result expectation is stated; other relevant results may also appear.

| Query and context | Expected result |
| --- | --- |
| `autosave` | Save automatically |
| `duplicate` or the curated phrase `avoid duplicates` | Suppress existing Shape matches |
| `dark`, with another theme currently selected | Color theme |
| `label color` | Shape Color Mode and By Label fallback color |
| `poly det` | Polygon detail |
| `auto_save` | Save automatically is first |
| `shape_color.mode` | Shape Color Mode is first |
| `自動保存` in the Japanese UI | Save automatically, shown with its Japanese label |
| `autosave` in the Japanese UI | The same localized Save automatically result |
| `  AUTO_SAVE  ` | Same result as `auto_save` |
| `zzzznonexistent` | Recoverable no-results state |

## Interaction acceptance checks

- Complete a keyboard-only open → search → select → jump → explicitly focus a control → edit → return-to-search journey without unintended changes or closing Settings.

- Verify that typing and arrow selection leave the right pane still; activation places the correct row near the top, including near the end of the scroll area, without focusing the editor.

- Verify query retention after activation, Escape/clear behavior, and a fresh query with preserved page position after reopening.

- Verify disabled-result navigation and an understandable prerequisite explanation without bypassing existing restrictions.

- Verify English and localized matching, accessible result/status feedback, native text composition, and readable results at larger system font sizes.

- Verify immediate search feedback on supported desktop platforms and preserve existing Settings apply, persistence, and save-failure behavior. Record user-facing implementation changes in a towncrier fragment when the feature is built.

## Design rationale and references

The sidebar follows the existing Apple-style Settings layout and provides direct navigation to real controls with surrounding context. Curated aliases cover familiar vocabulary at low runtime and maintenance cost; they deliberately stop short of general intent understanding. Scope stays limited to visible dialog controls so every result has an actionable destination.

- [Apple System Settings search](https://support.apple.com/en-ie/guide/mac-help/mchl8d10839d/mac) supports the familiar-word → result → destination interaction.

- [Apple search guidance](https://developer.apple.com/design/human-interface-guidelines/searching) supports clear scope and a discoverable search location.

- [W3C focus order](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) and [status messages](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html) inform the keyboard and accessible-feedback acceptance criteria. [WCAG2ICT](https://www.w3.org/TR/wcag2ict-22/) is informative desktop guidance, not a new compliance claim.

- [Qt's native clear button](https://doc.qt.io/qt-6/qlineedit.html#clearButtonEnabled-prop) covers the basic search-field affordance without a custom widget.

This issue records the agreed specification. Implementation is proposed in #2700 for human review.

wkentaro:
Finding a setting currently requires knowing its section or scrolling through the dialog. This adds local sidebar search so familiar terms such as `autosave`, `duplicate`, and `smoothness` lead directly to the relevant control, without AI or a new dependency.

Closes #2699.

| Before (Settings hierarchy (macOS)) | After (Settings hierarchy (macOS)) |
|:---:|:---:|
| Before | After |

Captured from a separate inactive native macOS window: setting names stay primary, section context is smaller, and the first setting keeps its section heading visible.

Activation scrolls the row near the top and highlights it but keeps focus on the result list, because a focused combo or slider would change its value on the next arrow key. Tab from the result then enters the highlighted setting through a `focusNextPrevChild` override; the static tab chain starts at the first editor and would otherwise jump back to the top of the pane. Up/Down in the search field are forwarded to the list with `sendEvent`, and Enter from the search field or result list is consumed in the page's event filter so it never reaches the default Close button. The filter is also installed on the content widget for resize tracking, but key handling is limited to the sidebar: editors that ignore Return propagate it up to the content widget, and Enter there must still reach the default button as before.

Non-obvious bits:

1. Activation grows the content's minimum height so settings near the end can also land near the top. It never shrinks back, since clearing the query must keep the current page position; the cost is blank space below the last section for the dialog's lifetime.

2. Group boxes get a smaller font for their headings, and each row re-sets its own point size. Copying the dialog font alone would resolve back to the group box, since the dialog font carries no resolved size; re-setting the size marks it resolved.

3. Combo items keep their English source label in a custom item role so choices such as `System default` match in both English and the current UI language.

4. The delegate reads the display text from the model rather than the style option: PySide6 leaves the copied option's text empty after `initStyleOption`.

Validation: the 114 settings unit and e2e cases pass, covering bilingual metadata and choices, keyboard and mouse navigation including Tab into the destination, persistence, disabled controls, highlight geometry after resize, restoration, empty results, Tab order, and Enter from an editor reaching the default button both before and after a search. `make lint` and `make check_translate` pass. Native IME interaction and live spoken screen-reader announcements have not been manually verified.


wkentaro:
**Verdict: approve at bd35bc7f, with trims recommended before merge.**

Verified locally on the PR head: settings unit and e2e tests pass (107), `make lint` and `make check_translate` pass. Traced the search-mode guards in both scroll-sync directions, Escape handling, whitespace queries, and tab order. No defects found.

**Worth adding?** Marginal yes. 25 settings in 7 groups is small, so the value is the alias layer (`autosave` → Save automatically, `smoothness` → Polygon detail), not the search box itself. No new dependency, stays inside the dialog.

**Correctness.** `_find_settings` is pure, stable-sorted, and requires every term. The `IntegerSlider` focus proxy is a real fix. One intentional behavior change not mentioned in the description: Enter in the empty search box is now swallowed instead of triggering the default Close button (verified). Fine for a search field, but worth a line in the body.

**Minimality.** Three pieces go beyond the ask:

- IME composition branch in `eventFilter` (~20 lines plus a test). On macOS, Windows, and ibus/fcitx the platform consumes Return/Escape during preedit before Qt emits a key event, so this path is almost certainly unreachable outside the synthetic test. Cut it.

- Screen-reader announcement timer (~25 lines). The description says it was never verified with a live screen reader. Defer unless you want to keep it as-is.

- 1.2s highlight flash. Harmless, keep.

**Tests.** 13 widget tests (~380 lines) to 1 e2e. They drive the real dialog with real Qt events, so this is integration testing, not mock-heavy unit testing. The problem is overlap:

- The e2e test and `test_search_keyboard_jump_edits_only_after_explicit_input` cover the same type → Enter → focus → edit path. Keep the e2e one, drop the widget one.

- `test_search_finds_setting_metadata_and_aliases` (12 cases), `test_search_localized_aliases_and_english_find_localized_control` (3), and `test_search_choice_values_and_english_labels` (8) are one parametrized table.

- `test_search_preserves_visual_order_after_direct_matches` hard-codes 5 key paths and breaks the moment any setting's note mentions "color". Assert relative order of two items or drop it.

- `test_search_keeps_native_composition_from_activating_or_closing` tests the dead branch above and goes with it.

- The two `QAccessible.queryAccessibleInterface` assertions check Qt's own role-text mapping. Low value.

Keep: matching table, e2e keyboard path, disabled destination, clear/reopen restoration, empty results, tab order (regression test for a bug found mid-PR). That halves the test diff with no lost coverage.

wkentaro:
Addressed in dd7e0329 and 41d472a6.

- Removed manual IME preedit tracking and forced commit/reset, along with the synthetic composition test. Native IME interaction remains explicitly unverified.
- Consolidated aliases, localized terms, choice labels, and locale codes into one matching table. Removed the duplicate widget keyboard journey; retained slider focus/persistence checks in the e2e test and mouse activation in the restoration test.
- Replaced the five-result snapshot with a relative-order assertion. Removed the two Qt accessibility-interface mapping checks; the matching table still verifies the accessible text supplied by the application.
- Documented and tested that Enter in an empty search field does not close Settings.

Kept the announcement timer because #2699 explicitly requires accessible result/status feedback without moving focus. Live spoken announcements remain unverified, as stated in the updated description. The highlight remains unchanged.

Validation: 110 focused settings unit/e2e cases pass, as do lint, translation checks, and all 21 CI checks. Net change from the reviewed head: 115 fewer lines.

wkentaro:
**Verdict renewed at 41d472a6: recommend-merge.**

Re-verified on the new head: 110 settings unit/e2e cases pass, `make lint` and `make check_translate` pass. Production diff since bd35bc7f is only the removal of the IME composition branch; no behavior change elsewhere.

Trims from the previous round are addressed: dead composition branch and its test gone, three matching tests folded into one bilingual table, the hard-coded five-path order assertion replaced with a relative-order check, `QAccessible` interface assertions dropped, and the widget keyboard-jump test replaced by an extended e2e path that also covers Enter in an empty field and the slider edit persisting to disk. The description now states the empty-Enter behavior.

Kept by author's choice: the screen-reader announcement timer, still flagged in the description as not manually verified. Not a blocker.

PR #2701, opened 2026-09-16
fix: migrate configs from all released versions

wkentaro:
Upgrading with an older Config File can discard all settings for the session because retired keys fail validation. Preserve supported settings across every released YAML config, including when saving Settings afterward.

Recovery UI is split into #2702, stacked on this PR. Related to #2376; historical audit details are recorded in https://github.com/wkentaro/labelme/issues/2376#issuecomment-5693906134.

The same Config File loaded against `origin/main` and this branch:

```yaml
instance_label_auto_increment: true
auto_save: false
labels: [cat]
```

**Before**

```text
ValueError: Unexpected key in config: instance_label_auto_increment
```

**After**

```text
Loaded: auto_save=False, labels=['cat']
```

Explicit current values take precedence over legacy keys. Migrations also run before Settings writes so resetting a setting to its default cannot revive an older override on restart. The writer uses ordinary round-trip YAML loading; malformed YAML and non-mapping documents are preserved rather than overwritten.

Validation:

- Replayed all 193 YAML defaults across 240 version tags; 47 tags predate YAML config. PyPI coverage was checked separately during the audit. The `config-migrations` CI job fetches full history and runs this replay on pull requests and main pushes. Repeat locally with `make check_config_migrations` after fetching version tags.
- 278 focused config, CLI, Settings, and GUI tests passed on this migration-only branch.
- Ruff, gruff, type checks, and typos passed.


wkentaro:
**Verdict: correct and worth merging, but not minimal. Cut the YAML alias/merge/tag machinery in the writer, then merge.**

Verified at d2e5737:

- The core fix is right. Retiring `instance_label_auto_increment`, `add_point`, `edit_line_color`, and `edit_fill_color` matches the released defaults (present in 37, 59, 89, and 89 tags respectively). Replaying all 193 released `default_config.yaml` files loads cleanly. `setdefault` gives explicit current keys precedence over legacy ones.

- Running `migrate_config` before every Settings write is the real fix for revived overrides: `store_data: true` used to survive a save of `with_image_data: false` because the default got pruned and the legacy key stayed. The integrated test (`set_overrides` then `load_config`, file ends empty) covers exactly that path.

- Wrapping `YAMLError` into `ValueError` is a small genuine fix: a malformed file used to escape `_try_set_overrides` uncaught. Raising instead of overwriting a non-mapping file is a behavior change, but losing a `- cat` file silently was worse. Fine.

**Over-engineered: `parse_config_document`, `_reject_cyclic_aliases`, `_copy_config_value`, and the `TaggedScalar`/`ScalarBoolean` handling (~60 lines).** I replaced `parse_config_document` with a plain round-trip load and reran the suite: only two tests fail, the `<<: {store_data: true}` merge-key case and the `!!str color_theme` tagged-key case. Anchored booleans (`&b true` / `*b`) already migrate correctly without any of it. Nobody writes merge keys, `!!str` keys, or cyclic aliases in `~/.labelmerc`; it is not documented, and the cost is a hand-rolled ruamel deep copy that touches `.ca`, `.fa`, and anchors and will silently drift when ruamel changes. Drop it, along with `test_writes_migrated_yaml_merges_and_anchored_booleans` and `test_normalizes_explicit_string_keys_before_writing`. If you want the anchored-boolean case kept, it can be one plain test without the conditional assert.

Tests: the new `migration_test.py` and the reset-to-default test go through `load_config`/`set_overrides` on real files, which is the right level. The `api_test.py` changes are only the rename. `tools/check_config_migrations.py` is a rerunnable version of the audit and `tools/` already has similar scripts; keep it.

wkentaro:
**Verdict: recommend-merge.** Re-reviewed at 8f5002b: the alias/merge/tag machinery and its two tests are gone, and only the `YAMLError` wrap, the non-mapping guard, and unconditional `migrate_config` remain in the writer. Config tests (166), ruff, and the released-config replay all pass.

PR #2702, opened 2026-09-16
feat: offer config reset with a backup

wkentaro:
When the existing config loader rejects a file, offer “Continue with defaults” as the primary action and “Back up and reset” as the secondary action. Reset saves an exact-byte backup beside the Config File, clears its overrides, and continues with defaults. The explanation mentions the backup, and the success dialog shows its full path.

The recovery uses existing validation and the atomic writer. Backup and write failures preserve the original file, and the backup captures whatever the file holds when the button is clicked. Command-line conflicts cannot reset a valid file. Related to #2376; migration fixes are merged in #2701.

**Preview**

```text
Your settings couldn’t be loaded.

You can use default settings for now. Resetting saves a backup
of your configuration file first.

Continue with defaults   ← default action
Back up and reset       ← secondary action

After reset: Configuration backup saved to <full backup path>
```

The short heading and default-action emphasis follow [Apple’s alert guidance](https://developer.apple.com/design/human-interface-guidelines/alerts). Diagnostics remain below the explanation: a macOS probe with Qt 6.11.1 showed that adding Qt’s Details control replaces the native alert with a generic widget dialog. The final native preview launched, but automated accessibility inspection timed out.

Validation: all 176 focused config and GUI tests pass, including backup disclosure/path, default and Escape actions, reopening, cancellation, write/backup failures, and CLI conflicts. Ruff, gruff, type checks, typos, and all 20 translation-catalog checks pass. The continuation tests are consolidated and the dialog fixture uses button roles directly.


wkentaro:
**Verdict: `recommend-triage`.** Worth adding and the recovery path itself checks out, but the diff is not minimal: it adds a hand-rolled type table that duplicates the existing settings schema and misses a key the schema already knows.

Verified locally at 01df6fde on top of #2701: the 210 config unit and e2e tests plus the full suite, ruff, gruff, ty, and the translation check all pass. Both reset buttons, cancellation, and write failure are covered end to end through the real dialog.

**Findings**

1. `_validate_value_type` in `labelme/_config/__init__.py` re-declares choices, bool, int, and color rules that `labelme/_config/_schema.py` already carries as `kind`, `choices`, `minimum`, and `maximum`. The two tables disagree today: `ai.default` is an enum in the schema but is not checked here, so a file containing `ai: {default: 42}` still loads and `MainWindow` crashes with `TypeError: QComboBox.findText(int)` instead of opening the recovery dialog. This crash is pre-existing on `main`, but this PR is where the type check now lives and the changelog promises recovery from invalid settings. Suggest driving the check from `SETTINGS` and keeping hand rules only for keys the schema lacks (epsilon, point_size, num_backups, double_click, shape colors, shortcuts, label_flags, file_search). That closes the gap and deletes roughly half the function.

2. `tests/unit/_config/recovery_test.py`: five tests re-test `parse_config_document` behavior that `writer_test.py` already covers (anchors and merge keys, `!custom` tag rejection, cyclic aliases, `!!str` keys, hex ints). Recovery only calls that parser, so drop them; the e2e tests in `tests/e2e/config_test.py` are the meaningful coverage and run in half a second. `test_partial_reset_requires_a_mapping` uses `pytest.raises(Exception)`, which passes on any failure; narrow it or drop it.

Nits, no action required: a dead `assert config_file is not None` directly under `if config_file is not None` in `_app.py`; recovery dumps with ruamel defaults, so block lists come out as `- cat` at column 0 while `set_overrides` writes them indented; the dialog header shows the file path even when the only error comes from a command-line override.

Not verified: the claim that this diff matches the earlier combined PR byte for byte.

wkentaro:
**Verdict: `recommend-merge`.** Both findings from the previous round are addressed at cd15032d.

Verified locally on top of the rebased #2701: full suite (1565 passed, 7 skipped), ruff, gruff, ty, and the translation check all pass. A file with `ai: {default: 42}` now raises `ConfigError` at `ai.default` and the e2e test recovers it through the real dialog; on the previous head the same file crashed `MainWindow`.

What changed: `_validate_config_item` now looks the key up in `SETTINGS` and applies the schema's enum, int range, color, string-list, language, and bool rules, keeping hand rules only for keys the schema does not carry. The five parser tests moved out of `recovery_test.py` into `writer_test.py` where `parse_config_document` is exercised directly, and the mapping test now expects `ValueError` or `YAMLError` instead of any exception.

One behavior note, not a blocker: `ai.default` is now rejected when it names a model absent from the current list, so a config carrying a model removed in a future release opens the recovery dialog instead of silently falling back. That matches how `ai.text_model` already behaved.

Not verified: the byte-for-byte match claim against the earlier combined PR.

PR #2719, opened 2026-09-18
feat: recommend balanced AI models

wkentaro:
The AI model download page presents a long undifferentiated catalog, making the first download harder to choose.

| Before (AI Models) | After (AI Models) |
|:---:|:---:|
| Before | After |

## Testing

- `just lint`

- `just check_translate`

- `QT_QPA_PLATFORM=offscreen uv run pytest -q tests/e2e/model_management_test.py`


wkentaro:
**Verdict: recommend-merge** (head a6bbc176)

Meaningful: yes. The page lists 9 models with near-identical rows; two explained picks up front give a first-time user a default without hiding anything.

Minimal: yes. 39 lines in one widget, no new abstraction, all models still rendered through the same row code. The remaining 320 lines are generated translation entries.

Correct, verified at this head:

- Both recommended keys exist in the manager's model dict, so the `manager.models[name]` lookup cannot raise, and every other model still appears exactly once under "All models".
- The reason strings match the declared capabilities: `sam2:latest` supports points and boxes; `sam3:latest` has `supports_point_prompts=False` and is a text model, so "text prompts and finding multiple objects" is accurate.
- Grid row renumbering is consistent (reason 1, capabilities 2, license 3, notice 4, status 5, buttons 6); no two widgets share a cell.
- `tests/e2e/model_management_test.py`: 8 passed locally. `just check_translate` passes, and no `.ts` file has an unfinished entry. CI is green on all 22 checks.

One non-blocking nit: this is an improvement, not a defect fix, so by `changelog.d/README.md` the fragment belongs in `2719.changed.md` (minor release) rather than `2719.fixed.md`, and the title would be `feat:` rather than `fix:`.

Not verified: translation quality beyond Japanese, and the rendered look on macOS/Windows (only the PR screenshots).

